PR #58, opened 2026-08-11
fix: Tier 0 — everything that was already broken

YoungJinJung:
> **Stacked on #20.** Base is `feature/wire-audit-state`; merge that first and this rebases onto main cleanly.

Closes the seven `tier-0-broken-now` issues — the ones that were not gaps in the plan but defects running on real machines right now. Three commits, one per batch, because they touch three separate layers.

## Batch 1 — the reviewer prompt (#48, #57, #40)

**#48 The reviewer judged specification drift without ever seeing the specification.** The goal lived inside the rolling window, so a long session driven by one instruction scrolled it away. Measured before the fix:

```
019fee58: steps=245  no goal in digest
019fee8e: steps=127  no goal in digest
019feea7: steps=254  no goal in digest
```

The goal is now pinned outside the window, later prompts appear as amendments, a missing goal is stated rather than omitted, and **a `spec_drift` verdict returned without a visible goal is discarded** — instruction alone is not enforcement. Verified on the real journals: a goal at step 0 of a 49-step session now reaches a 40-record digest.

**#57 The prompt had no ceiling.** It now has a character budget with a priority order: goal and constraints, then invalidated premises, then recent actions, then older ones. Writing the test found a bug in my first attempt — the header was exempt from truncation while being itself unbounded, so a session retracting many premises would overrun through the one region nothing trims. Premises are capped at the source and the omission is reported.

**#40 Trajectory text was interpolated as prose, i.e. as instructions.** It is now fenced as data with an explicit security instruction; fence markers are stripped from content so it cannot close the region and speak as the prompt author; phrases that only make sense as an attempt to address the reviewer are marked rather than deleted, because a suppressed injection is itself evidence.

Every verdict now records what the reviewer could see (`goal_present`, `truncated`, `injection_suspected`, `steps_shown/total`), so labelling grades judgments knowing their basis.

## Batch 2 — storage and secrets (#39, #46)

**#39** Journals were a durable, world-readable copy of command history, forwarded to a model by the digest. **39 lines across 5 journals matched credential shapes here, including an AWS key id.** Redaction now happens before the record reaches disk and **keeps structure so gate rules still parse the redacted form** — pinned by a test that the pipe-to-shell rule still fires on a redacted `curl … | sh`. Directories and journals are owner-only. The arms-race limitation is stated in the module rather than implied away.

[...]

PR #67, opened 2026-08-12
fix: the plugin could not start on a default macOS PATH

YoungJinJung:
Review of the plugin work merged in #60/#62/#64/#65. Three defects, each reproduced against the merged code before fixing.

## [P0] The wrapper cannot run the package on a default macOS PATH

`scripts/spotter-hook` called an unqualified `python3`. On this machine `/usr/bin/python3` is **3.9.6**, which cannot import `tomllib`. Running the merged wrapper in a hook-like environment:

```
$ env -i PATH=/usr/bin:/bin HOME=$HOME SPOTTER_HOME=/tmp/t ./scripts/spotter-hook < payload.json
ModuleNotFoundError: No module named 'tomllib'
exit=1
journals written: 0
```

So: every hook invocation crashed, **exit 1** rather than the required fail-open 0, and a hook's stderr is discarded — from the user's side supervision was simply **absent while appearing installed**. That is precisely the failure mode `spotter doctor` exists to end.

Now:

```
spotter: no python 3.11+ on PATH; this session is NOT being supervised
spotter: set SPOTTER_PYTHON=/path/to/python3.11+ to fix
exit=0
```

The probe tests for the capability actually needed (`import tomllib`) rather than parsing a version string, honours `SPOTTER_PYTHON`, and tries `python3.14 … 3.11` before falling back to `python3`. It deliberately does **not** fall back to `uv run`: that can download an interpreter, and a multi-second stall inside `PreToolUse` is worse than not observing.

## [P0] `${PLUGIN_ROOT}` is the Copilot spelling

Evidence from the working plugin installed on this machine:

| file | variable |
| --- | --- |
| `ponytail/hooks/claude-codex-hooks.json` — what its Codex **and** Claude manifests point at | `${CLAUDE_PLUGIN_ROOT}` |
| `ponytail/hooks/copilot-hooks.json` | `${PLUGIN_ROOT}` |

An unexpanded variable makes the command `"/scripts/spotter-hook"`, which does not exist. The command now reads `${CLAUDE_PLUGIN_ROOT:-${PLUGIN_ROOT}}`, so one manifest serves all three runtimes.

## [P1] Installing the plugin silently dropped the user's configuration

Discovery was `SPOTTER_CONFIG` → `$PWD/spotter.toml` → none. The location this project's own docs use — `~/.spotter/spotter.toml` — was not in the chain, so an installed plugin ignored the gates and cadence already configured. Added to the chain, with a test that a configured `forbidden_paths` rule actually fires through the wrapper.

## Also

`dirname` replaced with parameter expansion. An external tool is one more thing that can be missing from a hook's PATH, and this script exists because such assumptions fail quietly — it was observed failing with `dirname: command not found` under a minimal PATH.

## Why the existing test did not catch any of this

`test_bundled_hook_runs_without_installing_package` passes `env={**os.environ, …}`, so it inherits the developer's PATH — including whatever modern Python is on it. The new tests build a hermetic PATH containing a `python3` that cannot run the package and assert fail-open, an actionable message, and no journal.

`spotter doctor` should also exercise the wrapper rather than only `sys.executable -m spotter`; that check belongs with #58, which owns `doctor.py`, and will land there.

### DIFF
--- a/hooks/hooks.json
+++ b/hooks/hooks.json
@@ -6,7 +6,7 @@
         "hooks": [
           {
             "type": "command",
-            "command": "\"${PLUGIN_ROOT}/scripts/spotter-hook\""
+            "command": "\"${CLAUDE_PLUGIN_ROOT:-${PLUGIN_ROOT}}/scripts/spotter-hook\""
           }
         ]
       }
@@ -16,7 +16,7 @@
         "hooks": [
           {
             "type": "command",
-            "command": "\"${PLUGIN_ROOT}/scripts/spotter-hook\""
+            "command": "\"${CLAUDE_PLUGIN_ROOT:-${PLUGIN_ROOT}}/scripts/spotter-hook\""
           }
         ]
       }
@@ -27,7 +27,7 @@
         "hooks": [
           {
             "type": "command",
-            "command": "\"${PLUGIN_ROOT}/scripts/spotter-hook\""
+            "command": "\"${CLAUDE_PLUGIN_ROOT:-${PLUGIN_ROOT}}/scripts/spotter-hook\""
           }
         ]
       }
@@ -38,7 +38,7 @@
         "hooks": [
           {
             "type": "command",
-            "command": "\"${PLUGIN_ROOT}/scripts/spotter-hook\""
+            "command": "\"${CLAUDE_PLUGIN_ROOT:-${PLUGIN_ROOT}}/scripts/spotter-hook\""
           }
         ]
       }
--- a/scripts/spotter-hook
+++ b/scripts/spotter-hook
@@ -1,16 +1,65 @@
 #!/bin/sh
-# Run the bridge directly from an installed plugin checkout; no editable Python
-# installation is required. A project-local config wins unless SPOTTER_CONFIG
-# explicitly selects another file.
-set -eu
+# Run the bridge from an installed plugin checkout; no Python installation is
+# required, because Spotter has no runtime dependencies.
+#
+# Two rules govern this script:
+#
+#   1. It must resolve a Python that can actually run the package. The system
+#      python3 on macOS is 3.9, which cannot import tomllib, so calling `python3`
+#      unqualified fails on every hook invocation — silently, from the user's
+#      point of view, since a hook's stderr is discarded.
+#   2. It must fail open. A supervisor that cannot start has to let the session
+#      continue; the alternative is breaking the work it exists to protect.
+#
+# Note the deliberate absence of `set -e`: a failed probe must not take the
+# whole hook with it.
+set -u
 
-plugin_root=$(CDPATH= cd -- "$(dirname -- "$0")/.." && pwd)
+# Parameter expansion rather than dirname: an external tool is one more thing
+# that can be missing from a hook's PATH, and this script exists precisely
+# because such assumptions fail quietly.
+case "$0" in
+    */*) script_dir=${0%/*} ;;
+    *) script_dir=. ;;
+esac
+plugin_root=$(CDPATH= cd -- "$script_dir/.." && pwd)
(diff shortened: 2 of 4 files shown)


YoungJinJung (commented):
Three issues still block this PR:

1. **[P0] The newly auto-discovered home config can still break the host session.** `spotter hook --config ...` parses configuration in `main()` before `_hook_main()` enters its fail-open boundary. A malformed `~/.spotter/spotter.toml` therefore calls `argparse.error()` and exits 2. Since this PR makes that config implicit, the wrapper must convert any non-zero hook exit back to 0 (with an actionable stderr warning), or the CLI must move config loading inside the hook fail-open boundary.

2. **[P1] `SPOTTER_PYTHON` is word-split.** Expanding it in `for candidate in ${SPOTTER_PYTHON:-} ...` breaks a valid executable path containing spaces. Probe the explicit override separately with quoted expansion, then scan the fixed command names.

3. **CI is red because the no-Python test is not hermetic.** Its PATH includes `/usr/bin:/bin`, so Ubuntu finds a usable `python` after the fake `python3`; the wrapper runs successfully and emits no warning. Give that test only the fake-bin directory.

I reproduced these paths and prepared regression coverage locally; the PR should not merge until the fixes and green CI are pushed.

YoungJinJung:
3건 처리했습니다. 2번(SPOTTER_PYTHON 워드 스플리팅)과 3번(비-hermetic 테스트)은 작업 트리에 이미 수정이 들어와 있어 검토 후 그대로 커밋했습니다. 1번은 근본 원인까지 내렸습니다.

**[P0] 래퍼만으로는 부족합니다.** 지적하신 대로 `main()`이 dispatch 전에 config를 파싱해서 `parser.error()` → **exit 2**인데, PreToolUse에서 2는 **deny**입니다. 즉 `spotter.toml` 오타 하나가 감시를 끄는 정도가 아니라 **세션의 모든 툴콜을 차단**합니다.

래퍼에서 exit 코드를 변환하는 방식은 **래퍼를 거치는 호출만** 가립니다. 그런데 이 머신에 실제 등록된 훅은 래퍼를 거치지 않습니다:

```
/Users/youngjin/.local/bin/spotter hook --config /Users/youngjin/.spotter/spotter.toml
```

직접 재현: `exit=2`.

그래서 **config 로딩을 훅 자신의 fail-open 경계 안으로** 옮겼습니다. 손상된 파일이면 stderr에 사유를 남기고 기본값으로 관찰을 계속합니다:

```
spotter: unusable config /tmp/badcfg/spotter.toml (Expected '=' after a key ...); using defaults
exit=0
저널: 1개
```

`hook` 외의 명령은 여전히 파싱 불가 config에서 exit 2로 거부합니다 — 거기서 조용히 다른 설정으로 도는 건 멈추는 것보다 나쁘니까요.

래퍼의 catch-all은 유지했습니다. 이제 훅 경로에서 non-zero가 나올 일이 없어야 하는데, **바로 그래서** 예측 못 한 실패(임포트 중 사망 등)를 덮는 가드가 남아야 합니다. 능력 검사는 통과하고 실행에서 죽는 인터프리터로 그 경로를 테스트합니다.

Codex의 수정 2건에 대한 확인:
- `SPOTTER_PYTHON` 별도 프로브 + 인용 — 공백 포함 경로 심볼릭 링크로 테스트됨
- 테스트 PATH를 fake-bin만으로 축소 — Ubuntu에서 `python`이 잡히던 문제 해소

248 tests / mypy --strict / ruff 통과.

youngjinjung-linq (commented):
Latest-head review (`6187c58`): no new blocking findings. Interpreter selection preserves quoted `SPOTTER_PYTHON`, the wrapper fails open, malformed and non-UTF-8 config errors are contained by the hook boundary, and non-hook commands remain strict. CI is green. This is ready from a correctness-review perspective.

Issue #55, opened 2026-08-11
Journal records carry no time, so every timing metric is unimplementable


YoungJinJung:
## Context

A journal record has four fields:

```
['kind', 'payload', 'snapshot', 'step']
```

There is no timestamp. Step numbers give order, never duration.

Several already-planned items silently depend on time and cannot be built as written:

- **#24 time-to-intervention** — "steps between deviation and intervention" is computable, but the metric the roadmap names is *time*, and wall-clock cost is what a user feels.
- **#33 elapsed time / efficiency** — unmeasurable from the journal; it would have to come from the Codex rollout, which does carry timestamps, creating a second source of truth.
- **#19 snapshot retention by age of last use** — the code comment already names this: dating a reused snapshot needs per-record journal timestamps.
- **#52 review budget per hour/day** — a rate limit needs a clock.
- **#44 label freshness** — labels have `labeled_at`; the records they point at do not, so "was this labeled before or after that step happened" is unanswerable.

The Codex rollout does timestamp every event, which is why this has not hurt yet — but reconciling two files by position rather than by time is exactly the kind of coupling that breaks quietly.

## Goal

A recorded time on every journal record, and one clear rule for which clock is authoritative.

## Scope

- Add a monotonic-safe wall-clock timestamp per record; the writer already holds a lock, so ordering and stamping are consistent.
- Decide authority: journal time or rollout time. Two sources will disagree, and every later analysis must know which one wins.
- Handle the existing corpus: records without a timestamp must be reported as unknown rather than defaulted to anything.
- Reopen #19's retention-by-last-use once time exists.

## Design questions

- Wall clock or monotonic? Wall clock is comparable across processes and files; monotonic survives clock adjustment but not process boundaries. Both may be needed.
- Does the hook's own latency belong on the record too, so #49's p99 comes from production rather than from a benchmark?

## Non-goals

- Distributed clock synchronisation.

## Acceptance criteria

- New records carry time; old ones are explicitly "unknown".
- One documented authoritative clock.
- #24 and #33 can be built without a second data source.

Issue #52, opened 2026-08-11
Reviewer spending has no ceiling and no visibility


YoungJinJung:
## Context

The shadow reviewer spawns a detached `codex exec` every `every_steps` proposals, per session. The guards that exist are correctness guards, not budget guards: an in-flight lock stops duplicate reviews of one session, and `SPOTTER_DISABLE` stops recursion.

What does not exist is any limit. Three concurrent sessions at cadence 25 produce reviews independently; a long session produces them indefinitely; a session that thrashes produces them faster. Nothing caps reviews per session, per hour, or per day, and nothing records what was spent — token usage is not journaled at all (#33).

The default is off, which is the right default and the reason this is not urgent. But the moment the cadence is enabled — as it is on this machine, at 25 — the user has enabled an unbounded background spend they cannot see. That is a poor bargain to offer, and a worse one to offer silently.

## Goal

A spending ceiling and a bill.

## Scope

- Per-session and per-day review caps, with the cap and the current count journaled so exhaustion is visible rather than mysterious.
- Record each review's token usage (available from `codex exec --json`) against the session.
- Report cost beside verdicts in `spotter analyze`, so the price of supervision is read next to its output.
- Define behaviour at the cap: stop reviewing, or degrade to a longer cadence. Stopping silently would reintroduce the invisible-failure problem from #41.

## Design questions

- Should the cap be counted in reviews or in tokens? Reviews are simpler; tokens are what is actually spent, and a long digest costs more than a short one.
- Does a cap belong in the plan's decision framework — that is, should exceeding it be reported as a supervision failure with its own rate?

## Non-goals

- Cost optimization.

## Acceptance criteria

- Configurable caps enforced per session and per day.
- Per-review token usage journaled and reported.
- Cap exhaustion is visible, never silent.

Issue #47, opened 2026-08-11
Version persisted schemas and define migrations


YoungJinJung:
### Task / problem

The Harden stage needs explicit schema evolution for persisted evidence. Journal, label, experiment, config, and integration-manifest records cannot rely on “missing field means old” once fields change meaning.

Add versioning/compatibility rules that support read-old/write-new where safe, refuse unsupported newer data rather than guessing, and cover mixed-version durable history. Define a migration/backup path for upgrades and keep runtime protocol/config/integration schema versions distinct where their compatibility differs.

### Implementation references — study, do not depend on

NVIDIA NeMo Relay is a useful reference for keeping multiple format/config/runtime versions explicit rather than treating one product version as the compatibility key:

- ATOF/ATIF and plugin/config versions live as separate contracts: https://github.com/NVIDIA/NeMo-Relay
- persistent install/upgrade refuses unsafe adoption of incompatible state and uses explicit generation/fingerprint checks: https://github.com/NVIDIA/NeMo-Relay/blob/main/docs/nemo-relay-cli/plugin-installation.mdx

OpenClaw's Codex integration is also a useful reference for separating wire-protocol validation from higher-level runtime state:

- protocol validators / event projection are kept at the App Server adapter boundary: https://github.com/openclaw/openclaw/tree/main/extensions/codex/src/app-server

These are references only. Spotter should keep its own schema/migration implementation and avoid adding runtime dependencies.

### Do not use one global schema version

Different persisted contracts evolve independently. Give each one its own schema/version identifier, for example:

```text
Journal / Trace IR event schema
ThreadState checkpoint schema
label/annotation schema
experiment manifest/result schema
snapshot/replay metadata schema
external-effect record schema
runtime/integration ownership manifest schema
user/repository config schema
```

Keep these distinct from:

```text
Spotter package version
CLI↔daemon IPC protocol version
Hook bridge protocol version
Codex App Server protocol/capability version
installation generation
resolved config generation
```

A package upgrade can be compatible with old journals while requiring a new Hook protocol, or vice versa. One `version = 3` field for all of them will eventually become ambiguous.

### Version every durable top-level record/file

For append-only records, include a record-level schema version so mixed history remains interpretable:

```json
{
  "schema": "spotter.trace_event",
  "schema_version": 2,
  "...": "..."
}
```

For file/container formats, include a top-level version plus optional record versions when records can evolve independently.

Do not infer version solely from missing fields once a released format can exist without those fields for legitimate reasons.

[...]

Issue #44, opened 2026-08-11
Labels are single-rater and unversioned, so the metrics rest on one opinion


YoungJinJung:
## Context

Every decision this project defers to measurement — P3 active mode, P4 injection, the ceiling call — resolves to labels, and labels currently have no reliability story at all:

- one rater produces every label, and that rater is often the same agent that wrote the code being judged;
- there is no inter-rater agreement, so a rate of 0% false positives may mean the gate is precise or may mean the labeler was generous;
- `LABELABLE_KINDS` and the verdict vocabularies have already changed once, and the label file has no schema version, so a future change silently reinterprets old labels;
- the same circularity applies to the reviewer: if the labeler and the reviewer are the same model family, the reviewer is being graded by something that shares its blind spots.

The project spends considerable effort refusing to state a rate at n=3. That care is wasted if the labels themselves are unexamined.

## Goal

Make the labels defensible enough to carry the decisions that depend on them.

## Scope

- Double-label a subset and report agreement (Cohen's kappa or plain percentage — the number matters less than its existence).
- Record who or what produced each label. The `Label` record has a `labeled_at` but no rater identity.
- Add a schema version to label records so vocabulary changes are migrations rather than silent reinterpretation.
- Written labeling criteria: what makes a gate flag a false positive, what makes a nudge correct. Two raters cannot agree on an unstated standard.
- Where an agent labels, state it, and check agent-versus-human agreement on a subset before agent labels are pooled with human ones.

## Design questions

- Is single-rater labeling acceptable for gate flags — where "was this command actually executed as a destructive action?" is close to objective — while reviewer verdicts need two raters because they are judgment calls?
- Does disagreement itself carry signal? A flag two raters split on is arguably a design problem in the rule, not a labeling problem.

## Non-goals

- A labeling platform.

## Acceptance criteria

- Rater identity and schema version on every new label.
- Written criteria in docs.
- An agreement figure on a subset, published with the rates it underwrites.

PR #71, opened 2026-08-12
feat: journal and label records carry time and a schema version

YoungJinJung:
> **Stacked on #58.** Base is `feature/tier0-reviewer-prompt`; merge that first and this rebases onto main cleanly.

Tier 1: closes **#55** and **#47**. They are one change because they are one line of the record and one set of readers — the `batch-record-schema` label groups them for exactly this reason.

## #55 — nothing recorded time

A record had four fields and none was a clock:

```
['kind', 'payload', 'snapshot', 'step']
```

Step numbers give order. Nothing gave duration. Four already-planned items silently depended on a clock that did not exist:

| item | dependency |
| --- | --- |
| #24 time-to-intervention | the metric the roadmap names is *time*, not steps |
| #33 elapsed time / efficiency | unmeasurable from the journal at all |
| #52 review budget per hour/day | a rate limit needs a clock |
| #19 retention by last use | its own code comment already named this as the blocker |

Records now carry wall-clock time. Wall rather than monotonic because it has to be comparable across processes and against the Codex rollout — the only reconciliation available.

**Absent time is reported as unknown, never defaulted.** A zero here is indistinguishable from a session that really was instantaneous. `spotter analyze` gains a span, admits `span=unknown`, and counts untimed records when a journal spans the upgrade:

```
019fee58-…: steps=441 proposals=216 snapshots=46 flagged=6 reviews=6 span=unknown   # existing corpus
tsdemo:      steps=2   proposals=1                                        span=0.0m  # written after
```

## #47 — no record said which rules produced it

The record shape has changed four times (`proposal_number`, `tool_use_id`, `cwd`, reviewer `hypothesis`). Every change so far was additive and the readers coped by treating absent fields as `None`. A change in *meaning* — a redefined verdict vocabulary, a `snapshot` field pointing at something else — would have been read silently and wrongly.

That matters more here than in an ordinary log, because journals are read by tools that **delete snapshots and spend tokens**, and they are the evidence base for every rate this project publishes.

- The version is **per record**, not per file: one journal spans an upgrade.
- A **newer** version is refused, not guessed.
- An **absent** version means the record predates versioning, which is readable by definition.
- Labels are versioned on the same terms — their verdict vocabulary has already changed once (#44 asks for this too).

```
$ head -c 80 tsdemo.jsonl
{"v": 1, "step": 0, "at": 1786514314.3874161, "kind": "tool_proposal", …
```

## Verification

Verified against the real corpus on this machine: existing journals still load, mixed-version journals load every record, and a fabricated future version is refused with a message naming both versions. 258 tests, ruff, mypy --strict clean.

## What this unblocks

#24 and #33 become implementable without a second data source, #52 can express a per-hour cap, and #19's retention-by-last-use can be reopened.


youngjinjung-linq (commented):
Latest-head review (`3fa24cc`): three data-integrity issues remain.

1. **[P1] A partially untimed journal still reports a numeric duration.** `_span_of()` computes `max(stamped) - min(stamped)` even when records are missing timestamps. With one legacy record followed by one timestamped record this prints `span=0.0m (+1 untimed)`, although the actual session duration is unknowable. This contradicts the PR contract that absent time is reported as unknown. If any record is untimed, report `span=unknown (+N untimed)`; only compute a duration when every record is timed.

2. **[P1] Schema-v1 timestamp corruption is silently downgraded to “untimed.”** `StepJournal.load()` accepts a v1 record with missing, string, boolean, NaN, or infinite `at`; strings/missing become `None`, while `True` becomes `1.0`. A v1 record is precisely the format that promises a wall-clock timestamp, so malformed timestamps must raise `SnapshotError`. Legacy v0 records may continue to omit it. Validate a finite numeric value (excluding bool) for v1+.

3. **[P2] Negative schema versions are accepted.** Both `_as_version()` and `load_labels()` only reject versions greater than the current version. Values such as `-1` therefore pass as understood schemas. Require `LEGACY_VERSION <= version <= SCHEMA_VERSION`.

CI being green does not cover these malformed/mixed-record cases; add one focused regression per behavior.

PR #91, opened 2026-08-13
feat: prove shared Codex App Server steering

### LINKED ISSUE #78: Prove shared Codex App Server observation and live steering
### Question / hypothesis

Can ordinary `codex` and Spotter use the same externally reachable Codex App Server so Spotter can observe the real user thread/turn and steer that exact active turn?

This is the architectural gate for the standalone-runtime direction. If it fails, building `spotterd` around App Server control would be premature.

### Experiment design

Test three concrete paths with the current Codex runtime:

1. **Codex-managed external App Server** — ensure/start the supported external daemon path, run ordinary `codex`, attach Spotter as another client, and compare thread/turn identities.
2. **Spotter-managed external App Server** — start the server from Spotter while preserving the normal `codex` user experience, then attach both clients.
3. **Embedded baseline** — run without an externally attachable server and record exactly which observation/control capabilities remain available.

For viable external paths, test several turns and concurrent Codex sessions. Record event ordering/duplication, active turn tracking, reconnect behavior, and `turn/steer` latency/result.

A path is viable only if all of the following hold:

- ordinary `codex` can use it without a special per-session launch ritual
- TUI and Spotter observe the same thread id
- Spotter tracks the correct active turn id across multiple actions
- Spotter receives the events needed for live supervision
- `turn/steer` changes the actual user-visible active turn
- concurrent sessions remain distinguishable
- disconnect/embedded cases can be surfaced as degraded rather than silently healthy

Choose one canonical integration strategy if a path passes. If no path passes, revise the standalone App Server architecture before implementing the daemon migration.

### Notes

No daemon rewrite, reviewer-policy work, or Homebrew packaging belongs in this issue. The output is an architectural decision backed by an E2E PoC.

YoungJinJung:
## Summary
- add a dependency-free App Server PoC observer
- join the same TUI-created thread and track its active turn
- verify `turn/steer` changes the user-visible TUI response
- document managed-daemon constraints and remaining lifecycle checks

## Result
Codex CLI 0.147.0 passed the Spotter-managed external App Server path: the TUI and observer shared thread/turn identity, `turn/steer` returned the active turn ID, and the TUI rendered `SPOTTER_STEER_OK`.

The Codex-managed standalone daemon, concurrent TUI, and reconnect checks remain follow-up work, so this PR does not close #78.

## Verification
- `.venv/bin/pytest` (251 passed)
- `.venv/bin/ruff check src tests`
- `.venv/bin/mypy src tests`
- `node --check scripts/app_server_poc.mjs`
- `node scripts/app_server_poc.mjs --self-test`
- live Codex 0.147.0 TUI + independent observer E2E

Refs #78

### DIFF
--- a/docs/app-server-poc.md
+++ b/docs/app-server-poc.md
@@ -1,0 +1,88 @@
+# App Server lifecycle / attach PoC
+
+This is the reproducible experiment harness for issue
+[#78](https://github.com/spotter-agent/spotter/issues/78). It is deliberately not a production
+App Server client.
+
+## Requirements
+
+- Codex CLI 0.147.0 or newer with `codex app-server --listen` and `codex --remote`
+- Node.js 22 or newer; the harness uses Node's built-in WebSocket client
+
+Run the self-check:
+
+```bash
+node scripts/app_server_poc.mjs --self-test
+```
+
+## Spotter-managed external server
+
+Start an App Server and copy the printed WebSocket URL:
+
+```bash
+codex app-server --listen ws://127.0.0.1:0
+```
+
+Verify a second client can initialize and list threads:
+
+```bash
+node scripts/app_server_poc.mjs probe --endpoint ws://127.0.0.1:PORT
+```
+
+Start the observer before starting the TUI:
+
+```bash
+node scripts/app_server_poc.mjs observe \
+  --endpoint ws://127.0.0.1:PORT \
+  --cwd /absolute/test/directory \
+  --steer "Reply with exactly: SPOTTER_STEER_OK" \
+  --expect SPOTTER_STEER_OK \
+  --out /tmp/spotter-app-server-poc.json
+```
+
+In another terminal, connect the ordinary TUI to the same endpoint:
+
+```bash
+codex --remote ws://127.0.0.1:PORT -C /absolute/test/directory
+```
+
+The report succeeds only after the observer joins the TUI-created thread, observes its active
+turn, receives a successful `turn/steer` response, sees the expected text in an `agentMessage`,
+and sees that turn complete.
+
+## Codex-managed daemon
+
+Test the managed path separately:
+
+```bash
+codex app-server daemon start
+codex app-server daemon version
+```
+
+As of Codex CLI 0.147.0, a Homebrew Cask install cannot start this daemon: the command requires
+the standalone installer's fixed `~/.codex/packages/standalone/current/codex` path. This path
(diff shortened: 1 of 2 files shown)


youngjinjung-linq:
## Review

### Overview
Verification harness for #78: a dependency-free Node 22 WebSocket observer (`scripts/app_server_poc.mjs`) plus a reproducible experiment doc. Joins the TUI-created thread as a second client, proves `turn/steer` changes the visible response, and emits a JSON report. Good that the PR is explicit about not closing #78 and honestly documents the Homebrew daemon constraint and the `no rollout found` race.

### Strengths
- Zero dependencies, single file, readable state machine — right size for a PoC.
- `--self-test` covers the pure functions offline, so CI can run `node --check` + self-test without a server.
- Idempotent `finish()`, distinct exit codes (0/1/2), and `process.exitCode` (not `process.exit`) so the report always flushes.
- Rollout-race retry is bounded (100ms × 20).

### Findings (all minor — PoC standards)
- **`parseArgs` silently swallows typo'd flags**: `--expct SPOTTER_OK` is accepted without error, leaving `options.expect` empty so `steerSucceeded()` passes without checking text. In a verification harness this is the one path where verification silently turns off — a one-line known-keys whitelist is worth it.
- **`--expect` without `--steer` is ignored**: the no-steer `turn/completed` path exits 0 without checking it. Reject the combination or note it in the doc.
- **Stale-turn steer**: the `thread/resume` handler steers the turn from the resume snapshot; if a newer `turn/started` landed meanwhile, `expectedTurnId` targets a finished turn and the run fails (doesn't hang). Acceptable for a PoC — worth a `ponytail:` comment.
- **Asymmetric failure reporting**: the late-steer-response path (`targetCompleted` branch in the `turn/steer` handler) exits 1 without setting `report.error`, unlike the `turn/completed` path. Only failure mode the report can't explain.
- **No `--cwd` + multiple threads**: each `thread/started` overwrites `targetThreadId`. Low risk given the doc's dedicated-test-dir guidance, but needs cleanup before the concurrent-TUI follow-up.

### Verdict
Approve. All findings are follow-up material, not merge blockers. If fixing only one before merge: the parseArgs whitelist.

Bogyie:
codex 기본 명령어 실행만으로 따로 띄운 App server에 자동 연결되지 않는 이슈도 반영해야 할 것 같아요

Bogyie:
으악..

Issue #87, opened 2026-08-12
Recover from daemon and App Server disconnects without losing thread identity


Bogyie:
### Problem / outcome

The target runtime has multiple independently failing components: `spotterd`, the Codex App Server connection, durable journals, and live thread attachments. The docs require explicit `RECOVERING` / reconnect behavior, but the current Hook-era process model has no long-lived connection or reconciliation path to implement it.

### Proposed direction

Add runtime recovery around the boundaries established in the Runtime stage:

- persist enough runtime attachment/provenance to distinguish reconnect from a new thread
- on App Server disconnect, transition connection/capabilities to degraded/reconnecting and retry with bounded backoff
- after reconnect, list/read currently loaded threads and reconcile them with Spotter's durable/live identities
- after `spotterd` restart, hydrate only the durable state required by #31 and reconcile it with current App Server state before reporting READY
- make pending reviewer/control work stale or recoverable according to target turn identity rather than replaying it blindly
- retain the fail-open PreToolUse contract while the daemon is unavailable/recovering
- measure recovery duration, reconnect success/failure, and any observation gap

### Implementation references — study, do not depend on

**OpenClaw** has already implemented a pooled/shared Codex App Server client and exposed the failure modes around ownership, resource cleanup, and reconnect state:

- shared client startup/identity/lease tracking: https://github.com/openclaw/openclaw/blob/main/extensions/codex/src/app-server/shared-client.ts
- conversation active-turn identity/control: https://github.com/openclaw/openclaw/blob/main/extensions/codex/src/conversation-control.ts
- runtime boundary and supervision constraints: https://github.com/openclaw/openclaw/blob/main/docs/plugins/codex-harness-runtime.md
- leaked handlers / phantom active-run failure caused by incomplete cleanup: https://github.com/openclaw/openclaw/issues/99271
- active session can be known-active while run identity/history is incomplete: https://github.com/openclaw/openclaw/issues/121756

**NeMo Relay** is a useful lifecycle reference even though Spotter will not depend on it:

- persistent integration lifecycle: https://github.com/NVIDIA/NeMo-Relay/blob/main/docs/nemo-relay-cli/plugin-installation.mdx
- Codex integration lifecycle: https://github.com/NVIDIA/NeMo-Relay/blob/main/docs/nemo-relay-cli/codex.mdx

[...]

Issue #85, opened 2026-08-12
Ingest App Server events into an identity-rich Trace IR and durable journal


Bogyie:
### Problem / outcome

The current runtime-independent trace is intentionally small (`kind` + unstructured payload), and Codex observation still arrives primarily through Hooks. The Observe stage needs the selected App Server stream normalized into a trace that preserves identity, ordering, timing, outcome, and provenance well enough to drive live state, detection, metrics, and replay.

### Proposed direction

Build an App Server → Trace IR ingestion path that normalizes the event families actually exposed by the production client, including where available:

- thread / turn / runtime-attachment lifecycle
- user messages and agent-visible plan/reasoning summaries
- tool/command start and result, stdout/stderr/exit outcome
- file edits/diffs and repository/worktree context
- MCP/tool/web operations
- token/timing usage

Evolve Trace IR to carry explicit fields such as event identity, agent/runtime identity, thread/turn/item/tool identity, authoritative timestamp, semantic kind/operation, resources/files, result/outcome, repository/worktree, and source provenance. Preserve unknown coverage rather than synthesizing facts.

Append normalized events to durable history while keeping the journal's role as history/recovery input rather than the normal hot-state database.

### Implementation references — study, do not depend on

**OpenClaw** has the most mature public Codex App Server event projection currently available:

- main event projector: https://github.com/openclaw/openclaw/blob/main/extensions/codex/src/app-server/event-projector.ts
- Codex harness runtime boundary: https://github.com/openclaw/openclaw/blob/main/docs/plugins/codex-harness-runtime.md
- shared App Server process/client identity: https://github.com/openclaw/openclaw/blob/main/extensions/codex/src/app-server/shared-client.ts

Its projector already separates assistant/reasoning/tool/usage projections, tracks active/completed item IDs, correlates notifications to thread/turn, handles unknown event diagnostics, and creates monotonic receipt-order timestamps when multiple mirrored events share the same wall-clock time. That decomposition is worth copying conceptually, without copying OpenClaw's product/session model.

**NVIDIA NeMo Relay** is the strongest reference for raw-vs-normalized trajectory design:

- Codex integration and captured-event limitations: https://github.com/NVIDIA/NeMo-Relay/blob/main/docs/nemo-relay-cli/codex.mdx
- runtime/event model: https://github.com/NVIDIA/NeMo-Relay
[...]

Issue #81, opened 2026-08-12
Establish Thread, Turn, and Runtime Attachment identity lifecycle


Bogyie:
### Decision / problem

The Hook-era implementation largely uses a single `session_id`, while the target runtime needs to distinguish long-lived agent threads, individual turns, and each TUI/runtime attachment period. Without that identity model, concurrent sessions, resume, reviewer freshness, steering, and recovery can target the wrong work.

### Current vs desired boundary

**Current:** journal/session identity is primarily Hook-session based, and `TraceEvent` carries only `kind` plus an unstructured payload.

**Desired:** runtime identity is explicit and transport-independent:

- Agent Thread — durable coding/conversation identity
- Turn — one user→agent execution unit
- Runtime Attachment — one live client/TUI attachment period
- Reviewer Job — one async judgment anchored to a thread + target turn

Define lifecycle transitions for new thread, turn start/completion/interruption, TUI detach, resume, concurrent attachments where supported, and daemon recovery/reconciliation.

### Constraints / tradeoffs

- Depends on the App Server identity behavior established by #78 and exposed by the production client.
- Keep Codex IDs as provenance; do not force them to be Spotter's permanent cross-adapter API.
- Do not merge live semantic state into this issue; #31 owns ThreadState contents.
- Unknown or missing runtime identity must be represented explicitly rather than guessed.
- Existing Hook-era journals need a compatibility story where exact thread/turn/attachment identity is unavailable.

### Proof / exit criteria

- Runtime code has explicit types/records for Thread, Turn, and Runtime Attachment identity.
- Event routing cannot accidentally merge two concurrent Codex threads.
- Turn completion/detach does not destroy durable thread identity.
- Resume/reconnect can associate live runtime state with the correct durable thread lineage.
- Reviewer/control callers can address a specific active turn without relying on “current session” globals.
- Tests cover concurrent threads, multiple turns, detach/resume, duplicate/out-of-order lifecycle events, and missing identity.

PR #93, opened 2026-08-13
feat: add Codex App Server client

### LINKED ISSUE #80: Implement the production Codex App Server client and capability negotiation
### Problem / outcome

The current `CodexAdapter` is only an in-memory event sink. The target runtime needs a real App Server client that can observe and control the same Codex runtime proven viable by #78, while exposing capabilities rather than hard-coding one Codex version.

### Proposed direction

Implement the runtime-facing Codex transport boundary:

- connect / initialize / disconnect against the canonical App Server strategy selected by #78
- subscribe to or consume the required live event stream
- expose thread listing/state reads needed for attach, resume, and recovery
- provide typed control methods for supported primitives such as steer and interrupt without leaking raw Codex method names into core policy
- negotiate and retain an explicit capability set, including observation surfaces, steer, interrupt, and atomic pre-tool veto availability
- preserve enough raw-event provenance for normalized events and debugging
- distinguish unsupported capability, temporary disconnect, protocol error, and transport failure

Extend the generic agent-adapter boundary only as far as the runtime actually needs; do not make Codex payload shapes the core API.

### Constraints / tradeoffs

- Depends on #78 selecting a viable App Server integration path.
- This issue provides transport/control primitives; live VERIFY/NUDGE policy and delivery semantics remain in #22.
- Reconnect/recovery policy is tracked separately; the client should expose the states/errors needed for it.
- Capability loss after a Codex update must degrade features explicitly rather than producing a binary “supported/unsupported” result for all of Spotter.

### Done when

- Spotter can initialize against the selected external App Server path.
- Required events can be consumed without using Hooks as the primary observation stream.
- Thread/runtime queries needed by later lifecycle work are available.
- Capabilities are represented explicitly and covered by tests for supported, missing, and changing surfaces.
- Control primitives are exposed through the adapter boundary and raw Codex transport details do not leak into Spotter policy code.
- Disconnect/protocol failures surface as structured runtime state rather than silent absence of events.

YoungJinJung:
## Why

The shared-server PoC proved Spotter can observe and steer the TUI's real turn, but production runtime work still lacked a typed, diagnosable App Server transport.

Closes #80

## What changed

- add an async initialized WebSocket/JSON-RPC client for raw events, thread queries, resume, steer, and interrupt
- expose supported, unknown, and unavailable capability state plus structured transport, protocol, and RPC failures
- reject unsupported server-initiated requests without blocking Codex and terminate event consumers on connection loss
- close failed/timed-out connections so late responses and additional RPCs cannot enter an inconsistent state
- cover concurrent requests, capability loss/change, disconnects, timeouts, and malformed protocol data with a local test server
- update current-state documentation after the #78 gate result

## Validation

- `.venv/bin/ruff check .`
- `.venv/bin/mypy src tests`
- `.venv/bin/pytest` (258 passed)
- wheel build and metadata inspection
- live initialize + thread/list against Codex App Server 0.147.0

## Notes

Trace IR normalization and event backpressure remain #85, thread/turn/attachment identity remains #81, and reconnect reconciliation remains #87. Atomic pre-tool veto remains unavailable through App Server and stays on the Hook boundary.

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,7 +54,9 @@ External Codex App Server
        (deterministic synchronous enforcement only)
 ```
 
-The immediate next step is **not** to blindly build the daemon. [#78](https://github.com/spotter-agent/spotter/issues/78) must first prove that ordinary Codex and Spotter can share the same external App Server, observe the same thread/turn, and that Spotter can steer the real active turn.
+[#78](https://github.com/spotter-agent/spotter/issues/78) proved shared observation and steering for a
+Spotter-managed external App Server. Current Runtime work turns that viable control boundary into
+the daemon, identity, lifecycle, and recovery components required for ordinary use.
 
 For the fastest project snapshot, read [Status](docs/status.md). For sequence and evidence gates, read [Roadmap](docs/roadmap.md).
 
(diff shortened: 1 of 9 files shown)


youngjinjung-linq:
## Review

### Overview
Closes #80 by promoting the PoC (#91) into a production transport: an async `websockets`-based JSON-RPC client (`src/spotter/app_server.py`), six tests that run a real local WebSocket server, doc updates reflecting the #78 gate result, and the project's first runtime dependency (`websockets>=17,<18`).

### Strengths
- Honest capability model: `UNKNOWN → AVAILABLE on success / UNAVAILABLE on -32601`, with `atomic_pre_tool_veto` pinned UNAVAILABLE — the code matches the documented "degraded rather than silently healthy" principle.
- Tests use a real `websockets.serve` server, not mocks — out-of-order concurrent responses, capability loss, reconnect renegotiation, malformed JSON/result are all covered. Skipping pytest-asyncio in favor of `asyncio.run` keeps dependencies minimal.
- `connect()` failure ordering is correct (`_abort_connect` before setting `last_error`, correctly surviving disconnect's `last_error = None`).
- Docs land in the same PR as the code, keeping the "#78 passed → #80 implemented" narrative consistent, including the Path A Homebrew limitation.

### Findings

**1. Server-initiated JSON-RPC requests are misclassified as notifications.** `_read_messages` routes `id + no method → response`, `method → event`. Codex App Server can send server→client *requests* (command/patch approvals) carrying **both** `id` and `method`; these land in the event queue and never get a response, so the server can block on that turn. An observer role may never trip this today, but the failure is silent. At minimum detect it and reply with an error — or declare it as a known constraint in the docstring.

**2. `next_event()` blocks forever on connection loss.** The event queue gets no termination signal, so a plain `while True: await client.next_event()` consumer hangs. Racing `wait_closed()` seems to be the intended pattern, but the API doesn't enforce it. Poisoning the queue on failure (or having `next_event` raise `last_error`) would be safer — worth settling before #85 builds directly on this API.

**3. Failure paths don't close the socket.** `_connection_failed` (request timeout, protocol error) sets DEGRADED, sets `_closed`, and fails all pending — but the socket and reader stay alive. The client reports "connection lost" while `_request` can still send: an inconsistent state that only `disconnect()` resolves. Triggering a close in `_connection_failed` would align state with reality.

**4. One request timeout escalates to whole-connection failure**, killing all other in-flight requests. Defensible if the goal is avoiding undiagnosable states — but confirm it's intentional; if so, one docstring line suffices.

**Minor:**
- `_events` queue is unbounded — fine if #85 owns backpressure.
- The post-connect request-timeout path is untested (only open_timeout via an unreachable server is covered).

### Verdict
Approve, with #1 at least documented before merge. Code quality, tests, and doc consistency are all high. #1 and #2 are API-contract issues that #85/#87 will inherit as-is — a half-line decision now (fix or explicit declared constraint) is cheaper than discovering them downstream.

Issue #79, opened 2026-08-12
Build the spotterd runtime and per-user service boundary


Bogyie:
### Problem / outcome

Spotter still runs as short-lived Hook/CLI processes. The target architecture requires a long-lived per-user `spotterd` that owns runtime state and asynchronous orchestration, but there is currently no daemon process, control RPC, runtime socket, or service-management boundary in the implementation.

Build the smallest standalone runtime foundation that other Runtime-stage work can depend on.

### Proposed direction

- add a `spotterd` process with an explicit lifecycle and health state
- define a small local control protocol for CLI/runtime operations and a real socket/handshake for liveness
- keep platform service registration behind a `ServiceManager` boundary rather than leaking `launchd` / `systemd --user` details into core logic
- support manual `daemon start|stop|restart|status` as operational escape hatches while preserving the target of ordinary `codex` requiring no per-session ritual
- separate Spotter daemon ownership from Codex App Server ownership; stopping Spotter must not blindly stop shared Codex infrastructure
- model `healthy`, `degraded`, `recovering`, and `unavailable` states explicitly enough for later status/doctor work

The exact managed-start mechanism must follow the integration strategy proven in #78 rather than pre-committing to one App Server ownership model.

### Constraints / tradeoffs

- Depends on the App Server viability decision in #78.
- A dead daemon must not turn a coding session into an outage; synchronous integration paths remain fail-open.
- Do not move semantic reviewer policy or ThreadState details into this issue; #31 owns live supervision state.
- Do not implement App Server transport itself here; that is a separate runtime component.
- Prefer a small versioned local protocol over exposing Python implementation details across process boundaries.

### Done when

- `spotterd` can start, stop, restart, and answer a liveness/control handshake.
- Multiple local clients can connect without corrupting daemon state.
- Service lifecycle is behind a testable platform abstraction.
- CLI/manual lifecycle behavior is covered by tests.
- Spotter daemon shutdown does not implicitly terminate a shared App Server.
- Failure of the daemon has an explicit degraded/fail-open behavior for downstream integrations.

Issue #31, opened 2026-08-11
Move independent supervision state into spotterd


YoungJinJung:
### Decision / problem

The Hook-era reviewer rebuilds partial audit state from durable journals. The standalone runtime needs one long-lived owner of independent supervision state so each event can update judgment-ready state incrementally.

### Current vs desired boundary

**Current:** per-Hook processes append records; claim/evidence state is rebuilt from journal history and many fields live only in a rolling reviewer window.

**Desired:** `spotterd` owns per-thread live `ThreadState`, updated from App Server → Trace IR events. It should include, where observable:

- goal and explicit constraints
- hypotheses and verified evidence
- touched files / scope
- validation state
- recent failures
- intervention history
- open verification questions
- active turn/runtime capability state

The journal becomes durable history and recovery input, not the normal hot-state store.

### Implementation references — study, do not depend on

**OpenClaw** is useful for seeing how much state a real Codex App Server integration eventually needs to track, and where naive "active session" projections fail:

- event projector and per-turn item state: https://github.com/openclaw/openclaw/blob/main/extensions/codex/src/app-server/event-projector.ts
- active-turn control identity: https://github.com/openclaw/openclaw/blob/main/extensions/codex/src/conversation-control.ts
- known failure where `hasActiveRun=true` coexists with missing run identity/history: https://github.com/openclaw/openclaw/issues/121756

**ECLoop** is useful for its explicit evidence-condition state-machine idea: conditions are compiled, tracked as trajectory evidence arrives, and unsupported commitment is postponed rather than relying on free-form reviewer memory:

- paper: https://arxiv.org/abs/2607.28815

**NeMo Relay** provides a useful session/root identity precedent and demonstrates why logical session identity and runtime-instance identity should be separate:

- session correlation design discussion: https://github.com/NVIDIA/NeMo-Relay/issues/434
- observability limitations: https://github.com/NVIDIA/NeMo-Relay/issues/425

These are references only; ThreadState remains a Spotter-owned Python model.

### Concrete ThreadState model

Keep the first implementation explicit and boring. A per-thread object with typed sub-state is preferable to a graph database or an unstructured reviewer memory blob.

A useful shape is:

```text
ThreadState
├─ identity
│  ├─ logical_thread_id
│  ├─ runtime_attachment_id
│  ├─ connection_epoch
│  ├─ active_turn_id?
│  └─ capability_snapshot
├─ task
│  ├─ user_goal
│  ├─ explicit_constraints[]
[...]

PR #94, opened 2026-08-13
feat: add runtime identity lifecycle

### LINKED ISSUE #81: Establish Thread, Turn, and Runtime Attachment identity lifecycle

YoungJinJung:
## Why

The App Server client exposes raw Codex thread and turn IDs, but runtime policy still lacked a transport-independent lifecycle model for concurrent sessions, resume, and targeted control.

This PR adds a provisional registry foundation. It has no production event consumer yet; App Server routing in #85 must validate and may reshape the interface.

Closes #81

## What changed

- add explicit Spotter Thread, Turn, Runtime Attachment, provenance, and address records
- derive deterministic agent-scoped Thread/Turn IDs while retaining external provenance
- isolate concurrent threads and track attachment lifecycle without a global current session
- make duplicate and out-of-order lifecycle events deterministic, including unobserved turn starts
- reopen a known external attachment when it reconnects after detach
- represent missing and Hook-era identity explicitly instead of guessing
- mark runtime identity as provisional and unconsumed in current-state documentation

## Validation

- `.venv/bin/ruff check .`
- `.venv/bin/mypy src tests`
- `.venv/bin/pytest` (268 passed)

## Notes

Semantic `ThreadState` remains #31. App Server normalization and identity routing remain #85; daemon/service ownership remains #79.

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,8 +91,8 @@ Legend: ✅ implemented · 🟡 partial/shadow · 🧪 proof required · 🎯 ta
 App Server can share the TUI's real thread and steer its active turn. The production client in
 [#80](https://github.com/spotter-agent/spotter/issues/80) now exposes events, thread queries,
 control methods, and per-capability degraded state. The remaining Runtime work establishes
-`spotterd` ([#79](https://github.com/spotter-agent/spotter/issues/79)), explicit thread/turn identity
-([#81](https://github.com/spotter-agent/spotter/issues/81)), and lifecycle/recovery boundaries.
+`spotterd` ([#79](https://github.com/spotter-agent/spotter/issues/79)) and the lifecycle/recovery
+boundaries around the implemented thread/turn/attachment identity foundation.
 
 ---
 
--- a/README.md
+++ b/README.md
@@ -90,9 +90,11 @@ Legend: ✅ implemented · 🟡 partial/shadow · 🧪 proof required · 🎯 ta
 [#78](https://github.com/spotter-agent/spotter/issues/78) proved that a Spotter-managed external
 App Server can share the TUI's real thread and steer its active turn. The production client in
 [#80](https://github.com/spotter-agent/spotter/issues/80) now exposes events, thread queries,
-control methods, and per-capability degraded state. The remaining Runtime work establishes
-`spotterd` ([#79](https://github.com/spotter-agent/spotter/issues/79)) and the lifecycle/recovery
-boundaries around the implemented thread/turn/attachment identity foundation.
+control methods, and per-capability degraded state. [#81](https://github.com/spotter-agent/spotter/issues/81)
+adds a provisional thread/turn/attachment registry; it remains unconsumed until App Server event
+routing in [#85](https://github.com/spotter-agent/spotter/issues/85). The remaining Runtime work also
+establishes `spotterd` ([#79](https://github.com/spotter-agent/spotter/issues/79)) and
+lifecycle/recovery boundaries.
 
 ---
 
--- a/docs/README.md
+++ b/docs/README.md
@@ -81,8 +81,8 @@ PreToolUse Hook only for deterministic blocking
 
 [#78](https://github.com/spotter-agent/spotter/issues/78) proved that the user's Codex TUI and
 Spotter can share a Spotter-managed external App Server and that `turn/steer` reaches the real
-active turn. The managed daemon, concurrent identity, and reconnect lifecycle remain explicit
-Runtime work.
+active turn. The identity foundation is implemented; daemon ownership, event routing, and reconnect
+remain explicit Runtime work.
 
 ## Roadmap vocabulary
 
--- a/docs/README.md
+++ b/docs/README.md
@@ -85,8 +85,8 @@ prove that an explicitly remote-connected TUI and Spotter can share the server a
 `turn/steer` reaches the real active user turn.
 [#78](https://github.com/spotter-agent/spotter/issues/78) proved that the user's Codex TUI and
 Spotter can share a Spotter-managed external App Server and that `turn/steer` reaches the real
-active turn. The identity foundation is implemented; daemon ownership, event routing, and reconnect
-remain explicit Runtime work.
+active turn. A provisional identity registry exists but has no production event consumer yet;
+daemon ownership, event routing, and reconnect remain explicit Runtime work.
 
 ## Roadmap vocabulary
 
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -612,6 +612,21 @@ State scope:
 | durable journal lineage | Thread |
 | fork branch point | Thread + Turn/Step |
 
+## Implemented identity foundation
+
+`RuntimeIdentityRegistry` implements this transport-independent boundary:
+
+- agent-scoped external thread and turn IDs are retained as provenance and mapped to stable opaque
+  Spotter IDs, so restart/reconciliation finds the same durable lineage;
+- each attachment period has its own identity, and concurrent attachments do not merge;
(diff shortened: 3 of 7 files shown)


youngjinjung-linq:
## Review (adversarial pass)

Bottom line: **the code is clean, but this PR builds an API with zero consumers and promotes it to "implemented ✅" in the docs.** Findings in severity order.

### 1. Zero consumers, yet status flips to ✅
Nothing imports `RuntimeIdentityRegistry` except its own tests — not `app_server.py`, not the hook path. The actual wiring belongs to #85, yet status.md flips Runtime identity ❌→✅. Every other ✅ in that table means "wired in and running"; letting the same symbol mean "unused class with tests" devalues the table. More substantively: **an interface designed without the real event stream that will feed it almost certainly gets reshaped by #85.** Example: `finish_turn` raises `ConflictingRuntimeIdentity` on conflicting terminal states — nobody has decided what a caller does with that, and whether raising is even right is unknowable until a consumer exists. 🟡 ("foundation implemented, unconsumed") is the honest marking.

### 2. Real bug: re-attaching the same attachment ID after detach returns CLOSED forever
The lookup-hit path in `attach()` returns the existing record without checking status. Scenario: TUI window closes (`detach`) → reconnects with the same `agent_attachment_id` ("window-1") → `attach()` returns a **status=CLOSED record** and the thread stays DORMANT. Tests cover duplicate-attach-while-active and anonymous re-attach, but miss exactly this path. Painful given resume is an explicit goal of this PR.

### 3. The PR body's "freshness" does not exist in the code
The body claims this models "concurrent sessions, resume, **freshness**, targeted control" — but no record carries a timestamp. `active_turns()` is a tuple in dict insertion order; "most recent" is unrepresentable. #71 just added time and schema versions to journal/label records; new identity records born without time contradict the project's own direction.

### 4. "Stable opaque ID mapping" is a one-line hash function
As `test_fresh_registry_reconciles_the_same_external_lineage` itself proves, Spotter IDs are a pure function (`uuid5`) of external IDs. The three reverse-lookup tables are bookkeeping for recomputable values, and "restart/reconciliation finds the same durable lineage" overclaims — nothing is durable (all in-memory); what exists is deterministic re-derivation. "Opaque" is also inaccurate: anyone holding the external ID can derive the Spotter ID. The docs should shrink to what the registry actually adds over the hash: lifecycle state tracking.

### 5. O(n) full scans + unbounded growth in a long-lived daemon
`thread()` scans **every attachment and every turn** in the registry to compute ACTIVE/DORMANT, and it's called from `observe_thread`/`attach`/`start_turn`/`finish_turn`/`active_turns`. There is no eviction. This code's explicit destination is a `spotterd` that stays up for weeks — per-event cost grows with cumulative turn count (O(n²) total). Not asking for a fix now, but the ceiling and upgrade path deserve at least a one-line comment in the code.

### 6. Separator injection can collide IDs at a trust boundary
`_stable_id` joins with `\0`, and nothing validates that external IDs can't contain NUL (JSON carries \u0000). `("a", "b\0c")` and `("a\0b", "c")` collide into one thread ID. A supervision tool whose identity foundation can be polluted by input from the supervised process is a principle-level issue for this project. One NUL-rejection line in `_required` closes it.

### Minor
- `order=True` on all three ID dataclasses — nothing sorts them.
- Does `agent_attachment_id` even exist in the App Server protocol? If not, the parameter and its lookup table are an imagined requirement.
- A late `start_turn` on a terminal turn retroactively assigns ownership (`turn.attachment_id or attachment_id`) — is silently changing a completed turn's ownership intended?

### Verdict
**Not mergeable as-is. Two asks:** (1) fix the detach/re-attach bug (#2) with a test; (2) walk the docs back — ✅→🟡, and drop or implement "durable lineage"/"freshness". Items 4–6 can wait for #85 and a real consumer. The most fundamental objection is ordering: the right shape of this module will be determined by #85's event routing, and freezing the interface plus declaring completion before that only creates the cost of reopening "implemented" later.

youngjinjung-linq:
## Re-review after 6fc0141f

Both merge blockers from the previous review are resolved. Verified `tests/test_identity.py` passes (10/10) on the branch.

### Addressed
- **Detach/re-attach bug (was #2):** `attach()` now reopens a CLOSED external attachment and flips it back to ACTIVE, with `test_attach_reopens_a_closed_external_attachment` covering exactly the reported scenario.
- **Status inflation (was #1):** ✅ walked back to 🟡 ("provisional lifecycle registry ... no production consumer"), and every doc site now says the registry is provisional and that #85 "must validate and may reshape this interface." This is the honest framing.
- **Overclaiming language (was #4):** "durable lineage" / "stable opaque IDs" replaced with "mapped deterministically"; the misleading test name became `test_new_registry_derives_the_same_external_identity`. Accurate now.

### Still open (carried forward, none blocking)
- **NUL separator injection (was #6):** `_stable_id` still joins with `\0` and `_required` still accepts NUL-bearing external IDs, so crafted IDs can collide. This is the one I'd still fix before merge — it's a one-line rejection in `_required` and it sits on a trust boundary.
- **No timestamps (was #3):** the PR body still promises "freshness", but no record carries time. The docs no longer claim it, so at minimum the body should drop the word — or note it's deferred to #85.
- **O(n) scans / unbounded growth (was #5):** unchanged and uncommented. Fine for a provisional module; wants a ceiling note before `spotterd` adopts it.
- Minor: `order=True` on the ID dataclasses is still unused surface; a late `start_turn` on a terminal turn still retroactively assigns `attachment_id`.

### New observation from the fix
Reopening a closed external attachment reuses the **same** `AttachmentId`, so consecutive attach/detach cycles of "window-1" are one identity — the notion of a distinct "attachment period" is gone for external IDs (anonymous attachments still get a fresh ID per attach). The updated doc wording ("concurrent external attachment IDs do not merge") is consistent with this, but it's a semantic #85 should confirm: a turn's `attachment_id` no longer tells you *which* connection period it belonged to.

### Verdict
Approve. Recommend the one-line NUL rejection lands before merge; everything else is legitimately #85's problem.

Issue #83, opened 2026-08-12
Implement transactional Codex setup and teardown with an integration manifest


Bogyie:
### Problem / outcome

The target product cannot rely on marketplace/plugin installation as its long-term integration boundary. Spotter needs an idempotent `spotter setup codex` / `spotter teardown codex` workflow that knows exactly which external state Spotter owns and can migrate existing plugin users without duplicate observation or Hooks.

### Proposed direction

Implement a Codex integration manager with a durable Integration Manifest.

`setup codex` should follow an explicit transaction:

`inspect → plan → backup → apply → start/connect → verify → commit manifest`

It should cover, as applicable to the strategy selected by #78:

- environment / Codex install detection and compatibility checks
- preparation/registration of the Spotter runtime service
- App Server integration configuration without treating a shared server as Spotter-private state
- installation of the minimal `PreToolUse` bridge
- detection and migration of existing Spotter marketplace/plugin integration
- backup/fingerprinting of external config before mutation
- synthetic verification / doctor handoff

`teardown codex` should reverse only changes attributable to Spotter and leave user-owned or shared Codex state intact.

### Constraints / tradeoffs

- Depends on #78 plus the runtime/service and App Server client boundaries.
- Repeated setup must not create duplicate Hooks, services, or config entries.
- Partial failure must reconcile or roll back instead of leaving a half-managed installation reported as READY.
- The manifest is persisted state and must participate in schema/version rules from #47.
- Migration from the current plugin path must avoid double-recording the same runtime events.

### Done when

- `spotter setup codex` is idempotent and produces a versioned Integration Manifest describing every Spotter-owned external mutation.
- A failure at each meaningful transaction phase leaves a recoverable/reconcilable state.
- Existing plugin-based Spotter integration is detected and can be migrated without duplicate Hooks/events.
- `spotter teardown codex` removes/restores only Spotter-owned changes.
- Shared App Server state is not stopped or deleted merely because Spotter is torn down.
- Setup → setup → teardown → setup is covered by integration fixtures/tests.

Issue #82, opened 2026-08-12
Route deterministic PreToolUse gates through bounded daemon IPC


Bogyie:
### Problem / outcome

The current Hook process loads config, journals the proposal, and evaluates deterministic gates locally. The target architecture keeps `PreToolUse` only as a synchronous enforcement bridge and moves the authoritative GateEngine behind `spotterd` through bounded local IPC.

Implement that bridge without turning Spotter runtime failure into a Codex outage.

### Proposed direction

- make the Hook payload a small request to `spotterd` containing only the proposal/context required for deterministic enforcement
- evaluate existing deterministic policies in the daemon-owned GateEngine
- return an explicit allow/deny result with rule/reason provenance
- enforce a strict local timeout and fail open on daemon unavailable, IPC error, malformed response, or unsupported proposal shape
- keep model calls, journal reconstruction, semantic review, metrics aggregation, and other unbounded work off this path
- make protocol/version mismatch diagnosable rather than silently interpreting incompatible responses
- preserve enough telemetry to measure Hook/IPC p50/p95/p99 latency and fail-open frequency

### Constraints / tradeoffs

- Depends on the `spotterd` control boundary.
- Existing gate semantics and fail-open guarantees must remain compatible unless a separate issue explicitly changes policy.
- Do not require a network service or reviewer provider for this path.
- The Hook should remain useful as a safe no-op/fail-open bridge if the Spotter package/runtime is partially unavailable during upgrade or uninstall.

### Done when

- A real `PreToolUse` invocation can round-trip through local IPC to the daemon and receive the same deterministic allow/deny semantics as the current local gate.
- Timeout/unavailable/protocol-error cases all fail open and emit diagnosable telemetry.
- No semantic/model work occurs on the synchronous path.
- Tests cover concurrent requests, daemon absence, timeout, malformed response, protocol mismatch, and representative allow/deny policies.
- IPC latency is measurable separately from total Hook invocation latency.

PR #101, opened 2026-08-13
feat: add spotterd runtime boundary

### LINKED ISSUE #79: Build the spotterd runtime and per-user service boundary

YoungJinJung:
## Why

Spotter needs a long-lived per-user runtime and real local liveness/control boundary before Hook IPC, managed setup, live state, or App Server event routing can move behind the daemon.

Closes #79

## What changed

- add the `spotterd` process with a bounded versioned JSON protocol over a private UNIX socket
- expose explicit `healthy`, `degraded`, `recovering`, and `unavailable` states
- add `spotter daemon start|stop|restart|status` through a platform-neutral `ServiceManager`
- serialize daemon ownership, handle long UNIX socket paths, and accept concurrent local clients
- keep shared Codex App Server ownership outside daemon shutdown
- update runtime/current-state documentation without claiming managed setup or event consumers

## Validation

- `.venv/bin/ruff check .`
- `.venv/bin/mypy src tests`
- `.venv/bin/pytest` (278 passed)

## Notes

Managed service registration remains #83, bounded Hook IPC remains #82, live supervision state remains #31, and App Server event routing remains #85.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,7 +23,7 @@ Spotter asks a more specific question than “is the agent wrong?”
 
 > **Can we detect a bad assumption, loop, scope drift, or missing validation while the agent is still working, and intervene before the mistake becomes expensive?**
 
-The repository is already beyond a scaffold. The current prototype implements Hook-based trajectory collection, deterministic gates, crash-safe journals, Git snapshots, fork/replay, a shadow reviewer, claim/evidence state, evaluation labels/metrics, and counterfactual experiment machinery.
+The repository is already beyond a scaffold. The current prototype implements Hook-based trajectory collection, deterministic gates, crash-safe journals, Git snapshots, fork/replay, a shadow reviewer, claim/evidence state, evaluation labels/metrics, counterfactual experiment machinery, and a standalone daemon/control foundation.
 
 The current prototype and target product architecture are different:
 
@@ -77,7 +77,7 @@ Legend: ✅ implemented · 🟡 partial/shadow · 🧪 proof required · 🎯 ta
 | Claim/evidence audit ledger | 🟡 | Works where observable outcomes exist |
 | Evaluation labels / metrics | ✅ | Coverage-aware evaluation and precision/FP metrics |
 | Counterfactual experiment harness | ✅ | Control/guidance same-prefix pairs |
-| Standalone `spotterd` runtime | ❌ | Target runtime boundary |
+| Standalone `spotterd` runtime | 🟡 | Process, local control handshake, and manual lifecycle implemented; managed startup/runtime consumers remain |
 | App Server primary observation | 🟡 | Client/capabilities implemented; Trace IR ingestion remains (#85) |
 | Event-driven signal engine | ❌ | Current reviewer trigger is periodic |
 | Live `VERIFY / NUDGE` | ❌ | Target: `turn/steer` |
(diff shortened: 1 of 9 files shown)


youngjinjung-linq:
## Review

### Overview
Closes #79: the `spotterd` long-lived process with a versioned newline-JSON control protocol over a private UNIX socket, `spotter daemon start|stop|restart|status`, five tests that exercise real sockets and a real subprocess, and honest doc updates (❌→🟡). The lessons from the #94 review clearly landed — docs say "manual lifecycle exists, managed setup remains" and the daemon explicitly owns nothing yet.

### Strengths
- **Handshake as liveness, not a PID file**: flock only serializes ownership; aliveness is a real protocol response. Stale sockets from a crash self-heal on the next start.
- **Inode-checked unlink in `close()`** — a nice detail that prevents racing a newer daemon's socket away.
- **Careful security posture**: socket 0600, runtime dir 0700 with `lstat`-based symlink/owner validation, uid+home-hash isolation for the `/tmp` fallback, and the 104-byte UNIX socket path limit handled (100-byte threshold).
- **Concurrent-start race converges on the daemon-side flock**; loser exits 1, winner is discovered by polling. `stop` is idempotent (exit 0 when already stopped).
- Tests cover 20 concurrent clients, protocol-mismatch isolation, dual-daemon rejection, and the full CLI lifecycle with a real spawned process.

### Findings (all minor)

**1. `_handle` doesn't catch `OSError`.** A client resetting mid-request makes `readline` raise `ConnectionResetError`, which escapes the except tuple and becomes an unhandled-task-exception log. The daemon survives, but it's noise — one `OSError` in the tuple fixes it.

**2. `cli.py` imports `spotter.daemon` at module top**, and `daemon.py` unconditionally imports `fcntl` — so the whole CLI now fails to import on Windows. Fine if the project is POSIX-only, but then it clashes with the "platform-neutral `ServiceManager`" phrasing (only the Protocol type is neutral). Either one doc line declaring POSIX-only scope, or a lazy import inside `_daemon_main`.

**3. A future protocol-version daemon can't be stopped by this CLI** — `stop` sees it as DEGRADED (available) and `shutdown` gets rejected with "incompatible daemon protocol", leaving manual kill as the only exit. Acceptable for v1, but worth deciding before managed registration in #83 (old CLI + new daemon is exactly the upgrade scenario).

**4. Health states beyond HEALTHY are test-only surface** — nothing in production calls `set_health(DEGRADED/RECOVERING)`. architecture.md's "reports healthy/degraded/recovering" reads as live behavior; it's really a reserved seat for #31/#85. Half a sentence would fix the impression.

**5. Untested paths**: stale-socket takeover, the long-path `/tmp` fallback, and `_secure_runtime_dir`'s rejection branch. All are security/recovery paths, so coverage there is worth more than average.

**6. `spotterd.log` grows unbounded** — no rotation. Fine for now; a one-line ceiling note in the code would do.

### Verdict
Approve. No blockers. The design quality is high (handshake liveness, inode guard, permission handling) and the docs don't overclaim. If picking one fix before merge: #1 — one word in an except tuple.

PR #102, opened 2026-08-13
feat(runtime): route gates through daemon IPC

### LINKED ISSUE #82: Route deterministic PreToolUse gates through bounded daemon IPC

YoungJinJung:
## Summary
- route deterministic `PreToolUse` decisions through versioned local `spotterd` IPC
- enforce a 200 ms total deadline and fail open on unavailable, timeout, malformed, mismatched, or unsupported responses
- record separate daemon evaluation, IPC, and total Hook timing telemetry
- preserve existing gate semantics and update runtime documentation

## Validation
- `ruff check src tests`
- `ruff format --check src tests`
- `mypy src`
- `pytest -q`

Closes #82

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,7 +69,7 @@ Legend: ✅ implemented · 🟡 partial/shadow · 🧪 proof required · 🎯 ta
 | Capability | Status | Notes |
 | --- | --- | --- |
 | Hook trajectory ingestion | ✅ | Current primary observation path |
-| Deterministic pre-action gates | ✅ | Bounded local daemon IPC; strict timeout and fail-open telemetry |
+| Deterministic pre-action gates | ✅ | Bounded daemon IPC; local fallback preserves enforcement if the daemon is unavailable |
 | Crash-tolerant journal | ✅ | Locking, fsync, torn-tail recovery |
 | Git snapshot / detached restore | ✅ | User HEAD/index remain untouched |
 | Fork / continuation replay | ✅ | Same-prefix continuation machinery |
--- a/README.md
+++ b/README.md
@@ -23,7 +23,7 @@ Spotter asks a more specific question than “is the agent wrong?”
 
 > **Can we detect a bad assumption, loop, scope drift, or missing validation while the agent is still working, and intervene before the mistake becomes expensive?**
 
-The repository is already beyond a scaffold. The current prototype implements Hook-based trajectory collection, deterministic gates, crash-safe journals, Git snapshots, fork/replay, a shadow reviewer, claim/evidence state, evaluation labels/metrics, counterfactual experiment machinery, and a standalone daemon/control foundation.
+The repository is already beyond a scaffold. The current prototype implements Hook-based trajectory collection, daemon-backed deterministic gates, crash-safe journals, Git snapshots, fork/replay, a shadow reviewer, claim/evidence state, evaluation labels/metrics, counterfactual experiment machinery, and a standalone daemon/control foundation.
 
 The current prototype and target product architecture are different:
 
@@ -69,7 +69,7 @@ Legend: ✅ implemented · 🟡 partial/shadow · 🧪 proof required · 🎯 ta
 | Capability | Status | Notes |
 | --- | --- | --- |
 | Hook trajectory ingestion | ✅ | Current primary observation path |
-| Deterministic pre-action gates | ✅ | Rule-based; safe/shadow-first posture |
+| Deterministic pre-action gates | ✅ | Bounded local daemon IPC; strict timeout and fail-open telemetry |
 | Crash-tolerant journal | ✅ | Locking, fsync, torn-tail recovery |
 | Git snapshot / detached restore | ✅ | User HEAD/index remain untouched |
 | Fork / continuation replay | ✅ | Same-prefix continuation machinery |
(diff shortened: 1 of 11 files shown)


youngjinjung-linq:
## Review

### Overview
Closes #82: routes the deterministic `PreToolUse` gate through `spotterd` IPC — a `gate` method on the daemon, a 200 ms total deadline, fail-open on failure, and separate `gate_ipc` telemetry (IPC / evaluation / total hook time). Tests cover concurrency, malformed and version-mismatched responses, a measured timeout bound, unsupported proposal shapes, and the hook's fail-open paths, plus an E2E wrapper test with a real daemon. Test isolation checked: `spotter_home` is autouse, so the daemon fixture never touches the real user socket.

### Blocker (1)

**This PR turns deterministic enforcement off in the default state.** Before merge, `rm -rf /` in active mode was denied by in-process `gate.check()` with zero infrastructure. After merge, with no daemon running — which is the default, since only manual `spotter daemon start` exists and managed startup is #83 — every gate decision becomes `daemon_unavailable` fail-open, and the destructive command is **allowed**. The PR's own test enshrines this: `test_missing_daemon_fails_open_with_telemetry` asserts that active-mode `rm -rf /` passes.

The core contradiction: the daemon's gate evaluation is a **pure function of the request**. The hook ships the gate config in every request, and the daemon runs the same `Gate` class the hook already imports. So whenever the daemon can't answer, the hook could evaluate locally — which is exactly what it did yesterday. The fail-open philosophy in `hook.py` ("unsupervised beats blocked") is about not *blocking* Codex when Spotter breaks; it is not about silently waving through destructive commands that were denied until this PR.

The fix is cheap: on `DaemonUnavailable`/`DaemonTimeout`, fall back to `decision = gate.check(event)` (~3 lines — `run_hook` already holds a `Gate` instance) while still recording the IPC failure status in telemetry. The #82 boundary and the old guarantee coexist. If skipping the local fallback is a deliberate decision (policy-divergence concerns, etc.), then that decision — and "enforcement requires a running daemon" — must be stated in the docs. Right now the README keeps "Deterministic pre-action gates ✅", which is a ✅ that silently doesn't function on a default install.

### Minor

- **`files=None` fail-opens the whole proposal**: `DaemonClient.gate` forwards `event.payload.get("files")` as-is, and the daemon gives up on evaluating even the command when `files` isn't a list. Safe today because `event_from_hook` always builds a list, but the moment #85's App Server-normalized events reuse this method it's a landmine. One `or []` on the client removes it.
- `request()`'s finally now calls `close()` without awaiting `wait_closed()` — understandable for the deadline, but the loop closes right after `asyncio.run`, which can surface transport-cleanup warnings. Half a line: await it outside the timeout scope under `suppress`.
- Every `PreToolUse` now pays `asyncio.run` + a fresh socket connection — ~1 ms against a 200 ms bound, and the telemetry exists precisely to measure it. Fine.

### Strengths
- Correct trust-boundary asymmetry: malformed *proposal* (agent input) fails open; malformed *gate config* (Spotter's own data) is a protocol error.
- The deadline wraps connect-through-response in one `asyncio.timeout`, so "total" really is total — and the test measures the bound (<0.1 s).
- Split telemetry (daemon_evaluation_ms / ipc_ms / hook_ms) keeps latency percentile computation off the synchronous path.

### Verdict
**Hold merge until the local fallback lands (or the no-fallback decision is made explicit and documented).** The direction (#82) and implementation quality are good, but today's net effect is: the same decisions, slower — and none at all without a daemon. Three lines of fallback dissolve the blocker; everything else here is minor.

youngjinjung-linq:
## Re-review after 9b8b81c8

The blocker is resolved. Verified `tests/test_daemon.py` + `tests/test_hook.py` pass (25/25) on the branch.

### Addressed
- **Local fallback (was the blocker):** `DaemonUnavailable` and `DaemonTimeout` now fall back to the same local `Gate` while keeping the degraded IPC status in telemetry. The tests flipped to assert the right thing — active-mode `rm -rf /` with no daemon is **denied** (`rm_root`, journaled as `gate_block`, telemetry `status=unavailable`). Enforcement no longer depends on the manual-start lifecycle.
- **`files=None` landmine:** normalized with `or []` on the client, plus a dedicated test proving a command-only proposal is still evaluated (`rm -rf /` with `files: None` → denied).
- **`wait_closed`:** restored under `suppress` in the finally, outside the deadline scope — correct placement, and it's a local close so unbounded-wait isn't a concern.
- Docs updated consistently across all five files, including the Failure policy section — "unavailable/timeout → local Gate; malformed/mismatched/unsupported → fail open" is now an explicit, documented policy rather than an accident.

### Remaining (minor, non-blocking)
- **`DaemonProtocolError` still fails open rather than falling back.** A protocol error is a daemon-side fault (malformed response, version mismatch) — the proposal itself is perfectly evaluable by the local Gate, same as the unavailable/timeout cases. As written, an old-CLI/new-daemon upgrade window silently disables enforcement again. The asymmetry is documented, which is fine for merge, but consider extending the fallback there too — it's the same one-line change, and the current split is hard to justify beyond "we wrote it that way."
- Cosmetic: on fallback paths `ipc_ms` now includes the local evaluation time (fallback runs inside the measured window). Microseconds against a 200 ms budget; the `status` field disambiguates. Not worth restructuring.

### Verdict
Approve. The net effect is now strictly additive: same enforcement as before this PR, plus the IPC boundary, timing telemetry, and diagnosable degradation. The protocol-error fallback extension can ride any later commit.

PR #110, opened 2026-08-13
feat(operations): add transactional Codex setup

### LINKED ISSUE #83: Implement transactional Codex setup and teardown with an integration manifest

YoungJinJung:
## Summary

- add versioned, atomic integration ownership manifests for `spotter setup|teardown codex`
- migrate legacy Spotter hooks/plugins transactionally with backup, verification, and rollback
- add managed launchd/systemd-user service registration plus portable mode
- document the App Server ownership boundary and lifecycle status

## Validation

- `.venv/bin/ruff check .`
- `.venv/bin/mypy src tests`
- `.venv/bin/pytest -q`
- `git diff --check`

Closes #83

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,13 +77,13 @@ Legend: ✅ implemented · 🟡 partial/shadow · 🧪 proof required · 🎯 ta
 | Claim/evidence audit ledger | 🟡 | Works where observable outcomes exist |
 | Evaluation labels / metrics | ✅ | Coverage-aware evaluation and precision/FP metrics |
 | Counterfactual experiment harness | ✅ | Control/guidance same-prefix pairs |
-| Standalone `spotterd` runtime | 🟡 | Process, versioned gate/control IPC, and manual lifecycle implemented; managed startup/runtime consumers remain |
+| Standalone `spotterd` runtime | 🟡 | Process, versioned gate/control IPC, and managed user-service startup implemented; runtime consumers remain |
 | App Server primary observation | 🟡 | Client/capabilities implemented; Trace IR ingestion remains (#85) |
 | Event-driven signal engine | ❌ | Current reviewer trigger is periodic |
 | Live `VERIFY / NUDGE` | ❌ | Target: `turn/steer` |
 | `INTERRUPT` | ❌ | Target: `turn/interrupt` |
 | `RESTART` | ❌ | Requires verified-state + side-effect-aware recovery |
-| Homebrew / setup lifecycle | 🎯 | Target product path |
+| Codex setup lifecycle | 🟡 | Transactional setup/teardown and manifest implemented; packaging and full App Server integration remain |
 
 ### Current runtime work
 
@@ -94,7 +94,9 @@ control methods, and per-capability degraded state. [#81](https://github.com/spo
 adds a provisional thread/turn/attachment registry; it remains unconsumed until App Server event
 routing in [#85](https://github.com/spotter-agent/spotter/issues/85). [#79](https://github.com/spotter-agent/spotter/issues/79)
 adds the `spotterd` process, versioned local control handshake, manual lifecycle commands, and a
-platform-neutral service boundary. Managed setup, event routing, and recovery remain separate work.
+platform-neutral service boundary. [#83](https://github.com/spotter-agent/spotter/issues/83) adds
+transactional Codex setup/teardown, managed user-service registration, and integration ownership.
+Event routing and recovery remain separate work.
 
 ---
 
@@ -312,13 +314,24 @@ claude plugin marketplace add spotter-agent/spotter
 claude plugin install spotter@spotter
 ```
 
-Plugin installation is the **current prototype path**, not the long-term product boundary.
+Plugin installation remains a compatibility path. New standalone integration can use:
+
+```bash
+spotter setup codex --dry-run
+spotter setup codex
+spotter teardown codex
+```
+
+The explicit external App Server still has to be selected with the launch command printed by setup
+until the remaining App Server ingestion/lifecycle work lands.
 
 Useful commands:
 
 ```bash
 spotter observe
 spotter hook  # hook bridge; JSON payload on stdin
+spotter setup codex [--dry-run|--portable]
+spotter teardown codex
 spotter daemon start|stop|restart|status
 spotter analyze
 spotter review --session <id>
--- a/README.md
+++ b/README.md
@@ -322,8 +322,8 @@ spotter setup codex
 spotter teardown codex
 ```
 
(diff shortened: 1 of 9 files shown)


youngjinjung-linq:
## Review

### Overview
Closes #83 (+1278): new `src/spotter/integration.py` (575 lines) — transactional `spotter setup|teardown codex` with a versioned manifest, flock serialization, fingerprinted backups, and rollback; legacy hook/plugin migration; a launchd/systemd-user `ManagedServiceManager` in `daemon.py` (~170 lines); CLI wiring; 15 tests (420 lines); doc updates.

### Strengths
- **The transaction discipline is real**: dedicated rollback tests for verification failure, manifest-save failure, and plugin-migration failure — plus config validation failing *before* any external mutation, pinned by test.
- **Consistent ownership conservatism**: a user-modified owned hook survives teardown (exact-match removal only), incidental mentions like `echo \"run spotter hook later\"` are protected by `shlex` token analysis, and portable mode refuses to claim a pre-existing daemon — all tested.
- The hook command's `[ -x ... ] && ... || true` shape never blocks Codex after Spotter is removed. Content-addressed backups accumulate without duplication.
- Manifest schema versioning, future-schema refusal, atomic 0600 writes — consistent with the project's post-#71 discipline.

### Findings

**1. Setup's success message prints a command that cannot work.** `app_server_endpoint` is fabricated as `unix://.../codex-app-server.sock` — no component creates that socket: the #79 daemon deliberately owns no App Server, and this PR's own docs say it doesn't create one. Yet setup ends by telling the user to run `codex --remote unix://...`, which will fail. Every PoC and validation doc in this repo used `ws://127.0.0.1:PORT`, and nothing verifies `codex --remote` even accepts unix endpoints (`CodexInstall.detect` only sniffs flag presence). Record the endpoint as pending (#85/#87) or reword the output to reflect its reserved status.

**2. On macOS, managed setup neuters `spotter daemon stop`.** The plist sets `KeepAlive=True` (unconditional restart), but `_daemon_main` always uses `ManualServiceManager` — a socket shutdown makes the daemon exit 0 and **launchd immediately resurrects it**. There's even a race where stop reports \"stopped\" just before the respawn. `_daemon_main` should read the manifest's runtime_mode and route to the right ServiceManager. (Linux uses `Restart=on-failure`, so clean shutdown stays down — making this a macOS-only trap, which is worse.)

**3. Teardown requires a Codex install.** `IntegrationManager.__init__` unconditionally calls `CodexInstall.detect()`, so a user who uninstalled Codex first gets \"Codex is not installed\" and can never clean up hooks/service. Teardown is mostly file edits — defer detection to the setup path.

**Minor:**
- `--portable` is silently ignored on teardown (asymmetric with `--dry-run`, which errors). One line.
- systemd unit `ExecStart`/`Environment` quoting via `json.dumps` works for simple paths but isn't systemd's escape grammar (`%` specifiers, etc.) — worth a known-ceiling comment.
- A crash before manifest save leaves the owned hook installed while teardown reports \"not configured\" — re-running setup heals forward, so acceptable, but lifecycle.md's \"[x] interrupted setup can be resumed/reconciled\" should be read with that footnote.

### Verdict
**Hold merge until #1 and #2 are addressed.** The engineering quality (transaction/rollback/tests) is the best in this series, but both top findings are of the \"reports success while behaving otherwise\" family — exactly what this project's own \"degraded rather than silently healthy\" principle forbids: a success path instructing users to run a command that fails (#1), and stop silently undone on the primary platform (#2). #3 is fine as a fast follow-up.

youngjinjung-linq:
## Re-review after ade3e7c2

All three findings are resolved, each with a dedicated regression test. Verified on the branch: full suite 96/96, `mypy src tests` clean, `ruff check` clean.

### Addressed
- **Fabricated endpoint (was #1):** `app_server_endpoint` is now `None`, strategy is `pending-external`, and both the plan output and the setup success message say "pending runtime integration (#85/#87)" instead of printing an unrunnable `codex --remote unix://...` command. `test_setup_records_app_server_endpoint_as_pending` + `test_setup_cli_does_not_print_an_unverified_remote_command` pin it.
- **macOS stop neutered by KeepAlive (was #2):** `_daemon_main` now loads the integration manifest and routes to `ManagedServiceManager` when `runtime_mode == "managed"`, so stop goes through `launchctl bootout` rather than a socket shutdown that launchd immediately undoes. `test_managed_manifest_routes_daemon_stop_through_the_service_manager` covers the routing.
- **Teardown required Codex (was #3):** `CodexInstall.detect()` is deferred to `_codex_install()`, reached only from the setup path — teardown now works with Codex uninstalled, tested by `test_teardown_does_not_require_codex_to_still_be_installed`.

### Minors also picked up
- `--portable` on teardown now errors symmetrically with `--dry-run`, with a test.
- systemd `%` specifiers are escaped (`%%`), with a ponytail ceiling comment naming the upgrade path and a test.

### Verdict
Approve — merging.

PR #111, opened 2026-08-13
Emit versioned release artifacts and runtime build identity

Bogyie:
### Motivation

- Ensure release artifacts are derived from an immutable tag and expose a reproducible package identity. 
- Make CLI/daemon/hook entry points installable and discoverable from the package without relying on Homebrew-specific paths. 
- Keep runtime compatibility contracts explicit and independently versioned from the package version. 

### Description

- Switch to VCS-driven versioning via Hatch VCS by making the package `dynamic = ["version"]` and adding `hatch-vcs` hooks in `pyproject.toml`, and add a `version-file` target. 
- Add a shared build identity in `src/spotter/build.py` with independent contract versions and helpers `build_identity()` and `version_string()`. 
- Add packaged entry points and small executables: `src/spotter/daemon.py` (provides `spotterd --version`) and `src/spotter/hook_cli.py` (provides `spotter-hook` wrapper), and wire `--version` into the CLI by calling `version_string`. 
- Add a GitHub Actions release workflow `.github/workflows/release.yml` that builds from a `v*` tag, verifies tag-derived binaries, produces `SHA256SUMS`, and publishes immutable release assets. 
- Add user-facing documentation `docs/releasing.md` describing tag-driven releases and artifact layout and a test `tests/test_build_identity.py` validating serialized build identity and that `spotter`, `spotterd`, and `spotter-hook` report the packaged identity. 

### Testing

- Ran formatting and linting checks with `ruff format --check .` and `ruff check .`, which passed. 
- Ran static typing checks with `mypy src tests`, which passed. 
- Ran the full test suite with `pytest`, which passed (`254 passed`). 
- Built distributables with `python -m build`, inspected wheel metadata for the version and entry points, and computed SHA-256 checksums for `dist/*`. 
- Verified installation in an isolated venv and checked `spotter --version`, `spotterd --version`, and `spotter-hook --help` all behaved as expected.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a7d556dfe8c832999c506bc9aad55f6)

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,38 @@
+name: Release
+
+on:
+  push:
+    tags:
+      - "v[0-9]*"
+
+permissions:
+  contents: write
+
+jobs:
+  release:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+      - uses: actions/setup-python@v5
+        with:
+          python-version: "3.11"
+      - run: python -m pip install --upgrade build
+      - name: Build from the immutable tag
+        run: python -m build
+      - name: Verify tag-derived package identity
+        env:
+          RELEASE_VERSION: ${{ github.ref_name }}
+        run: |
+          python -m venv verify
+          verify/bin/pip install dist/*.whl
+          test "$(verify/bin/spotter --version)" = "spotter ${RELEASE_VERSION#v}"
+          test "$(verify/bin/spotterd --version)" = "spotterd ${RELEASE_VERSION#v}"
+          verify/bin/spotter-hook --help >/dev/null
+      - name: Create portable checksums
+        run: cd dist && sha256sum * > SHA256SUMS
+      - name: Publish immutable release assets
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: gh release create "${GITHUB_REF_NAME}" dist/* --verify-tag --generate-notes
--- a/.gitignore
+++ b/.gitignore
@@ -8,4 +8,4 @@ __pycache__/
 .venv/
 build/
 dist/
-
+src/spotter/_version.py
--- a/docs/releasing.md
+++ b/docs/releasing.md
@@ -1,0 +1,16 @@
+# Releases
+
+Spotter releases are built only when a `vMAJOR.MINOR.PATCH` tag is pushed. The
+tag is the single source of the package version: Hatch VCS writes that identity
+into both the wheel and source distribution. Do not edit a second version file.
+
+Each GitHub Release contains the platform-independent wheel, source archive,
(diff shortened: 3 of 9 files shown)


YoungJinJung:
## Review

### Overview
+179 from a Codex cloud task: VCS-driven versioning (hatch-vcs, `dynamic = ["version"]`), a `spotter.build` module with independent contract versions, `spotterd`/`spotter-hook` entry points, a tag-driven release workflow with version verification and SHA256SUMS, and `--version` wiring. The *direction* is right and wanted (packaging is on the roadmap), but the PR is built against a main that no longer exists.

### Blockers

**1. The PR is based on pre-#101 main and collides with two weeks of runtime work.** The diff *creates* `src/spotter/daemon.py` as a new 13-line stub whose `main` errors with "the standalone daemon runtime is not implemented in this release" — but main's `daemon.py` is now a 500+ line real implementation (#79/#82/#83: DaemonServer, gate IPC, ManagedServiceManager), and `pyproject.toml` already declares `spotterd = "spotter.daemon:main"` plus the `websockets` dependency. Merged as-is this is an add/add conflict at best; if it ever landed, the stub would replace the real daemon entry point with an error message. The PR body's "254 passed" confirms the stale base (current suite is 330+). Needs a rebase that **drops `daemon.py` and the duplicate `spotterd` entry-point line entirely** — both already exist.

**2. `spotter.build` duplicates authoritative constants instead of importing them.** `IPC_PROTOCOL_VERSION = 1`, `JOURNAL_SCHEMA_VERSION = 1`, `INTEGRATION_MANIFEST_VERSION = 1` etc. are fresh copies, while the real values live in `spotter.daemon.PROTOCOL_VERSION`, `spotter.snapshot.SCHEMA_VERSION`, and `spotter.integration.MANIFEST_SCHEMA`. Two sources of truth guarantee drift — the first protocol bump that forgets `build.py` makes `build_identity()` report a false compatibility contract, which is worse than not reporting one. Import the authoritative constants; delete the copies.

### Worth keeping (after rebase)
- hatch-vcs tag-derived versioning with `no-local-version`, and the release workflow's identity verification step (`spotter --version` must equal the tag) — that check is the good kind of paranoia.
- `spotter-hook` as a packaged entry point: `integration.is_spotter_hook` already recognizes the `spotter-hook` basename, so this slots into the managed-setup path cleanly.
- `--version` on the CLI, `SHA256SUMS`, `--verify-tag`.

### Minor
- The release workflow builds and version-checks but never runs the test suite before publishing — add `pip install -e .[dev] && pytest` or gate on a CI workflow.
- `docs/releasing.md` says "consumers must not infer protocol compatibility from the release version alone" — correct principle, but it's only true once finding 2 is fixed and the contracts actually come from the modules that own them.

### Verdict
**Not mergeable as-is — needs a rebase onto current main that deletes the daemon stub and deduplicates the entry point, plus the constants fix (2).** The versioning/release direction itself is good and worth landing after that rework.

PR #113, opened 2026-08-13
feat(operations): add runtime-aware diagnostics

### LINKED ISSUE #84: Upgrade status and doctor for runtime capabilities and degraded health
### Problem / outcome

`spotter status` and `spotter doctor` currently diagnose the Hook-era prototype: local storage, Hook registration, synthetic Hook round-trip, reviewer ledger, and journal freshness. The standalone runtime needs diagnostics that can distinguish a healthy daemon from a daemon that has lost observation/control, and a configured integration from a partially migrated one.

### Proposed direction

Extend the existing commands rather than replacing their useful checks.

`status` should summarize current subsystem state, including:

- `spotterd` lifecycle / control-RPC health
- Codex integration mode and manifest state
- App Server connection and ownership/provenance where known
- negotiated observation/control/enforcement capabilities
- active/dormant thread and reviewer-queue counts where available
- journal/storage/reviewer health already reported today
- explicit degraded/unavailable consequences, e.g. observation unavailable while PreToolUse enforcement remains active

`doctor` should perform deeper checks for daemon/IPC, App Server attach, required event visibility, capability negotiation, schema/permissions, integration ownership, and duplicate/stale Hook configuration. Add safe repair hooks only where the correction is deterministic and non-destructive.

### Constraints / tradeoffs

- Depends on the runtime/client/setup surfaces that it diagnoses.
- Preserve the current principle that silence is not health.
- Do not report unsupported optional capabilities as failure when the current integration intentionally degrades without them.
- Avoid destructive `--repair` behavior; ambiguous external config should require a human decision.

### Done when

- `status` clearly distinguishes daemon, observation, live control, enforcement, storage, and reviewer health.
- A running daemon with a disconnected App Server is reported as degraded, not healthy.
- `doctor` validates the managed Codex integration end to end and can identify duplicate/stale legacy integration.
- Existing Hook/storage/ledger diagnostic coverage is preserved or deliberately migrated.
- Exit codes distinguish healthy, degraded/warn, and broken states consistently.
- Tests cover partial subsystem failures rather than only all-healthy/all-dead cases.

YoungJinJung:
## Summary

- make `status` and `doctor` report daemon, integration, observation, control, enforcement, runtime-state, and reviewer-queue health separately
- validate exact Hook/manifest ownership, stale legacy plugins, managed service registration, and configured App Server reachability
- return consistent healthy/degraded/broken exit codes while preserving storage, ledger, freshness, and synthetic Hook diagnostics
- keep unavailable #85 runtime state explicit instead of fabricating thread or queue counts

## Validation

- `.venv/bin/ruff check .`
- `.venv/bin/mypy src tests`
- `.venv/bin/pytest -q`
- `git diff --check`

Closes #84

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,7 +83,7 @@ Legend: ✅ implemented · 🟡 partial/shadow · 🧪 proof required · 🎯 ta
 | Live `VERIFY / NUDGE` | ❌ | Target: `turn/steer` |
 | `INTERRUPT` | ❌ | Target: `turn/interrupt` |
 | `RESTART` | ❌ | Requires verified-state + side-effect-aware recovery |
-| Codex setup lifecycle | 🟡 | Transactional setup/teardown and manifest implemented; packaging and full App Server integration remain |
+| Codex setup lifecycle | 🟡 | Transactional setup/teardown, ownership diagnostics, and degraded capability reporting implemented; packaging and full App Server integration remain |
 
 ### Current runtime work
 
@@ -96,7 +96,9 @@ routing in [#85](https://github.com/spotter-agent/spotter/issues/85). [#79](http
 adds the `spotterd` process, versioned local control handshake, manual lifecycle commands, and a
 platform-neutral service boundary. [#83](https://github.com/spotter-agent/spotter/issues/83) adds
 transactional Codex setup/teardown, managed user-service registration, and integration ownership.
-Event routing and recovery remain separate work.
+[#84](https://github.com/spotter-agent/spotter/issues/84) makes `status` and `doctor` distinguish
+daemon, observation, control, enforcement, integration drift, storage, and reviewer health. Event
+routing and recovery remain separate work.
 
 ---
 
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -685,6 +685,11 @@ Codex integration
 
 This allows a Codex upgrade to degrade one feature without forcing a binary “supported/unsupported” result for the whole integration.
 
+`spotter status` now reports this split without probing external services. `spotter doctor` performs
+the deeper App Server connection/initialize probe when the integration manifest has an endpoint.
+Until #85 wires a persistent runtime consumer, thread counts and reviewer queue state are reported
+as unknown/unavailable instead of being derived from incompatible Hook-era sessions.
+
 ---
 
 # 11. Failure and degraded mode
@@ -712,6 +717,10 @@ PreToolUse gate:      active
 Journal:              writable
 Reviewer:             available
 ```
+
+The aggregate command health is healthy/degraded/broken (`0`/`1`/`2`). In particular, a healthy
+daemon with no App Server observation path is degraded while the independent PreToolUse enforcement
+surface can still be reported active.
 
 ### `spotterd` crash
 
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -718,9 +718,10 @@ Journal:              writable
 Reviewer:             available
 ```
 
-The aggregate command health is healthy/degraded/broken (`0`/`1`/`2`). In particular, a healthy
-daemon with no App Server observation path is degraded while the independent PreToolUse enforcement
-surface can still be reported active.
+The aggregate command health is healthy/degraded/broken (`0`/`1`/`2`). Known unimplemented or
+optional unconfigured surfaces are informational rather than permanent warnings; absence of every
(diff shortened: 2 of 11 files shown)


youngjinjung-linq:
## Review

### Overview
Closes #84 (+492): read-only `check_integration()` (manifest / owned-hook / legacy-plugin / service-registration validation), `check_runtime()` (daemon / observation / control / enforcement consequences), a doctor-only deep App Server probe, and `status` gaining runtime output plus 0/1/2 exit codes. Eight new tests, and CODEX_HOME isolation added to existing fixtures — without which the suite would have read the developer's real `~/.codex`; good catch. Verified the full suite passes (105/105) on the branch.

### Strengths
- `is_spotter_hook` extracted to module level so doctor reuses integration's detection logic — one source of truth instead of a copy.
- Diagnostics are strictly read-only, with the circular import avoided via a commented local import.
- Drift detection is practical: duplicated owned hooks, stale legacy plugins, tampered manifests, missing managed registration — all FAIL, all tested.
- The `enforcement` consequence wording matches #102's fallback semantics exactly (daemon down → WARN with "bounded local fallback", not FAIL).
- `status` never probes; only doctor does the deep probe — the right cost split.

### Findings (one substantive)

**Two perma-WARN checks destroy the information value of the exit codes.** `check_runtime()` unconditionally appends `runtime state` (unknown until #85) and `review queue` (not implemented) as WARN. Since `worst()` aggregates them, **neither `spotter status` nor `spotter doctor` can ever exit 0 until #85 lands.** Doctor's own docstring ("non-zero exit when broken") and the CLI help become false, and a perfectly healthy hook-era install exits 1 forever. For scripts/CI/monitoring, "expected not-yet-implemented" and "newly degraded" become indistinguishable — the whole point of this series' "degraded rather than silently healthy" principle is *distinguishability*, and an always-WARN does the opposite. The test suite testifies to this itself: the only test exercising exit 0 (`test_status_exit_code_matches_runtime_health`) can produce it only by monkeypatching `check_runtime` wholesale. **Suggested fix**: keep reporting the unimplemented capabilities as output lines but exclude them from the verdict (e.g. an `INFO` status ignored by `worst()`, or compute the verdict over checks minus known-unimplemented). A few lines, not a redesign.

**Minor:**
- A user-modified owned hook — the exact state #110's teardown deliberately respects as user property — is FAIL "drifted" → exit 2 "broken" here. FAIL is defensible (the enforcement path may be severed), but note the tension with #110's semantics.
- `_hook_entries` silently skips non-dict hooks while integration's `_read_hooks` raises on the same shape — acceptable tolerance for read-only diagnostics, but asymmetric.
- A fresh install with zero observations exits 1 — same root as the main finding: "nothing has happened yet" vs "degraded" are conflated.

### Verdict
Recommend merging after excluding the perma-WARNs from the verdict (or explicitly deciding to accept never-zero exits and fixing the docstring/help text to match). The diagnostic content itself — ownership validation, drift detection, consequence-first reporting — is correct and well-tested; the only substantive issue is the aggregation semantics, and it's a few lines.

youngjinjung-linq:
## Re-review after 61ed1a02

The substantive finding is resolved cleanly. Verified on the branch: full suite 108/108, `mypy` clean, `ruff` clean.

### Addressed
- **Perma-WARN exit poisoning:** a new `INFO` tier now carries known-unimplemented surfaces (`runtime state`, `review queue`), unconfigured optional runtimes, and not-yet-observed freshness — and `worst()` ignores it by construction (it only tests FAIL/WARN), so the verdict is computed over real signals only. Doctor's "non-zero exit when broken" contract is true again.
- **The proof I asked for exists:** `test_healthy_legacy_hook_without_observations_can_exit_zero` reaches `status` exit 0 end-to-end with no monkeypatching — a real healthy install can now be pronounced healthy.
- **Distinguishability preserved in both directions:** per-runtime absence is INFO, but a new aggregate `runtime registration` check still WARNs when Spotter is wired into *no* runtime — informational tolerance didn't swallow the genuine signal.
- Minors picked up too: the drift FAIL message now says "missing, user-modified, or duplicated … run \`spotter setup codex\` to reconcile", acknowledging the #110 user-modification semantics and giving remediation; the fresh-install conflation is gone (`warned = False`, freshness INFO); legacy-hook-without-manifest reports "legacy Hook local enforcement available" as INFO. Docs updated to match in both lifecycle.md and architecture.md.

### Verdict
Approve — merging.

PR #114, opened 2026-08-13
feat(observation): ingest App Server events into Trace IR

### LINKED ISSUE #85: Ingest App Server events into an identity-rich Trace IR and durable journal

YoungJinJung:
## Summary

- normalize authoritative Codex App Server lifecycle and item families into identity-rich, transport-neutral Trace events
- persist Trace identity, timestamp, operation correlation, and provenance while retaining legacy journal compatibility
- reconcile exact replays, terminal-before-start delivery, timestamp regressions, and conflicting outcomes across restarts
- keep Hook-era sessions explicitly separate and omit raw reasoning content and unknown wire payloads

## Validation

- `.venv/bin/ruff check .`
- `.venv/bin/mypy src tests`
- `.venv/bin/pytest` (332 passed)

Closes #85

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,7 +78,7 @@ Legend: ✅ implemented · 🟡 partial/shadow · 🧪 proof required · 🎯 ta
 | Evaluation labels / metrics | ✅ | Coverage-aware evaluation and precision/FP metrics |
 | Counterfactual experiment harness | ✅ | Control/guidance same-prefix pairs |
 | Standalone `spotterd` runtime | 🟡 | Process, versioned gate/control IPC, and managed user-service startup implemented; runtime consumers remain |
-| App Server primary observation | 🟡 | Client/capabilities implemented; Trace IR ingestion remains (#85) |
+| App Server primary observation | 🟡 | Client, identity-rich Trace IR normalization, and durable ingestion implemented; daemon routing remains |
 | Event-driven signal engine | ❌ | Current reviewer trigger is periodic |
 | Live `VERIFY / NUDGE` | ❌ | Target: `turn/steer` |
 | `INTERRUPT` | ❌ | Target: `turn/interrupt` |
@@ -91,8 +91,11 @@ Legend: ✅ implemented · 🟡 partial/shadow · 🧪 proof required · 🎯 ta
 App Server can share the TUI's real thread and steer its active turn. The production client in
 [#80](https://github.com/spotter-agent/spotter/issues/80) now exposes events, thread queries,
 control methods, and per-capability degraded state. [#81](https://github.com/spotter-agent/spotter/issues/81)
-adds a provisional thread/turn/attachment registry; it remains unconsumed until App Server event
-routing in [#85](https://github.com/spotter-agent/spotter/issues/85). [#79](https://github.com/spotter-agent/spotter/issues/79)
+adds a thread/turn/attachment registry. [#85](https://github.com/spotter-agent/spotter/issues/85)
+uses it to normalize authoritative lifecycle, message, plan, reasoning-summary, command, file-change,
+MCP, search, diff, and token events into durable Trace IR. Exact replays are deduplicated across
+restarts, operation outcomes correlate by item ID, and unknown notification families remain explicit.
+[#79](https://github.com/spotter-agent/spotter/issues/79)
 adds the `spotterd` process, versioned local control handshake, manual lifecycle commands, and a
 platform-neutral service boundary. [#83](https://github.com/spotter-agent/spotter/issues/83) adds
 transactional Codex setup/teardown, managed user-service registration, and integration ownership.
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -1,7 +1,7 @@
 # Architecture
 
 > **Status:** this document describes both the current hook-based prototype and the target architecture. Active implementation is tracked by the [Roadmap](roadmap.md) and native GitHub Milestones.
-> The `spotterd` process/control foundation, bounded Hook gate IPC, shared-server PoC, and production App Server transport are implemented. App Server primary Trace IR observation and live supervision delivery remain **target behavior**.
+> The `spotterd` process/control foundation, bounded Hook gate IPC, shared-server PoC, production App Server transport, and durable Trace IR ingestion are implemented. Daemon event routing and live supervision delivery remain **target behavior**.
 
 ---
 
@@ -467,6 +467,23 @@ side-effect class
 
 Normalization must not discard provenance. Live control, replay, and causal analysis must be able to reconnect normalized records to the underlying runtime item/turn.
 
+## Implemented App Server ingestion
+
+`CodexTraceNormalizer` is the transport boundary. It maps App Server lifecycle and authoritative
+completed-item records into the runtime-neutral fields above. Raw reasoning content is not copied;
+only the App Server's exposed reasoning summary enters Trace IR. Unknown notification methods become
+`runtime_event_unknown` records containing the method and any proven identity, rather than silently
+disappearing or leaking a new wire shape into core consumers.
+
+`AppServerTraceIngestor` writes one journal per Spotter thread identity, separately named from
+Hook-era session journals. Its recovery scan rebuilds deduplication and lifecycle state after a
+restart. Item starts and outcomes correlate by App Server item ID, not journal adjacency. A terminal
+event seen before its start is retained with `observed_start=false`; a later start cannot regress it.
+Timestamp regressions are accepted but marked `out_of_order`, and conflicting terminal outcomes fail
+explicitly. Hook records carry legacy-session provenance with unknown thread, turn, and attachment
+dimensions, so consumers never need Codex transport objects and never infer that a Hook session is an
+App Server thread.
+
 ---
 
 # 7. Reviewer jobs and freshness
@@ -647,9 +664,9 @@ State scope:
 - Hook-era `session_id` becomes a `RuntimeIdentity` with unknown thread/turn/attachment fields rather
   than being promoted into an identity it cannot prove.
 
-The registry owns identity and lifecycle only and has no production consumer yet. Semantic
-`ThreadState` remains #31; App Server event normalization/routing in #85 must validate and may
-reshape this interface.
+The registry owns identity and lifecycle only; App Server Trace normalization and journal recovery
+consume it without promoting it into semantic state. Semantic `ThreadState` remains #31, and daemon
+connection reconciliation remains #87.
 
 ---
 
--- a/src/spotter/hook.py
+++ b/src/spotter/hook.py
@@ -13,6 +13,7 @@ import subprocess
 import sys
 import time
 from contextlib import suppress
+from dataclasses import replace
 from pathlib import Path
 from typing import Any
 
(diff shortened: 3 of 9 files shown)


YoungJinJung:
## Review

### Overview
Closes #85 (+918): `CodexTraceNormalizer` (the only layer allowed to know Codex wire shapes) + `AppServerTraceIngestor` (per-thread journals, restart-safe dedup/lifecycle reconciliation), `TraceEvent` gaining identity/provenance/timing fields with backward-compatible journal serialization, and Hook events stamped with explicit legacy identity. This finally consumes the #81 registry in production — the sequencing objection from the #94 review is now fully discharged.

### Strengths
- **Privacy discipline is enforced by test**: raw reasoning `content` never enters Trace IR — only the App Server's exposed summary (`test` asserts `"content" not in payload`).
- **`_known` whitelisting** keeps unstable wire shapes out of IR, with `futureField: must not leak` pinned by test; unknown methods become `runtime_event_unknown` with proven identity instead of disappearing.
- Operation correlation by item ID rather than journal adjacency, tested with an interleaved agent message; per-turn ordering scope, so a reused item ID in another turn isn't conflated.
- Journal round-trip is backward compatible: records without `trace` metadata load as identity-less events.

### Findings

**1. Blocker — `_recover()` uses `strict=True`, which refuses torn tails: a crash mid-append makes the ingestor unconstructable.** `StepJournal.load(strict=True)` raises `SnapshotError` on exactly the artifact this class's own crash leaves behind (its docstring: "strict=True refuses any undecodable content … destructive readers (prune) must not treat a torn tail as proof of absence"). The ingestor is not a destructive reader — it is the journal's *owner/writer*, and its recovery scan is the path that runs right after the crash that produces the torn tail. As written, one power-loss mid-write and `AppServerTraceIngestor.__init__` raises forever until someone hand-edits the journal — the exact opposite of the class's stated "restart-safe" purpose and of the journal's own crash-tolerance contract. `repair_tail=True` (owner semantics) or prefix-tolerant loading is the fix.

**2. Content-hash dedup conflates "exact replay" with "legitimate recurrence".** `_event_id = sha256(method + params)`, and `ingest` drops any seen hash. Notification families with no timestamp and no unique ID in params — `thread/status/changed`, `turn/diff/updated`, `plan` — produce identical params on legitimate repeats: a thread going idle→active→idle→active loses the second `active` transition silently. Item lifecycle events are safe (unique IDs); the blanket hash isn't. Either scope hash-dedup to methods that carry intrinsic identity, or accept and document the loss for the stateless families.

**3. Unbounded state, unbounded recovery.** `_seen`/`_operations`/`_last_at`/`_terminal_turns` grow for the life of the process, and `_recover()` re-reads *every* journal in full at startup — O(all history) memory and startup time in the long-lived daemon this is destined for. Same class of ceiling flagged on #94; wants at least a known-ceiling comment naming the upgrade path (per-thread retention, snapshot of recovery state) before #87 wires it into the daemon loop.

**Minor:**
- The journal record gains a `trace` field without a `SCHEMA_VERSION` bump. Fine if the version contract is breaking-changes-only, but that contract deserves stating once — #71 made schema versioning first-class.
- `ingestion.py` imports `spotter.app_server` (hence `websockets`) for the two-field `AppServerEvent`. Not on the Hook path today — keep it that way, or move the dataclass, given #113 deliberately kept websockets off the bundled Hook import path.

### Verdict
**Hold merge for finding 1** — it inverts the PR's central promise (restart safety) in the most common crash scenario, and the fix is one argument. Finding 2 needs a decision (scope or document); 3–5 can follow. Everything else here is strong work — the normalizer boundary, privacy discipline, and correlation semantics are exactly what #85 called for.

YoungJinJung:
## Re-review after 8cd51688

Both substantive findings are resolved, each with a targeted regression test. Verified on the branch: full suite passes (334), `mypy src tests` clean, `ruff check` clean.

### Addressed
- **Torn-tail blocker:** `_recover()` now loads with `repair_tail=True` — owner semantics, exactly the fix asked for. `test_recovery_repairs_a_torn_journal_tail` writes a literal truncated line, restarts the ingestor, and proves the journal is repaired and ingestion continues. The restart-safety promise now survives the crash that produces the artifact.
- **Replay-vs-recurrence conflation:** `_event_id` returns `None` for any method outside `_REPLAY_SAFE_METHODS` (lifecycle + item events, which carry intrinsic identity), so stateless families like `thread/status/changed` are never hash-deduped. `test_identical_stateless_notifications_are_not_mistaken_for_replays` pins two identical status events both landing in the journal. Scoping the dedup instead of documenting the loss was the better of the two options I offered.
- **Unbounded recovery ceiling:** now carries a `ponytail:` comment naming the O(all-history) scan and the #87 upgrade path (per-thread checkpoint + retention).
- Tightened along the way: a second `item/started` for an already-started operation is now dropped (`previous is not None`), giving one start per operation rather than only started-after-completed.

### Verdict
Approve — merging. The two remaining minors from the first review (additive `trace` field without a `SCHEMA_VERSION` bump; `ingestion` importing `app_server`/websockets for a two-field dataclass) stand as noted, both fine to ride along to #87.

Issue #28, opened 2026-08-11
Replace periodic review with event-driven signals


YoungJinJung:
### Problem / outcome

The current prototype invokes the reviewer on a fixed cadence. The Detect stage needs cheap evidence-driven candidate signals so model review is targeted, cheaper, and earlier.

### Proposed direction

Add a `SignalEngine` over live App Server/Trace IR state. Signals emit candidate hypotheses with evidence; they never emit semantic verdicts.

Initial candidates should cover:

- repeated equivalent tool calls
- failure streaks
- repeated reads with no frontier expansion
- sudden touched-scope growth
- edits accumulating without validation
- stale hypothesis reuse
- repeated behavior after a deterministic block
- tool/token/time budget anomalies

Journal candidates and make them labelable. Feed the triggering candidate plus relevant constraints/state and available probes into the reviewer. Keep any periodic cadence only as an explicit fallback.

### Implementation references — study, do not depend on

**ECLoop** is directly relevant to the shape of cheap runtime conditions. It compiles explicit evidence requirements from the task/repository, tracks whether the trajectory has satisfied them, and postpones unsupported commitment. The important lesson is that structured state/conditions outperform asking a model to repeatedly rediscover the same evidence requirements from prose.

- paper: https://arxiv.org/abs/2607.28815

**Shepherd's live-intervention experiment** is useful as the closest implemented meta-agent supervisor loop. It currently drives a supervisor on an interval/effect-stream basis, which is useful as a baseline and as a source of intervention/evaluation patterns, but Spotter should specifically improve on the cadence cost with event-driven candidate generation.

- experiment: https://github.com/shepherd-agents/shepherd-experiments/tree/main/exp/live-intervention
- meta-agent code: https://github.com/shepherd-agents/shepherd-experiments/blob/main/exp/live-intervention/src/experiment_live_intervention/core/meta_agent_llm.py
- experiment configuration/batch runner: https://github.com/shepherd-agents/shepherd-experiments/blob/main/exp/live-intervention/src/experiment_live_intervention/core/run_batch_e2b.py

**NeMo Relay** also demonstrates a useful principle in its best-effort skill-load detection: bounded parsers can recognize known wrapper shapes, but absence of a signal must not be interpreted as proof that the underlying action did not occur.

- Codex captured-event / best-effort detection notes: https://github.com/NVIDIA/NeMo-Relay/blob/main/docs/nemo-relay-cli/codex.mdx

[...]

PR #115, opened 2026-08-13
feat(runtime): add daemon-owned ThreadState

### LINKED ISSUE #31: Move independent supervision state into spotterd

YoungJinJung:
## Why

App Server Trace IR needs one long-lived, thread-isolated owner of judgment-ready state. Rebuilding audit context from the journal on every Hook cannot support incremental signals or freshness-safe reviewer snapshots.

## What changed

- add a deterministic reducer for immutable, versioned task/evidence/workspace/execution/supervision/coverage state
- keep hypotheses, observations, verified facts, summaries, constraints, and interventions provenance-distinct
- make duplicate IDs idempotent and lifecycle gaps, ordering conflicts, unknown events, and evidence invalidation explicit
- let `spotterd` own per-thread snapshots and conservatively hydrate durable Trace history without restoring stale control readiness
- persist connection epochs in Trace journals for later reconnect reconciliation

## Validation

- `.venv/bin/ruff check .`
- `.venv/bin/mypy src tests`
- `.venv/bin/pytest` (347 passed)

Closes #31

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,8 +77,8 @@ Legend: ✅ implemented · 🟡 partial/shadow · 🧪 proof required · 🎯 ta
 | Claim/evidence audit ledger | 🟡 | Works where observable outcomes exist |
 | Evaluation labels / metrics | ✅ | Coverage-aware evaluation and precision/FP metrics |
 | Counterfactual experiment harness | ✅ | Control/guidance same-prefix pairs |
-| Standalone `spotterd` runtime | 🟡 | Process, versioned gate/control IPC, and managed user-service startup implemented; runtime consumers remain |
-| App Server primary observation | 🟡 | Client, identity-rich Trace IR normalization, and durable ingestion implemented; daemon routing remains |
+| Standalone `spotterd` runtime | 🟡 | Process, gate/control IPC, managed startup, and per-thread immutable live state implemented; App Server connection ownership remains |
+| App Server primary observation | 🟡 | Client, identity-rich Trace IR, durable ingestion, and incremental ThreadState reducer implemented; daemon connection routing remains |
 | Event-driven signal engine | ❌ | Current reviewer trigger is periodic |
 | Live `VERIFY / NUDGE` | ❌ | Target: `turn/steer` |
 | `INTERRUPT` | ❌ | Target: `turn/interrupt` |
@@ -95,6 +95,9 @@ adds a thread/turn/attachment registry. [#85](https://github.com/spotter-agent/s
 uses it to normalize authoritative lifecycle, message, plan, reasoning-summary, command, file-change,
 MCP, search, diff, and token events into durable Trace IR. Exact replays are deduplicated across
 restarts, operation outcomes correlate by item ID, and unknown notification families remain explicit.
+[#31](https://github.com/spotter-agent/spotter/issues/31) reduces that Trace IR into isolated,
+immutable, versioned ThreadState snapshots owned by `spotterd`; restart hydration deliberately
+requires live turn reconciliation before control becomes ready again.
 [#79](https://github.com/spotter-agent/spotter/issues/79)
 adds the `spotterd` process, versioned local control handshake, manual lifecycle commands, and a
 platform-neutral service boundary. [#83](https://github.com/spotter-agent/spotter/issues/83) adds
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -115,8 +115,8 @@ concurrent clients, reports `healthy` / `degraded` / `recovering`, and makes abs
 
 Manual process management implements the `ServiceManager` boundary used by the CLI. Managed
 `launchd` / `systemd --user` registration belongs to setup lifecycle work. This daemon currently
-owns no App Server, semantic `ThreadState`, Hook IPC, or event routing, so stopping it cannot stop a
-shared Codex App Server and existing Hook behavior remains independent.
+owns semantic `ThreadState` and bounded Hook gate IPC, but no App Server connection or event routing,
+so stopping it cannot stop a shared Codex App Server and existing Hook behavior remains independent.
 
 ## 1.3 Target runtime
 
@@ -582,6 +582,17 @@ ThreadState
   connection/capability state
 ```
 
+The implemented `ThreadStateStore` is the daemon-owned, single-event-loop hot-state boundary.
+`ThreadStateReducer` consumes only normalized `TraceEvent` values and returns deeply immutable,
+versioned snapshots. It keeps task, evidence/verification, workspace, execution, supervision, and
+coverage as distinct typed sub-state; reasoning summaries remain hypotheses, and verified facts can
+only be created by an explicit verification-satisfied event. Duplicate event IDs are idempotent,
+out-of-order lifecycle is recorded as partial coverage, and thread identities remain isolated.
+
+Journal hydration deterministically replays Trace IR but clears active-turn/control readiness. A
+recovered daemon must receive a live `runtime_reconciled` event before control can be considered
+safe; #87 owns that connection and reconciliation loop.
+
 ## 8.2 Durable journal
 
 The journal exists for:
@@ -664,8 +675,8 @@ State scope:
 - Hook-era `session_id` becomes a `RuntimeIdentity` with unknown thread/turn/attachment fields rather
   than being promoted into an identity it cannot prove.
 
-The registry owns identity and lifecycle only; App Server Trace normalization and journal recovery
-consume it without promoting it into semantic state. Semantic `ThreadState` remains #31, and daemon
+The registry owns identity and lifecycle only. App Server Trace normalization and the daemon's
+semantic `ThreadState` consume those identities without importing Codex wire shapes. Daemon
 connection reconciliation remains #87.
 
 ---
--- a/docs/status.md
+++ b/docs/status.md
@@ -7,7 +7,7 @@
 
 ## 30-second summary
 
-Spotter is currently a **Hook-based research prototype** with a standalone runtime foundation. It already has real trajectory journals, daemon-backed deterministic gates, Git snapshots, fork/replay, a shadow reviewer, an audit ledger, labeling/metrics, a counterfactual experiment harness, and a manually managed `spotterd` process with versioned local IPC.
+Spotter is currently a **Hook-based research prototype** with a standalone runtime foundation. It already has real trajectory journals, daemon-backed deterministic gates, Git snapshots, fork/replay, a shadow reviewer, an audit ledger, labeling/metrics, a counterfactual experiment harness, identity-rich App Server ingestion, and daemon-owned per-thread live state.
 
 The target product is a standalone runtime:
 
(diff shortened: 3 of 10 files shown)


YoungJinJung:
## Review

### Overview
Closes #31 (+1130): `thread_state.py` (709 lines) — a pure `ThreadStateReducer` over normalized Trace IR producing deeply immutable, versioned snapshots (task / evidence / workspace / execution / supervision / coverage), a single-owner `ThreadStateStore` mounted on `DaemonServer`, `connection_epoch` added to `TraceEvent` and journal round-trip, and 312 lines of tests. Verified on the branch: full suite 347/347, `mypy` clean, `ruff` clean.

### Strengths
- **Epistemic discipline is structural, not aspirational**: reasoning summaries can only ever be `HYPOTHESIS`; `VERIFIED_FACT` requires an explicit `verification_satisfied` referencing live observation evidence; invalidation cascades transitively through `depends_on`/`evidence_ids` and downgrades satisfied conditions — all tested.
- **Conservative hydration is the headline behavior and it's pinned twice**: replay reproduces identical state, but `hydrate()` clears `active_turn_id`/`control_ready`, so a restarted daemon cannot steer until a live `runtime_reconciled` arrives (#87's job). Epoch changes and observation gaps also drop control readiness.
- Inconsistencies are recorded (`coverage.inconsistencies`, `PARTIAL` history) instead of silently repaired — late starts for completed turns, unknown conditions, out-of-order timestamps all leave visible marks.
- Deep immutability enforced by test (`FrozenInstanceError`), thread isolation including a parent-child non-propagation case, and both known ceilings carry `ponytail:` comments.

### Findings (none blocking — nothing feeds this in production until #87)
**1. A large slice of the reducer's semantic surface has no producer.** The #85 normalizer emits none of: `constraint`, `hypothesis`, `test_result`, `verification_condition`, `verification_satisfied`, `evidence_invalidated`, `observation_gap`, `intervention` — so verified facts, validation status, and the whole condition lifecycle are currently unreachable outside tests. `runtime_reconciled` is explicitly #87's; the rest belong to future signal/reviewer work (#28/#22) and will get reshaped when real producers exist. Same pattern flagged on #94 — mitigated here by honest docs, but a one-line comment marking which kinds are live vs reserved would keep the next reader oriented.

**2. `applied_event_ids` lives in the immutable snapshot and is copied per event.** `applied | {event.event_id}` rebuilds a frozenset every event — O(n) copy per event, O(n²) per thread lifetime, and every retained snapshot pins its own copy. Dedup is a *store-level* concern, not a snapshot property (the `repr=False` already hints it doesn't belong there): moving the seen-set into `ThreadStateStore` as one mutable set per thread fixes both the quadratic churn and the snapshot weight, without touching reducer purity from the caller's perspective. The existing `ponytail:` comment covers lifetime retention but not this copy cost. Cheapest to fix now, while nothing consumes the API.

**3. The 50-item evidence window silently ages out verification targets.** `evidence.items` is both a "recent" ring buffer (`_bounded`) and the identity store that `verification_satisfied`/`evidence_invalidated` look up into — evidence older than 50 items can no longer satisfy a condition and lands in `unknown_verification_evidence`. To the system's credit that's *visible* (inconsistency, not silence), but discarded-by-retention is not the same thing as inconsistent input. When #28 produces real conditions, either exempt referenced evidence from the bound or give verification its own keyed map.

**Minor:** `_semantic` applies `_workspace` twice for outcome events (top-of-function and again in the outcome branch) — idempotent unions make it harmless, but it's the kind of redundancy that hides a future bug. `connection_epoch` is another additive journal field riding without a `SCHEMA_VERSION` bump, same standing note as #114.

### Verdict
Approve — merging. Findings 2 and 3 should be addressed before #87 routes live traffic; they're both cheaper to fix now than after the daemon loop depends on snapshot identity.

Issue #89, opened 2026-08-12
Add repository-aware purge, retention, uninstall, and reinstall lifecycle


Bogyie:
### Problem / outcome

The current `prune` command already handles local snapshot refs, stale journals, and fork worktrees for a known repository. The standalone product lifecycle needs a broader ownership-aware cleanup path across durable user data, integration metadata, logs, runtime files, and Git resources in every repository Spotter has touched.

Extend rather than duplicate the existing housekeeping machinery.

### Proposed direction

- maintain the minimum repository registry/provenance required to locate Spotter-owned refs/worktrees outside the current working directory
- define retention policies for journals, logs, snapshots, experiment/fork artifacts, and other runtime-generated data, preserving lineage still referenced by evaluation results
- add explicit `spotter purge` scopes such as data / snapshots / all with dry-run or preview for destructive repository cleanup
- keep package uninstall separate from user-data deletion
- make dangling integration/service/Hook state fail open if a user runs the package-manager uninstall without first running teardown
- support reinstall with existing compatible data and route incompatible persisted state through #47's migration/refusal contract
- use Git-aware cleanup for refs/worktrees; do not remove administrative state with raw filesystem deletion

### Implementation references — study, do not depend on

NVIDIA NeMo Relay has already solved several lifecycle failure cases that are directly relevant even though Spotter will not depend on Relay:

- persistent install / rollback / uninstall lifecycle: https://github.com/NVIDIA/NeMo-Relay/blob/main/docs/nemo-relay-cli/plugin-installation.mdx
- Codex-specific generation fencing and stale-process handling: https://github.com/NVIDIA/NeMo-Relay/blob/main/docs/nemo-relay-cli/codex.mdx

Useful patterns from Relay:

- serialize install / force-install / rollback / uninstall for one user+host;
- give generated integration state an immutable **generation identity**;
- do not let a cached host process silently adopt a replacement installation;
- verify exact ownership before removing Hook/trust/provider/integration state;
- treat a process that can outlive registration as a separate lifecycle object;
- when ownership/generation is ambiguous, **refuse destructive cleanup** and provide a recovery path instead of guessing;
- uninstall exact owned state while preserving unrelated user configuration;
- doctor detects modified/missing/duplicated owned integration state;
- acknowledge that exactly-once cleanup/delivery cannot be guaranteed across every crash boundary and make ambiguous state diagnosable.

These are design references only. Spotter should implement the lifecycle in its existing Python/service/integration boundaries.

[...]

Issue #22, opened 2026-08-11
Deliver live VERIFY/NUDGE via turn/steer


YoungJinJung:
### Problem / outcome

Reviewer decisions currently stop at the journal. The Intervene stage needs `VERIFY` and `NUDGE` to reach the **correct active Codex turn** while Main continues running, without reviving the old Hook-centric `additionalContext` design.

### Proposed direction

Implement live soft intervention on the standalone runtime boundary:

- anchor reviewer jobs to `thread_id` + `target_turn_id`
- deliver `VERIFY` / `NUDGE` through the shared App Server's `turn/steer`
- model job states explicitly: `QUEUED → RUNNING → DECIDED → DELIVERED | STALE | DISCARDED | FAILED`
- define freshness/stale policy when the target turn has already ended
- prevent duplicate delivery and journal delivery provenance/latency
- make ignored interventions measurable
- keep supervision provenance explicit: user intent remains authoritative; Spotter guidance asks Main to re-evaluate with evidence rather than comply blindly
- treat **steer history contamination / task-ownership drift** as a first-class product risk, because current Codex steering is persisted/model-visible as user input rather than a separate supervisor role

#### OpenClaw — production App Server control implementation

OpenClaw already implements the relevant Codex App Server control primitive and is the best current code reference for this issue:

- active-turn tracking + `turn/steer` / `turn/interrupt` dispatch: https://github.com/openclaw/openclaw/blob/main/extensions/codex/src/conversation-control.ts
- supervision/session control surface: https://github.com/openclaw/openclaw/blob/main/extensions/codex/src/supervision-tools.ts
- Codex App Server event projection: https://github.com/openclaw/openclaw/blob/main/extensions/codex/src/app-server/event-projector.ts
- runtime boundary notes: https://github.com/openclaw/openclaw/blob/main/docs/plugins/codex-harness-runtime.md

Important OpenClaw failure cases should be treated as Spotter test cases, not incidental OpenClaw bugs:

- RPC acceptance is not durable/idempotent adoption: https://github.com/openclaw/openclaw/issues/103920
- a steered message can be persisted but never consumed if the target turn ends in the enqueue/consume window: https://github.com/openclaw/openclaw/issues/92433
- queue serialization can make a nominally live steer arrive only after the active turn is effectively over: https://github.com/openclaw/openclaw/issues/48003
- late steering after a terminal assistant answer can perturb final-output selection: https://github.com/openclaw/openclaw/issues/112468
[...]

PR #118, opened 2026-08-13
feat(runtime): recover App Server connections safely

### LINKED ISSUE #87: Recover from daemon and App Server disconnects without losing thread identity

YoungJinJung:
## Why

A connected socket is not enough to safely preserve live thread/turn identity across App Server disconnects or daemon restarts. Spotter needs a single recovery owner that makes observation gaps and stale control targets explicit while keeping Hook enforcement independent.

## What changed

- add a daemon-owned `DISCONNECTED → CONNECTING → RECONCILING → READY` loop with bounded jittered backoff and one cleanup path
- assign monotonically recovered connection epochs, per-connection attachment IDs, endpoint/server fingerprints, and capability snapshots
- hydrate durable ThreadState conservatively, reconcile every loaded thread, journal explicit observation gaps, and mark missing attachments unavailable
- fence steer/interrupt by the exact reconciled epoch and active turn; late connection callbacks and cross-epoch operation correlations cannot retarget current work
- move Trace event deduplication out of immutable snapshots and keep old evidence addressable for later verification, addressing the two follow-ups from PR #115
- update current-state, architecture, and lifecycle documentation while leaving endpoint selection/ownership explicitly out of scope

## Validation

- `.venv/bin/ruff check .`
- `.venv/bin/mypy src tests`
- `.venv/bin/pytest` (353 passed)

Closes #87

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,8 +55,8 @@ External Codex App Server
 ```
 
 [#78](https://github.com/spotter-agent/spotter/issues/78) proved shared observation and steering for a
-Spotter-managed external App Server. Current Runtime work turns that viable control boundary into
-the daemon, identity, lifecycle, and recovery components required for ordinary use.
+Spotter-managed external App Server. The daemon now owns connection epochs, reconnect/reconciliation,
+and conservative control freshness; endpoint selection in setup remains the last ordinary-use gap.
 
 For the fastest project snapshot, read [Status](docs/status.md). For sequence and evidence gates, read [Roadmap](docs/roadmap.md).
 
@@ -77,8 +77,8 @@ Legend: ✅ implemented · 🟡 partial/shadow · 🧪 proof required · 🎯 ta
 | Claim/evidence audit ledger | 🟡 | Works where observable outcomes exist |
 | Evaluation labels / metrics | ✅ | Coverage-aware evaluation and precision/FP metrics |
 | Counterfactual experiment harness | ✅ | Control/guidance same-prefix pairs |
-| Standalone `spotterd` runtime | 🟡 | Process, gate/control IPC, managed startup, and per-thread immutable live state implemented; App Server connection ownership remains |
-| App Server primary observation | 🟡 | Client, identity-rich Trace IR, durable ingestion, and incremental ThreadState reducer implemented; daemon connection routing remains |
+| Standalone `spotterd` runtime | 🟡 | Process, gate IPC, per-thread immutable state, and App Server recovery ownership implemented; setup endpoint selection remains |
+| App Server primary observation | 🟡 | Configured endpoints route through daemon-owned epoch/reconciliation into durable Trace IR and ThreadState; setup does not select an endpoint yet |
 | Event-driven signal engine | ❌ | Current reviewer trigger is periodic |
 | Live `VERIFY / NUDGE` | ❌ | Target: `turn/steer` |
 | `INTERRUPT` | ❌ | Target: `turn/interrupt` |
@@ -97,14 +97,15 @@ MCP, search, diff, and token events into durable Trace IR. Exact replays are ded
 restarts, operation outcomes correlate by item ID, and unknown notification families remain explicit.
 [#31](https://github.com/spotter-agent/spotter/issues/31) reduces that Trace IR into isolated,
 immutable, versioned ThreadState snapshots owned by `spotterd`; restart hydration deliberately
-requires live turn reconciliation before control becomes ready again.
+requires live turn reconciliation before control becomes ready again. [#87](https://github.com/spotter-agent/spotter/issues/87)
+adds the single-owner connection loop, bounded reconnect backoff, durable observation gaps,
+connection/attachment epochs, multi-thread reconciliation, and exact epoch+turn control fencing.
 [#79](https://github.com/spotter-agent/spotter/issues/79)
 adds the `spotterd` process, versioned local control handshake, manual lifecycle commands, and a
 platform-neutral service boundary. [#83](https://github.com/spotter-agent/spotter/issues/83) adds
 transactional Codex setup/teardown, managed user-service registration, and integration ownership.
 [#84](https://github.com/spotter-agent/spotter/issues/84) makes `status` and `doctor` distinguish
-daemon, observation, control, enforcement, integration drift, storage, and reviewer health. Event
-routing and recovery remain separate work.
+daemon, observation, control, enforcement, integration drift, storage, and reviewer health.
 
 ---
 
@@ -330,8 +331,9 @@ spotter setup codex
 spotter teardown codex
 ```
 
-App Server endpoint selection remains pending until the remaining ingestion/lifecycle work lands;
-setup does not print or claim an endpoint that it has not verified.
+App Server endpoint selection remains pending; when a verified endpoint is present in the integration
+manifest, `spotterd` owns its connection and recovery loop. Setup does not print or claim an endpoint
+that it has not verified.
 
 Useful commands:
 
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -1,7 +1,7 @@
 # Architecture
 
 > **Status:** this document describes both the current hook-based prototype and the target architecture. Active implementation is tracked by the [Roadmap](roadmap.md) and native GitHub Milestones.
-> The `spotterd` process/control foundation, bounded Hook gate IPC, shared-server PoC, production App Server transport, and durable Trace IR ingestion are implemented. Daemon event routing and live supervision delivery remain **target behavior**.
+> The `spotterd` process/control foundation, bounded Hook gate IPC, shared-server PoC, production App Server transport, durable Trace IR ingestion, and daemon reconnect/reconciliation are implemented. Live supervision delivery remains **target behavior**.
 
 ---
 
@@ -113,10 +113,10 @@ handshake rather than treating a PID file as liveness. The server serializes own
 concurrent clients, reports `healthy` / `degraded` / `recovering`, and makes absence explicit as
 `unavailable`.
 
-Manual process management implements the `ServiceManager` boundary used by the CLI. Managed
-`launchd` / `systemd --user` registration belongs to setup lifecycle work. This daemon currently
-owns semantic `ThreadState` and bounded Hook gate IPC, but no App Server connection or event routing,
-so stopping it cannot stop a shared Codex App Server and existing Hook behavior remains independent.
+Manual process management implements the `ServiceManager` boundary used by the CLI. For a configured
+endpoint, the daemon owns one App Server connection/reconnect loop, routes epoch-tagged Trace IR into
+durable journals and `ThreadState`, and reconciles loaded threads before reporting control-ready.
+It never owns or stops the shared App Server process; existing Hook enforcement remains independent.
 
 ## 1.3 Target runtime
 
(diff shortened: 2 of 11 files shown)


YoungJinJung:
## Review

### Overview
Closes #87 (+963): `runtime_connection.py` (483 lines) — a single-owner `AppServerRecoveryLoop` (`DISCONNECTED → CONNECTING → RECONCILING → READY` with `DEGRADED`/`BACKING_OFF`), monotonic connection epochs persisted through journals, per-connection attachment IDs, endpoint/server fingerprints, durable `observation_gap`/`runtime_reconciled`/`runtime_attachment_unavailable` records, exact epoch+turn control fencing, and epoch-scoped ingestion correlation. Verified on the branch: full suite 353/353, `mypy` clean, `ruff` clean.

### Both #115 follow-ups landed, properly
- **Dedup moved out of snapshots**: `applied_event_ids` is gone from `ThreadState`; exact dedup now lives in `ThreadStateStore._seen_event_ids` as a mutable per-thread set — kills the O(n) frozenset copy per event and the snapshot weight, with the retention ponytail comment carried over to #89.
- **Evidence stays addressable**: `_bounded` removed from `evidence.items` appends, and `test_old_evidence_remains_addressable_for_later_verification` proves a proof older than 51 outcomes still satisfies a condition. The unbounded-evidence ceiling is documented on `EvidenceState` with the right caveat (compaction only after persisted conditions share the boundary).
- The reserved-vs-live event-kind comment from finding 1 landed too, and `runtime_reconciled`/`observation_gap` now have a real producer.

### Strengths
- **The fencing is genuinely conservative**: `_validate_target` requires READY + exact connection epoch + `control_ready` + exact active turn, and the reconnect test proves a pre-disconnect target raises `StaleControlTarget` afterward. `runtime_reconciled` with `active_turn: false` now explicitly clears the active turn instead of preserving stale state.
- Epochs are monotonic **across daemon restarts** (seeded from `ingestor.last_connection_epoch`), tested; operation correlation and ordering keys are epoch-scoped so cross-epoch item IDs can't be conflated, tested.
- Coverage honesty: a first-ever attachment is `PARTIAL` history with a `new_attachment`-sourced gap — "we attached mid-life" is not pretended to be complete observation.
- Teardown ordering via `AsyncExitStack` is right: consumer cancelled before disconnect (LIFO), which also neutralizes the known `next_event()`-blocks-forever sharp edge from the #93 review.
- `websockets` stays off the Hook import path (lazy import in `DaemonServer.start`, only when an endpoint is configured — which setup still doesn't emit, and the docs say so).

### Findings (non-blocking)
**1. A dead consumer leaves the loop READY but blind.** `_consume` catches `IngestionError` (keeps going) and `AppServerError` (returns), but an unanticipated exception — e.g. a `ThreadStateError` escaping `_record`, or any future reducer bug — kills the consumer task silently: the connection stays up, `wait_closed` keeps pending, state stays READY, events pile into the client queue, and nothing is ingested. `_cancel_task`'s `suppress(..., Exception)` then swallows the real traceback at teardown. Low reachability today (the reachable reducer raises are store-controlled), but the failure class is exactly what this project calls "silently healthy." Cheap fix: a broad `except Exception` in `_consume` that records `last_error` and forces a reconnect (or at least DEGRADED), plus not suppressing non-CancelledError in `_cancel_task`.
**2. Minor:** `DaemonServer.recovery: Any` and the string-typed `_on_recovery_state` mapping are looser than the rest of the codebase — understandable for the import-path constraint, but a `Protocol` would keep mypy honest for free.

### Verdict
Approve — merging. Finding 1 deserves a small hardening commit before live steering work (#22) leans on READY; it doesn't block this PR because the reachable paths are already handled and everything else here is the strongest recovery semantics in the series.

PR #119, opened 2026-08-13
fix(runtime): reconnect after consumer failure

YoungJinJung:
## Why

PR #118 review identified a failure mode where an unexpected event-consumer exception could leave the App Server loop reporting READY while no events were being ingested.

## What changed

- turn unexpected consumer exceptions into an explicit DEGRADED transition and force the connection through the reconnect loop
- preserve the consumer root exception through `AsyncExitStack` cleanup instead of suppressing it
- replace the daemon recovery `Any` boundary with a small Protocol and type-only RecoveryState import without loading websockets on the Hook path
- add a regression test that injects a reducer failure after READY and proves degraded/backoff state plus failure telemetry

## Validation

- `.venv/bin/ruff check .`
- `.venv/bin/mypy src tests`
- `.venv/bin/pytest` (354 passed)

Follow-up to #118

### DIFF
--- a/src/spotter/daemon.py
+++ b/src/spotter/daemon.py
@@ -17,7 +17,7 @@ from enum import StrEnum
 from fcntl import LOCK_EX, LOCK_NB, LOCK_UN, flock
 from io import TextIOWrapper
 from pathlib import Path
-from typing import Any, Protocol, cast
+from typing import TYPE_CHECKING, Any, Protocol, cast
 
 from spotter.config import GatesConfig
 from spotter.gates import Gate, GateDecision
@@ -26,6 +26,9 @@ from spotter.paths import secure_dir, spotter_home
 from spotter.snapshot import StepRecord
 from spotter.thread_state import ThreadState, ThreadStateStore
 from spotter.trace import TraceEvent
+
+if TYPE_CHECKING:
+    from spotter.runtime_connection import RecoveryState
 
 PROTOCOL_VERSION = 1
 CONTROL_TIMEOUT = 1.0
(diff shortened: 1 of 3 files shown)


YoungJinJung:
## Review

Follow-up to #118, addressing both findings from that review (+70/−5). Verified on the branch: full suite 354/354, `mypy` clean, `ruff` clean.

### Addressed
- **READY-but-blind consumer death (was #118 finding 1):** `_consume` now catches unexpected exceptions, records `last_error`, transitions DEGRADED, disconnects the client — which unblocks `wait_closed` in the main loop and forces the connection through the normal reconnect path — and re-raises with the root cause chained. `_cancel_task` no longer suppresses non-cancellation exceptions, so the real traceback survives `AsyncExitStack` unwind instead of vanishing. The regression test is exactly the injection I asked for: a reducer failure *after* READY, proving DEGRADED appears in transitions, BACKING_OFF is reached, `reconnect_failures` increments, and the error text carries both the consumer context and the root cause ("reducer exploded").
- **`Any` boundary (was #118 finding 2):** `recovery` is now typed against a small `RuntimeRecovery` Protocol, and `_on_recovery_state` takes a real `RecoveryState` via a `TYPE_CHECKING`-only import — mypy honesty restored without pulling websockets onto the Hook import path.

### One observation
The failure path now traverses two routes at once — the consumer's own `disconnect()` → `wait_closed` return in the main loop, *and* the re-raised exception surfacing through `_cancel_task` during stack unwind. Both converge on the same DEGRADED/backoff outcome (the test pins it), but the double reporting means `last_error` gets written twice with slightly different framing. Harmless today; if it ever confuses telemetry, collapsing to one canonical path is a small cleanup, not a bug.

### Verdict
Approve — merging. The "silently healthy" gap flagged on #118 is closed before anything depends on READY.

PR #120, opened 2026-08-13
feat(observe): add observability coverage audit

### LINKED ISSUE #37: Measure the observability ceiling
### Question / hypothesis

What fraction of important coding-agent failures are visible early enough for Spotter to act on them? If the observation surface hides the relevant evidence, a stronger reviewer cannot solve the problem.

### Experiment design

Label failed or degraded trajectories and record:

- whether the failure was `visible`, `invisible`, or not judgeable
- the earliest step/turn where intervention became warranted and observable
- whether goal/constraint context was available
- whether relevant tool outcomes were observable
- failure classes that are structurally invisible on the current surface

Measure the current Hook-based prototype as a baseline, then repeat after the Observe stage moves primary observation to the Codex App Server.

**NeMo Relay issue #425** is a particularly important reference because it demonstrates a real structural observability failure in Codex MultiAgentV2: delegated content can be encrypted/unavailable in the parent trace and parent-child lineage can be incomplete even though the child successfully runs.

- https://github.com/NVIDIA/NeMo-Relay/issues/425
- Codex integration notes: https://github.com/NVIDIA/NeMo-Relay/blob/main/docs/nemo-relay-cli/codex.mdx

The lesson for Spotter is that missing evidence has at least two very different causes:

```text
SOURCE_LIMIT
  Codex/App Server did not expose the information in usable form

ADAPTER_LOSS
  source exposed it, but Spotter failed to normalize/correlate/project it
```

These must not be mixed. A parser/Trace IR bug should not be called the theoretical observability ceiling, and encrypted/unexposed source information should not be blamed on reviewer quality.

**OpenClaw's App Server event projector** is useful as a checklist of event families that a mature Codex integration can extract and as a reference for keeping unknown/unprojected methods diagnosable:

- https://github.com/openclaw/openclaw/blob/main/extensions/codex/src/app-server/event-projector.ts
- runtime boundary: https://github.com/openclaw/openclaw/blob/main/docs/plugins/codex-harness-runtime.md

These are references only; Spotter does not depend on NeMo Relay/OpenClaw.

### Measure a coverage matrix, not one percentage

A single "X% observable" figure loses too much information. Record coverage by evidence family and failure class.

Suggested evidence dimensions:

```text
user goal / explicit constraints
thread / turn identity
plan summary where exposed
reasoning summary where exposed
file reads / repository exploration
commands/tool calls
stdout/stderr/exit outcome
file edits/diffs
MCP calls/results
subagent spawn/lineage/task content
validation/test outcomes
intervention input/delivery
external-effect outcome
runtime timing/token usage
```

For each relevant item classify:

```text
OBSERVED_EXACT
OBSERVED_PARTIAL
OBSERVED_ENCRYPTED
OBSERVED_UNCORRELATED
SOURCE_NOT_EXPOSED
ADAPTER_DROPPED
OBSERVATION_GAP
LEGACY_UNAVAILABLE
NOT_APPLICABLE
UNKNOWN
```

This lets Spotter distinguish architectural limits from implementation bugs.

### Raw-source audit vs normalized Trace IR audit

For the App Server measurement, retain a bounded raw-event sample/reference under #85 so the audit can compare:

```text
Codex/App Server raw stream
       ↓
what was actually exposed?
       ↓
Spotter CodexAdapter / Trace IR
       ↓
what did Spotter preserve?
       ↓
ThreadState / detector input
```

For a labeled miss, ask in order:

1. was the required fact present in the raw source?
2. if yes, did #85 normalize/correlate it correctly?
3. if yes, did #31 preserve it in state?
4. if yes, was it visible early enough for Detect/Intervene?
5. only then is it a detector/reviewer miss rather than an observation miss.

### Define "visible" relative to the intervention time

A fact arriving after the action or failure has already compounded is not operationally visible for that intervention.

For each labeled intervention opportunity record:

```text
failure/decision boundary
required evidence set
first source-observable point for each evidence item
first Spotter-normalized point
first ThreadState-available point
earliest actionable point
```

Then classify the opportunity:

```text
VISIBLE_IN_TIME
VISIBLE_TOO_LATE
STRUCTURALLY_INVISIBLE
LOST_BY_ADAPTER
LOST_BY_GAP
UNJUDGEABLE
```

This feeds #24 directly.

### Failure taxonomy should include observation requirements

When labeling a failure, record not just the semantic failure type but the minimal evidence a reasonable supervisor would need.

Example:

```text
failure = wrong implementation target
required evidence:
  - explicit issue scope
  - file/module being edited
  - relevant existing abstraction discovery
```

If the existing abstraction was never exposed because Main never inspected it, that is different from the App Server hiding a tool result. One is an agent exploration problem; the other is an observation-surface problem.

### Distinguish "not observed because Main never did it"

Important categories:

```text
NOT_PERFORMED
  Main never read/tested/queried the evidence

PERFORMED_BUT_NOT_EXPOSED
  action happened but source omitted/encrypted result

EXPOSED_BUT_LOST
  adapter/state failed to preserve it
```

ECLoop-like premature commitment often belongs to `NOT_PERFORMED`: the runtime can observe that evidence prerequisites remain unsatisfied even though the missing evidence itself never existed in the trajectory.

Do not call that a low App Server ceiling.

### Observation-gap accounting

#87 introduces explicit reconnect/daemon gaps. Any failure whose decisive evidence may have occurred inside a gap should not be labeled structurally invisible.

Classify it separately:

```text
GAP_CONTAMINATED
```

and report how much of the dataset is contaminated by observation interruptions.

### Hook vs App Server comparison

Run the same labeling framework on:

**Hook-era baseline**

- current Hook journal surface;
- explicitly lower identity/time/outcome coverage.

**App Server era**

- raw source audit;
- normalized Trace IR audit;
- live ThreadState availability.

Report deltas by evidence/failure family, not just overall percentage.

A useful table shape:

| Evidence/failure family | Hook exact | App Server source exact | Trace IR exact | Lost/partial | Structurally absent |
| --- | ---: | ---: | ---: | ---: | ---: |
| tool outcome | | | | | |
| file diff | | | | | |
| user constraint | | | | | |
| subagent lineage | | | | | |
| ... | | | | | |

### Build a source-vs-adapter conformance corpus

Take representative real/synthetic App Server traces and annotate expected normalized facts:

- command start/output/completion;
- file edit/diff;
- MCP progress/result;
- plan/reasoning summary events;
- token usage;
- turn completion;
- subagent events where available;
- unknown/new method;
- disconnect gap.

Run this as regression coverage for #85. The observability study should not repeatedly rediscover deterministic adapter bugs by manual inspection.

### Missing/encrypted content rule

NeMo Relay #425 should become a permanent regression/epistemic rule:

- ciphertext is not delegated-task plaintext;
- a visible child thread is not proof that parent-child task content is available;
- absent lineage must remain partial/unknown;
- downstream detector/reviewer prompts should know when evidence is unavailable rather than receive fabricated reconstruction.

Report at minimum:

- failure/opportunity count and labeling coverage;
- `VISIBLE_IN_TIME`, `VISIBLE_TOO_LATE`, `STRUCTURALLY_INVISIBLE`, `ADAPTER_LOSS`, `GAP_CONTAMINATED`, `UNJUDGEABLE` counts;
- source-vs-Trace-IR coverage by evidence family;
- proportion of misses caused by Main never gathering evidence vs runtime not exposing it;
- Hook→App Server improvement by category;
- earliest-observable/actionable timing distributions where timestamps exist.

Treat the existing ~10% observable tool-result figure as a Hook-era diagnostic, not the final ceiling.

If the post-App-Server source ceiling is low for important failure classes, prioritize observation/control-surface expansion rather than reviewer tuning. If raw source coverage is high but Trace IR coverage is low, fix #85 first. If both are high but detection misses remain, then #28/#38 become the bottleneck.

### Failure-mode tests

- raw App Server event contains needed field but adapter drops it → ADAPTER_LOSS
- raw source never exposes field → STRUCTURAL/SOURCE limit
- child task content encrypted but child execution visible
- outcome occurs during reconnect gap
- Main never performs expected evidence-gathering action
- Hook event and App Server event provide conflicting/partial observations
- newly added unknown App Server method appears and is counted rather than silently discarded
- late result arrives after intervention opportunity closed → VISIBLE_TOO_LATE

### Notes / constraints

- A null/unfavorable result is useful: it identifies an instrumentation limit rather than a model-quality problem.
- #85 owns normalized ingestion; #31 owns live state; #87 owns gaps; #24 owns delay analysis.
- Do not infer hidden chain-of-thought or reconstruct encrypted content for the sake of raising the coverage number.
- NeMo Relay/OpenClaw are references only, not dependencies.

### Done when

- Hook-era and post-App-Server observations use the same explicit coverage taxonomy.
- raw-source coverage and adapter/Trace-IR preservation are measured separately.
- structural source limits, adapter loss, missing Main evidence gathering, and reconnect gaps are distinguishable.
- at least one conformance corpus protects the main App Server event families from silent normalization loss.
- post-App-Server visible-in-time ceiling is reported by important failure/evidence class with sample size/coverage.
- the resulting architecture decision is explicit: observation surface, adapter, or detector/reviewer is identified as the next bottleneck.

YoungJinJung:
## Why

Issue #37 needs source limits, normalization loss, live-state loss, reconnect gaps, and late evidence to remain distinguishable. Counting raw events or treating an implemented adapter as outcome evidence would overstate the observability ceiling.

## What changed

- add the explicit evidence-family, coverage, and intervention-opportunity taxonomy
- retain a bounded mode-0600 value-free App Server field-shape audit with locking, fsync, torn-tail repair, compaction, and duplicate exclusion
- measure Hook and App Server Trace IR separately, plus source-to-Trace-IR and Trace-IR-to-ThreadState preservation
- add a synthetic conformance corpus for lifecycle, command, file, MCP, plan, reasoning summary, usage, unknown methods, and encrypted child content
- add spotter observability and document the current aggregate baseline

## Current evidence

The available local snapshot contains 9 Hook journals, 0 App Server journals, 0 source audit samples, and 0/9 session-level observability labels. The command therefore reports that the post-App-Server ceiling cannot be stated. This PR intentionally does not close #37; representative labeled failed/degraded App Server trajectories are still required.

## Validation

- ruff format --check .
- ruff check .
- mypy src tests
- pytest (376 passed)

Refs #37

### DIFF
--- a/README.md
+++ b/README.md
@@ -347,6 +347,7 @@ spotter analyze
 spotter review --session <id>
 spotter label --session <id> --step <n> --verdict fp
 spotter metrics
+spotter observability [--session <id>]
 spotter fork --session <id> --step <n>
 spotter prune --repo /path/to/repo  # dry-run unless --apply is supplied
 spotter experiment --session <id> --step <n> --guidance "..." --check "..."
--- a/docs/README.md
+++ b/docs/README.md
@@ -12,6 +12,7 @@ Start here if you are reading the repository for the first time.
 | What happens from install through uninstall/reinstall? | [Lifecycle](lifecycle.md) | command-by-command operational lifecycle |
 | What did the App Server proof of concept establish? | [App Server PoC](app-server-poc.md) | exploratory protocol and transport findings |
 | Was shared TUI control validated end to end? | [App Server validation](app-server-validation.md) | historical validation evidence and limitations |
+| How is source/Trace IR/ThreadState coverage measured? | [Observability baseline](observability-baseline.md) | taxonomy, safe audit method, current sample limits |
 | What should become trustworthy next? | [Roadmap](roadmap.md) | Runtime → Observe → Detect → Intervene → Recover → Harden, with evidence gates |
 | What prior papers/systems should I study, and what does Spotter borrow from them? | [Reference](reference.md) | literature + implementation precedents, boundaries, and implementation hints |
 | What hypotheses remain unproven, and how will Spotter evaluate them? | [Research](research.md) | evidence posture, research questions, experiments, and metrics |
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -484,6 +484,13 @@ explicitly. Hook records carry legacy-session provenance with unknown thread, tu
 dimensions, so consumers never need Codex transport objects and never infer that a Hook session is an
 App Server thread.
 
+The daemon also writes a bounded, value-free source-shape audit beside session journals. It records
+which App Server field paths existed, which normalized fields and evidence families survived, and
+whether the fact remained available in live `ThreadState`. Source values, raw reasoning, command
+output, and encrypted content are never copied into this audit. `spotter observability` compares the
+Hook and App Server surfaces without treating adapter/state loss as a theoretical source limit. See
+[Observability ceiling baseline](observability-baseline.md).
+
 ---
 
 # 7. Reviewer jobs and freshness
--- a/docs/observability-baseline.md
+++ b/docs/observability-baseline.md
@@ -1,0 +1,91 @@
+# Observability ceiling baseline
+
+This document separates the measurement instrument from the evidence it has collected. The
+instrument exists; the post-App-Server ceiling required by
+[#37](https://github.com/spotter-agent/spotter/issues/37) is not yet measured.
+
+## What is measured
+
+`spotter observability` reports three distinct projection stages:
+
+```text
+Codex/App Server source field shapes
+          ↓
+normalized Trace IR
+          ↓
+live ThreadState
+```
+
+It also reports Hook and App Server Trace IR separately. Counts are grouped by evidence family, not
+collapsed into one event percentage. This prevents a parser or state-projection loss from being
+misreported as a structural source limit.
+
+Coverage uses the following explicit vocabulary:
+
+| Status | Meaning |
+| --- | --- |
+| `OBSERVED_EXACT` | the relevant exposed fact was preserved |
+| `OBSERVED_PARTIAL` | useful evidence survived, but not the complete fact |
+| `OBSERVED_ENCRYPTED` | only encrypted/opaque content was exposed |
+| `OBSERVED_UNCORRELATED` | evidence exists without proven thread/turn identity |
+| `SOURCE_NOT_EXPOSED` | the runtime did not expose the required fact |
+| `ADAPTER_DROPPED` | source or Trace IR exposed the fact, but a later projection lost it |
+| `OBSERVATION_GAP` | the fact may have occurred during a known disconnect gap |
+| `LEGACY_UNAVAILABLE` | an older record lacks the provenance needed to judge it |
+| `NOT_PERFORMED` | Main never performed the evidence-gathering action |
+| `NOT_APPLICABLE` | the evidence family is irrelevant to this opportunity |
+| `UNKNOWN` | available evidence cannot support a stronger classification |
+
+For a labeled failure or degraded trajectory, required evidence is classified relative to the
+failure/decision boundary as `VISIBLE_IN_TIME`, `VISIBLE_TOO_LATE`, `STRUCTURALLY_INVISIBLE`,
+`LOST_BY_ADAPTER`, `LOST_BY_GAP`, or `UNJUDGEABLE`. Missing, encrypted, or late facts are never
+reconstructed.
+
+## Source audit safety
+
+The daemon retains a bounded audit at `sessions/source-audit/samples.jsonl`. Samples contain method
+names, field paths, normalized field names, evidence families, correlation identity, epoch, and
+coverage classifications. They do not contain source values, command output, reasoning content, or
+encrypted content. The file is mode `0600`, fsynced, and compacted to the most recent 1,000 samples
+after it grows beyond 2,000.
+
+The synthetic conformance corpus in `tests/fixtures/app_server_conformance.json` protects command,
+file, MCP, plan, reasoning-summary, usage, lifecycle, unknown-method, and encrypted-child behavior.
+It is regression coverage only; synthetic cases are not outcome evidence and do not contribute to
+the measured ceiling.
+
+## Current aggregate snapshot
+
+On 2026-08-13, the available local dataset reported:
+
+```text
+sessions: 9 (hook=9, app_server=0)
+events: 2659; observation_gaps: 0; unknown_events: 0
+source-vs-adapter samples: none
+session observability labels: 0/9
+```
+
+The Hook records predate current provenance and therefore classify as `LEGACY_UNAVAILABLE`; they
+cannot establish exact source or timing coverage. There are no App Server journals, no raw-shape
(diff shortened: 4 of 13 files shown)


youngjinjung-linq:
## Review

### Overview
Refs #37 — deliberately not closing it (+1303): the evidence-family/coverage/opportunity taxonomy, a bounded value-free `SourceAuditStore` (0600, flock, fsync, torn-tail repair, compaction), separate Hook/App Server/source/ThreadState measurement, an 11-case synthetic conformance corpus, the `spotter observability` command, and a written zero-sample baseline. Verified on the branch: full suite 376/376, `mypy` clean, `ruff` clean.

### Strengths
- **The epistemics are the headline and they hold**: the PR ships an instrument while explicitly refusing to state a ceiling ("Refs", a new **No** row in the evidence table, the CLI printing "post-App-Server source ceiling cannot be stated", synthetic cases barred from counting as outcome evidence). This is the null-results-first-class culture applied to the project's own tooling.
- **Privacy is enforced by test, not assertion**: `SECRET_VALUE`, `SYNTHETIC_PRIVATE_REASONING`, and `SYNTHETIC_CIPHERTEXT` are proven absent from serialized samples; the audit stores field *paths* only.
- The taxonomy actually prevents the conflation #37 warns about — `ADAPTER_DROPPED` vs `SOURCE_NOT_EXPOSED` vs `LOST_BY_GAP` vs `NOT_PERFORMED` are distinct and `classify_opportunity` is tested across every class. Hook `tool_result` without an exit code honestly classifies `OBSERVED_PARTIAL`.
- Audit I/O failure cannot break observation (`audit_source` catches, records, warns — with a test-visible `last_source_audit_error`).

### Findings (none blocking)
**1. `load()` is strict against the torn tail its own writer repairs.** `_repair_tail` runs only on `record()`; a crash mid-append followed by `spotter observability` (before any new write) hits the partial last line and raises `ObservabilityError` → exit 1. Same shape as the #114 torn-tail finding, lower stakes (a report, not observation). Tolerating an unterminated final line in `load()` is a three-line fix.
**2. `"encrypted" in field.lower()` matches "unencrypted".** A field literally named `unencryptedContent` would classify the whole sample `OBSERVED_ENCRYPTED` — the substring test needs a word-boundary or segment match. One line.
**3. `_PRESERVED_FIELDS`/`_PRESERVED_METHOD_FIELDS` hand-mirror the normalizer's `_known(...)` whitelists.** Drift is asymmetric: a field *added* to the normalizer but not the map silently reduces audit sensitivity (no `ADAPTER_DROPPED` possible for it), while the conformance corpus only pins the 11 fixture shapes. A test asserting the maps against the normalizer's actual whitelist would make the mirror self-checking — same dual-source-of-truth concern as raised on #111, in a milder form.
**4. Minor**: the audit performs a locked fsync write per event inside the async consumer — consistent with the existing synchronous journal write in the same path and fine at human-driven event rates, but it doubles per-event fsyncs; worth a ceiling note if event rates ever grow. Also `--session` filtering for a hook session keeps `thread_id=None` (unscoped) source samples rather than excluding them.

### Verdict
Approve — merging. Findings 1–3 are small sharpening commits for whenever the instrument starts receiving real App Server samples; none change what the instrument currently claims, which is — correctly — nothing.

PR #121, opened 2026-08-13
fix(observe): harden source audit edge cases

YoungJinJung:
## Summary

Addresses the three actionable follow-ups from the merged PR #120 review:

- ignore an unterminated final audit line during read-only reports
- match encrypted field segments without treating unencryptedContent as ciphertext
- make the normalizer and source audit consume one shared field whitelist, including MCP/dynamic duration and web-search fields

## Verification

- ruff format --check .
- ruff check .
- mypy src tests
- pytest (378 passed)

Refs #37

### DIFF
--- a/src/spotter/ingestion.py
+++ b/src/spotter/ingestion.py
@@ -16,6 +16,8 @@ from spotter.identity import (
     TurnStatus,
 )
 from spotter.observability import (
+    APP_SERVER_ITEM_FIELDS,
+    APP_SERVER_METHOD_FIELDS,
     SOURCE_AUDIT_RELATIVE_PATH,
     CoverageStatus,
     SourceAuditStore,
@@ -58,7 +60,7 @@ class CodexTraceNormalizer:
             identity = self._identity(started_thread_id)
             return TraceEvent(
                 "thread_started",
-                _known(thread, "cwd", "sessionId", "status", "name"),
+                _known(thread, *APP_SERVER_METHOD_FIELDS["thread/started"].values()),
                 event_id,
                 _seconds(thread.get("createdAt")),
                 identity,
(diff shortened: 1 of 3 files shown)


youngjinjung-linq:
## Review

Follow-up to the #120 review, addressing all three actionable findings (+91/−66). Verified on the branch: full suite 378/378, `mypy` clean, `ruff` clean.

### Addressed
- **Torn-tail-strict `load()` (was #1):** an unterminated final line is now dropped during read-only loads, with the test asserting a successful load *before* the next repairing write. `spotter observability` no longer errors after a crash-mid-append.
- **`unencrypted` false positive (was #2):** segment-boundary regex replaces the substring test, and `test_unencrypted_field_name_is_not_classified_as_ciphertext` pins it.
- **Hand-mirrored whitelists (was #3):** `APP_SERVER_ITEM_FIELDS`/`APP_SERVER_METHOD_FIELDS` are now the single source consumed by both `CodexTraceNormalizer._known(...)` calls and the audit's `_PRESERVED_FIELDS` (derived, not duplicated) — and `test_normalizer_and_audit_share_the_item_field_whitelist` round-trips every item type to `OBSERVED_EXACT`, making the mirror self-checking exactly as asked. Bonus: `webSearch` and `dynamicToolCall.durationMs` audit coverage came along for free.

### One remaining wrinkle (non-blocking)
The new regex `(?:^|[._-])encrypted(?:$|[._-]|[A-Z])` fixed over-matching but introduced a narrow under-match: it only recognizes lowercase `encrypted` after a start/separator boundary, so a camelCase mid-word form like `childEncryptedContent` (or `isEncrypted`) no longer classifies as `OBSERVED_ENCRYPTED`. That failure direction is the uncomfortable one for this project — ciphertext classified as visible rather than visibility classified as ciphertext. Codex's actual field (`encryptedContent`, leading lowercase) matches fine today, so this is theoretical until a new wire shape appears; splitting field segments on `[._-]` + camelCase humps and comparing `segment.lower() == "encrypted"` would close it in a few lines.

### Verdict
Approve — merging.

PR #122, opened 2026-08-13
fix(observe): detect camel-case encrypted fields

YoungJinJung:
## Summary

Addresses the remaining PR #121 review wrinkle by splitting source field paths on separators and camelCase boundaries before checking for an encrypted segment. This recognizes encryptedContent, childEncryptedContent, and isEncrypted without regressing unencryptedContent.

## Verification

- ruff format --check .
- ruff check .
- mypy src tests
- pytest (381 passed)

Refs #37

### DIFF
--- a/src/spotter/observability.py
+++ b/src/spotter/observability.py
@@ -23,6 +23,7 @@ if TYPE_CHECKING:
     from spotter.thread_state import ThreadState
 
 SOURCE_AUDIT_RELATIVE_PATH = Path("source-audit") / "samples.jsonl"
+_FIELD_SEGMENT = re.compile(r"[._-]|(?<=[a-z0-9])(?=[A-Z])")
 
 
 class ObservabilityError(ValueError):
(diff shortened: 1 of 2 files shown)


YoungJinJung:
## Review

Closes out the remaining wrinkle from the #121 review (+20/−4). Verified on the branch: full suite 381/381, `mypy` clean, `ruff` clean.

`_is_encrypted_field` now splits field paths on `[._-]` and camelCase humps (`(?<=[a-z0-9])(?=[A-Z])`) and casefold-compares each segment — exactly the suggested shape. The parametrized matrix covers all four cases that mattered: `encryptedContent`, `childEncryptedContent`, and `isEncrypted` classify `OBSERVED_ENCRYPTED`, while `unencryptedContent` stays `UNKNOWN`. The casefold comparison also handles an all-caps `ENCRYPTED` segment for free.

The under-match failure direction flagged on #121 — ciphertext classified as visible — is closed. Approve — merging.

Issue #86, opened 2026-08-12
Reduce Codex observation Hooks to the minimal PreToolUse enforcement bridge


Bogyie:
### Task / problem

The current plugin registers `SessionStart`, `UserPromptSubmit`, `PreToolUse`, and `PostToolUse`. The target Codex integration uses App Server events as the primary observation stream and retains Hooks only where synchronous deterministic enforcement still requires `PreToolUse`.

Once App Server ingestion has proven the required coverage, remove redundant observation Hooks and their duplicated hot-path responsibilities.

### Scope

- remove `PostToolUse` observation/review-trigger responsibilities after tool results/diffs are reliably available from App Server
- remove `UserPromptSubmit` once user-goal/message coverage is proven
- remove `SessionStart` once thread/runtime lifecycle is authoritative from App Server
- keep `PreToolUse` only for the bounded enforcement bridge unless a trustworthy atomic App Server veto capability later replaces it
- prevent double journaling/double metrics when both old and new paths temporarily coexist during migration
- update plugin/standalone integration artifacts and current-vs-target docs accordingly

### Implementation references — study, do not depend on

OpenClaw's Codex runtime documents the same split Spotter is converging on: App Server notifications are useful for async lifecycle/tool observations, while native Hooks remain the place for low-level synchronous blocking where the host exposes it.

- runtime Hook/App Server boundary: https://github.com/openclaw/openclaw/blob/main/docs/plugins/codex-harness-runtime.md
- App Server event projection: https://github.com/openclaw/openclaw/blob/main/extensions/codex/src/app-server/event-projector.ts

NeMo Relay is useful as a migration/duplication reference:

- Codex Hook coverage and limitations: https://github.com/NVIDIA/NeMo-Relay/blob/main/docs/nemo-relay-cli/codex.mdx
- persistent integration lifecycle: https://github.com/NVIDIA/NeMo-Relay/blob/main/docs/nemo-relay-cli/plugin-installation.mdx

Relay explicitly warns that keeping both a source-installed integration and generated integration active can forward the same lifecycle payload twice. Spotter should treat temporary Hook/App Server overlap as the same class of problem: coexistence is a migration state that needs explicit source authority and deduplication, not an excuse to double-consume observations.

### Migration matrix: prove responsibility parity before removal

Do not remove Hooks by name. Build a responsibility matrix and prove each responsibility has moved.

| Current Hook responsibility | Target authority | Removal prerequisite |
| --- | --- | --- |
[...]

Issue #33, opened 2026-08-11
Track runtime cost, efficiency, and outcome metrics


YoungJinJung:
### Problem / outcome

Spotter cannot be judged on intervention quality without also showing what supervision costs. The runtime needs first-class telemetry for Main efficiency, Spotter overhead, timing, and mechanically decidable outcomes.

### Proposed direction

Record and report, per thread/turn where available:

- main-agent tokens
- reviewer tokens and model-call count
- tool-call and repeated-action counts
- elapsed/tool execution time
- reviewer and delivery latency
- runtime CPU/memory and storage growth where practical
- task success/regression where an objective check exists

Use the App Server/Trace IR timeline as the primary source for token and timing data after the Observe stage, and persist authoritative observation/receipt timing in durable history so later analysis does not infer time from record order. Old records without trustworthy time remain explicitly unknown.

Surface costs next to intervention counts/results in `spotter analyze` and `spotter metrics`; keep Main cost and Spotter cost separate.

### Implementation references — study, do not depend on

**OpenClaw's Codex event projector** is a useful concrete reference for timing/order and token/tool telemetry:

- https://github.com/openclaw/openclaw/blob/main/extensions/codex/src/app-server/event-projector.ts

It explicitly tracks event families for usage, tool lifecycle/output, assistant/reasoning events, and completion, and uses a monotonic receipt-order timestamp allocator for mirrored events that may share the same wall-clock timestamp. Spotter should preserve the same distinction between *when the source says something happened* and *when Spotter received/processed it*.

**NeMo Relay's ATOF/ATIF/observability stack** is useful for the raw-event → normalized trajectory → telemetry/export separation:

- runtime/observability overview: https://github.com/NVIDIA/NeMo-Relay
- Codex capture limitations: https://github.com/NVIDIA/NeMo-Relay/blob/main/docs/nemo-relay-cli/codex.mdx
- projection-loss example: https://github.com/NVIDIA/NeMo-Relay/issues/381

The important lesson from #381 is that an event can exist in the raw stream yet disappear or lose ordering/parentage in a normalized/exported view. Spotter metrics should therefore retain provenance/coverage rather than assuming every missing normalized field means the runtime never emitted it.

These are references only; Spotter should not depend on OpenClaw/NeMo Relay.

### Use a timestamp model that supports latency attribution

Each relevant Trace IR/operational event should carry where available:

```text
source_timestamp?          # only if upstream provides trustworthy source time
[...]

Issue #38, opened 2026-08-11
Measure detector precision and miss rate


YoungJinJung:
### Question / hypothesis

Are Spotter's gates, candidate signals, and reviewer verdicts precise enough **and** sensitive enough to justify active intervention? A detector that rarely fires must not look perfect merely because only its positives are labeled.

### Experiment design

Evaluate emitted and non-emitted cases together:

- label gate/candidate/reviewer positives with written criteria
- sample unflagged proposals and trajectories to estimate misses
- stratify carefully where random sampling is too sparse, and document the sampling bias
- report coverage per rule/failure class
- separate abstention or missing observability from an incorrect verdict
- record rater identity and double-label a subset to estimate labeling agreement

Use mechanically scored failures from #21 where possible to seed useful negative/miss samples.

### Measurement / decision rule

Report, with sample size and coverage:

- precision / false-positive rate
- miss rate or recall estimate
- abstention / blind-spot rate
- agreement on the double-labeled subset

Do not report precision alone. Use these numbers to decide which deterministic rules may become active and what conservative evidence threshold the Intervene stage may use for live `VERIFY/NUDGE`.

### Notes

This issue absorbs the recall gap from #50 and label-reliability work from #44. Persisted label/schema evolution remains tracked separately in #47.

YoungJinJung:
PR #311 fixes two measurement-truth defects found in the #38 audit: gate `FP/(TP+FP)` is now reported as false discovery (`1 - precision`) while true FPR is explicitly unavailable without a per-rule TN frame; multi-event-kind silence batches now produce one batch-level tally whose rate matches eligibility/exclusion metadata. Focused 42 tests and isolated full 1000-test suite pass. #38 remains open for representative labels, agreement, coverage, miss/abstention evidence, and activation thresholds.

YoungJinJung:
Reopened after PR #311 merged. #311 corrected FDR/FPR terminology and multi-kind sampling-frame denominator alignment; it did not collect the representative positive/silence labels, preselected double-label agreement subset, blind-spot coverage, or activation thresholds required to complete #38.

PR #123, opened 2026-08-13
feat(metrics): report coverage-aware runtime costs

YoungJinJung:
## Why

#33 requires cost and timing to be visible without making unavailable data look free or deriving latency across unrelated clocks. #86 is dependency-unblocked, but its own parity gate remains unmet because the current #37 baseline has zero App Server sessions and zero labeled failures, so removing observation Hooks now would be unsafe.

## What changed

- project durable journals into separate Main, Spotter semantic, and deterministic/runtime accounting domains
- correlate action starts/outcomes by semantic identity and report correlation coverage instead of inventing identity for uncorrelated events
- report token scope and turn coverage, reviewer calls/session tokens, Hook/IPC/evaluation latency coverage, source/receipt timing coverage, tool durations, and storage
- keep Hook and App Server surfaces separate during migration and print unavailable values as unknown
- avoid subtracting source and receipt wall clocks

This is a complete foundational slice of #33; resource sampling, queue/delivery stages, and objective outcome joins remain, so this PR does not close the issue.

## Validation

- ruff format --check .
- ruff check .
- mypy src tests
- pytest (384 passed)
- spotter metrics against the current 9-session local corpus

Refs #33

### DIFF
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -338,6 +338,13 @@ version-mismatched responses and unsupported proposal shapes fail open. Daemon e
 time, and total Hook time are recorded separately so latency percentiles can be computed without putting
 aggregation on the synchronous path.
 
+`spotter metrics` projects these durable records into separate accounting domains: Main semantic
+actions and token observations, Spotter semantic reviewer calls/tokens, deterministic Hook/IPC
+latency, timing coverage, and journal storage. Hook and App Server action surfaces are reported
+separately during migration rather than summed. Token totals remain `cumulative/unknown-scope`, and
+missing token/timing/latency data is printed as unknown with an explicit coverage denominator. Source
+wall time is never subtracted from receipt wall time to manufacture a latency across clock domains.
+
 ## 4.5 Turn completion
 
 ```text
--- a/docs/observability-baseline.md
+++ b/docs/observability-baseline.md
@@ -72,9 +72,9 @@ samples, and no failed/degraded session labels in this snapshot. Consequently:
 - no post-App-Server visible-in-time percentage can be stated;
 - no Hook-to-App-Server improvement can be stated;
 - no important failure-class ceiling can be stated; and
-- #37 remains open.
+- the post-App-Server evidence gate remains unmet.
 
-## Evidence needed to finish #37
+## Evidence still needed after #37's instrumentation work
 
 Collect representative failed or degraded App Server trajectories, label each failure boundary and
 minimal required evidence set, and run:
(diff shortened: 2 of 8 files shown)


YoungJinJung:
## Review

### Overview
Refs #33 — a foundational slice, correctly not closing it (+408/−4): `runtime_metrics.py` projecting durable journals into Main / Spotter-semantic / deterministic accounting domains, wired into `spotter metrics`, with coverage denominators everywhere and three targeted tests. Verified on the branch: full suite 384/384, `mypy` clean, `ruff` clean.

### Strengths
- **The #113 lesson generalized to metrics**: unavailable values render as `unknown` with explicit denominators (`hook=unknown (0/0)`, `receipt_wall=0/1`) — pinned by `test_unavailable_runtime_metrics_render_unknown_not_zero`. Nothing unavailable masquerades as free or zero.
- **Clock-domain discipline is honored, not just stated**: source and receipt wall clocks are counted for coverage but never subtracted to manufacture a latency.
- **Correlation honesty**: uncorrelated observations count in the denominator without inventing semantic identity (`test_uncorrelated_action_observations_do_not_invent_semantic_identity`), and token totals carry the `cumulative/unknown-scope` label instead of pretending attribution.
- Hook and App Server surfaces stay separate through the migration window, consistent with the #86 gate being explicitly unmet.

### Findings (non-blocking)
**1. This is the third independent "did this outcome fail" classifier.** `_outcome_failure` here, `thread_state._failed`, and `observability._hook_outcome_visible` now each define failure slightly differently — this one adds `interrupted`/`cancelled` as failures and `tool_response.ok` handling that `_failed` lacks. Divergent classifiers mean `spotter metrics` failure counts and ThreadState `recent_failures` can disagree about the same event, which is exactly the drift the #121 shared-whitelist fix eliminated for field preservation. Same medicine: one shared outcome classifier consumed by all three.
**2. Presentation nit**: `actions=1/2 correlated` reads like a 50% correlation ratio, but the denominator is raw observations (start+outcome for one action) — a dedup ratio, not a correlation rate. `actions=1 (from 2 observations)` or similar would stop the misread.
**3. Minor**: `cumulative_main_tokens` sums the *latest* `token_usage.total` per session — correct if the server's total is monotone per thread, which the `cumulative/unknown-scope` label honestly hedges; worth one comment naming the monotonicity assumption.

### Verdict
Approve — merging. Finding 1 is the one worth a follow-up commit before #38/#34 build cost dashboards on these numbers.

PR #125, opened 2026-08-13
feat(metrics): measure reviewer latency stages

YoungJinJung:
## Summary
- persist reviewer job queued and inference-started markers with stable per-session job identities
- measure cross-process queue delay from durable receipt time and inference duration from a same-process monotonic clock
- report queue/inference latency with explicit stage coverage and queued/started/decided counts

## Verification
- `.venv/bin/ruff check .`
- `.venv/bin/mypy src tests`
- `.venv/bin/pytest` (398 passed)

Part of #33.

### DIFF
--- a/src/spotter/cli.py
+++ b/src/spotter/cli.py
@@ -146,6 +146,7 @@ def build_parser() -> argparse.ArgumentParser:
         "--reservation",
         help="review: token for a budget slot the caller already reserved (internal)",
     )
+    parser.add_argument("--review-job-id", help=argparse.SUPPRESS)
     parser.add_argument(
         "--check", help="experiment: success command run in each fork worktree (exit 0 = pass)"
     )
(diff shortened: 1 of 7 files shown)


YoungJinJung:
## Review

### Overview
Part of #33 (+120/−8): durable `review_job_queued` → `review_inference_started` → `reviewer_decision` stage markers with stable per-session job IDs (`proposal:N`), queue delay measured across processes from durable receipt time, inference duration from a same-process monotonic clock, and a decided/started/queued funnel in `spotter metrics`. Verified on the branch: full suite 398/398, `mypy` clean, `ruff` clean.

### Strengths
- **Clock discipline extended correctly to a new stage pair**: cross-process queue delay subtracts receipt-wall from receipt-wall (same journal clock family, `max(0, …)` guarding skew), and inference uses `perf_counter_ns` within one process — no cross-domain subtraction, exactly the #123 rule applied rather than restated.
- **The funnel is diagnostic, not just decorative**: `decided/started/queued` separates spawn failures (queued but never started) from reviewer errors (started but not decided, with `reviewer_error` now carrying the job ID), and each latency sample has the right denominator (queue over started, inference over decided).
- The spec_drift-discard path preserves `inference_ms` on the replacement decision, and the manual `spotter review` path honestly records `timing.inference_ms: None` rather than inventing a queue it never had — both pinned by tests.
- The job ID travels via a suppressed internal flag and the spawn argv is asserted in `test_cadence_forwards_user_config` — the cross-process contract is tested where it's made.

### Minor (non-blocking)
- `review_inference_started` is journaled before the model call, so `started` includes runs that then fail — correct for the funnel, but worth remembering when reading `inference=avg(…)` (failed runs contribute a started marker and no inference sample; the denominator already accounts for it).
- `ReviewerDecision` is rebuilt positionally to attach `inference_ms`; `dataclasses.replace` would survive future field reordering. One line, cosmetic.

### Verdict
Approve — merging. Clean continuation of the #33 slice with the queue/delivery stage the #123 PR body promised as remaining work.

PR #126, opened 2026-08-13
feat(metrics): report turn wall-time coverage

YoungJinJung:
## Summary
- derive App Server turn wall time only from source timestamps in the same turn and connection epoch
- refuse reconnect-spanning or incomplete clock pairs instead of inventing a duration
- report source turn-time samples with completed-turn coverage in `spotter metrics`

## Verification
- `.venv/bin/ruff check .`
- `.venv/bin/mypy src tests`
- `.venv/bin/pytest` (399 passed)

Part of #33.

### DIFF
--- a/src/spotter/runtime_metrics.py
+++ b/src/spotter/runtime_metrics.py
@@ -37,6 +37,7 @@ class RuntimeCostReport:
     reviewer_jobs_started: int
     reviewer_queue_ms: tuple[float, ...]
     reviewer_inference_ms: tuple[float, ...]
+    turn_wall_ms: tuple[float, ...]
     tool_duration_ms: tuple[float, ...]
     gate_calls: int
     hook_ms: tuple[float, ...]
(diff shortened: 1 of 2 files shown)


YoungJinJung:
## Review

Part of #33 (+65/−8): App Server turn wall time derived only from *source* timestamps within the same turn **and** connection epoch, with reconnect-spanning or incomplete clock pairs refused rather than invented, and completed-turn coverage as the denominator. Verified on the branch: full suite 399/399, `mypy` clean, `ruff` clean.

The details are right:
- `_turn_clock` requires turn identity + epoch + a finite `occurred_at`, which excludes Hook-era events (no epoch) from source-clock durations entirely — no cross-surface contamination.
- Source-minus-source is the one subtraction the clock discipline permits, and the epoch key makes the reconnect case structurally impossible rather than merely unlikely — `test_turn_duration_never_crosses_connection_epochs` pins epoch-1-start/epoch-2-complete yielding `unknown (0/1)`, where the denominator still honestly counts the completed turn the sample was refused for.
- A same-epoch timestamp regression drops the sample instead of emitting a negative duration, and the miss stays visible in the coverage ratio.
- `setdefault` keeps the first observed start, so duplicate start deliveries can't shrink a duration.

Nothing to object to — small, consistent with the #123/#125 discipline, tested at exactly the boundary that matters. Approve — merging.

PR #127, opened 2026-08-13
feat(metrics): sample daemon resource costs

YoungJinJung:
## Summary
- sample daemon cumulative CPU and peak RSS on the first and every 64th gate request
- carry optional runtime-addressed samples through the existing gate telemetry without risking enforcement
- deduplicate samples by runtime/sequence and report latest resource costs with gate-call coverage

## Verification
- `.venv/bin/ruff check .`
- `.venv/bin/mypy src tests`
- `.venv/bin/pytest` (400 passed)

Part of #33.

### DIFF
--- a/src/spotter/daemon.py
+++ b/src/spotter/daemon.py
@@ -4,13 +4,16 @@ import argparse
 import asyncio
 import hashlib
 import json
+import math
 import os
 import plistlib
+import resource
 import shutil
 import stat
 import subprocess
 import sys
 import time
+import uuid
 from contextlib import suppress
 from dataclasses import dataclass
 from enum import StrEnum
@@ -37,6 +40,7 @@ START_TIMEOUT = 5.0
 STOP_TIMEOUT = 5.0
 _MAX_REQUEST_BYTES = 64 * 1024
 _SAFE_UNIX_PATH_BYTES = 100
+_RESOURCE_SAMPLE_EVERY = 64
 
 
 class RuntimeHealth(StrEnum):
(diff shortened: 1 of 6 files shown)


YoungJinJung:
## Review

Part of #33 (+155/−7): daemon `getrusage` samples (cumulative CPU + peak RSS) on the first and every 64th gate request, carried as optional `runtime_sample` telemetry through gate IPC into journals, deduplicated by (runtime_id, sample_seq) with latest-per-runtime aggregation and gate-call coverage in `spotter metrics`. Verified on the branch: full suite 400/400, `mypy` clean, `ruff` clean.

### Strengths
- **The `ru_maxrss` unit trap is handled**: Linux KiB vs macOS bytes, normalized at the source with a comment — the classic cross-platform footgun avoided rather than discovered later.
- **Telemetry cannot harm enforcement, in either direction**: `getrusage` failure returns `None` inside the daemon, and a malformed sample on the client validates to `None` instead of raising `DaemonProtocolError` — the deliberate inverse of the #102 strictness rule, correct here because turning telemetry corruption into gate fail-open would be the worse trade. The comment says so.
- **Runtime-addressed aggregation is restart-correct**: uuid-per-daemon-instance + monotonic seq means metrics sums the latest cumulative CPU across incarnations and takes max RSS, instead of double-counting resampled cumulative values — with dedup pinned by keying on (runtime_id, seq).
- The sampling cadence is proven (65 calls → exactly 2 samples, seq 1 and 64), concurrency yields exactly one sample across parallel gates, and an E2E hook test asserts a real sample lands in the journal.
- Coverage framing carried through: `samples=N/M gate calls` — the #123 denominators habit, applied again.

### Minor (non-blocking)
- The final CPU of a daemon incarnation after its last sample is inherently under-counted — bounded sampling accuracy; the coverage ratio makes it readable, but a half-line doc note near `_RESOURCE_SAMPLE_EVERY` would spare the future reader the derivation.
- `peak_rss=137445376 bytes` will read poorly at real scale — a MiB rendering is a one-liner whenever output polish matters.

### Verdict
Approve — merging.

Issue #21, opened 2026-08-11
Build a mechanically scored task set


YoungJinJung:
### Question / hypothesis

Can Spotter produce reproducible control/guidance runs whose outcome is mechanically decidable? This task corpus is the foundation for later intervention, harm, and recovery experiments.

### Experiment design

Build a small task corpus where every task records:

- repository or fixture version
- user prompt
- known initial failing state
- objective `--check` command
- provenance needed for reruns
- expected execution budget

A selected SWE-bench Verified subset is one option, but equivalent local fixtures are acceptable. Run the existing `spotter experiment --run --check ...` harness on both arms and record infrastructure failures separately from task failures.

### Implementation references — study, do not depend on

**Shepherd's live-intervention experiment repository** is the most useful concrete reference for how to freeze a supervised-agent benchmark so experiment code, task sets, prompts, and evaluation remain separable and reproducible:

- experiment layout: https://github.com/shepherd-agents/shepherd-experiments/tree/main/exp/live-intervention
- README / run protocol: https://github.com/shepherd-agents/shepherd-experiments/blob/main/exp/live-intervention/README.md
- batch runner: https://github.com/shepherd-agents/shepherd-experiments/blob/main/exp/live-intervention/src/experiment_live_intervention/core/run_batch_e2b.py
- evaluation runner/backend: https://github.com/shepherd-agents/shepherd-experiments/tree/main/exp/live-intervention/src/experiment_live_intervention/core
- frozen development/validation task configs live under `configs/`

Useful patterns:

- task-set files are versioned/frozen rather than selected ad hoc during a run;
- generation and evaluation are separate programs/steps;
- experiment condition/config is persisted independently from task definitions;
- raw run logs are retained and grading happens against an explicit harness afterward;
- development and held-out sets are separate;
- infrastructure/backend adapters are distinct from scoring logic.

**ECLoop** is useful as evidence that SWE-bench Verified can expose runtime-intervention effects without changing the base model/scaffold, but Spotter should not copy its benchmark claims as evidence for Spotter itself:

- paper: https://arxiv.org/abs/2607.28815

These are references only. Spotter should keep the task corpus/harness dependency-free where practical and use its existing experiment machinery.

### Task manifest contract

Introduce a versioned task manifest rather than encoding benchmark assumptions in CLI flags.

A task should carry enough information to reconstruct and validate the test condition, for example:

```yaml
task_schema_version: 1
task_id: fixture/parser-regression-001
source:
  kind: git
[...]

PR #128, opened 2026-08-13
feat(tasks): validate frozen corpus manifests

YoungJinJung:
## Summary
- define dependency-free versioned TOML contracts for task and task-set manifests
- freeze ordered task manifests and fixture trees with SHA-256 identities, rejecting drift, duplicates, traversal, and unsupported schemas
- add `spotter tasks validate <set.toml>` for static corpus validation without agent or scorer execution

## Verification
- `.venv/bin/ruff check .`
- `.venv/bin/mypy src tests`
- `.venv/bin/pytest` (405 passed)

Part of #21.

### DIFF
--- a/README.md
+++ b/README.md
@@ -350,10 +350,11 @@ spotter metrics
 spotter observability [--session <id>]
 spotter fork --session <id> --step <n>
 spotter prune --repo /path/to/repo  # dry-run unless --apply is supplied
+spotter tasks validate path/to/task-set.toml
 spotter experiment --session <id> --step <n> --guidance "..." --check "..."
 ```
 
-The experiment harness existing does **not** mean positive intervention advantage has been established. A mechanically scored task set and enough executed runs remain evidence gaps.
+Task-set validation freezes task and fixture hashes and checks the versioned scorer/budget contract; it does not execute setup, prechecks, scorers, or agent arms yet. The experiment harness existing does **not** mean positive intervention advantage has been established. Preflight execution, a mechanically scored corpus, and enough executed runs remain evidence gaps.
 
 ---
 
--- a/src/spotter/cli.py
+++ b/src/spotter/cli.py
@@ -62,6 +62,7 @@ from spotter.snapshot import (
     snapshot_references,
     stale_journals,
 )
+from spotter.task_corpus import TaskCorpusError, validate_task_set
 from spotter.trace import TraceEvent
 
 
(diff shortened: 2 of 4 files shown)


YoungJinJung:
## Review

### Overview
Part of #21 (+321/−4): dependency-free versioned TOML contracts for tasks and task sets, SHA-256 content addressing (file digest for manifests, deterministic path+bytes tree digest for fixtures), traversal/duplicate/schema rejection, and `spotter tasks validate` for static validation with an honest README note that nothing executes yet. Verified on the branch: full suite 405/405, `mypy` clean, `ruff` clean.

### Strengths
- **The freeze is real**: manifest drift and fixture drift each break a hash (both tested), the tree digest covers relative paths so a rename changes identity (tested), and the set/task `task_id` cross-check prevents a set pointing at the wrong frozen file.
- **The task shape encodes the experiment's epistemics**: `precheck.expected = "failure"` forces every task to be demonstrably broken before the agent runs — a mechanically scored corpus where "nothing to fix" is unrepresentable.
- `_contained_path` resolves and containment-checks both manifests and fixture roots, so a symlinked manifest escaping the corpus is caught.
- Stdlib only (`tomllib`/`hashlib`), versioned schemas rejected on mismatch, scoped exactly to validation — the right first slice of #21.

### Findings (non-blocking)
**1. Directory symlinks slip past the symlink rejection.** `fixture_digest` rejects file symlinks (`candidate.is_symlink()`), but `rglob` traverses *into* a symlinked directory, and the real files found inside it are not themselves symlinks — so a fixture containing `dir -> /outside` hashes external content despite the "symlinks are deliberately unsupported" docstring, and `_contained_path` never sees those paths. `os.walk(followlinks=False)` (or rejecting any candidate with a symlink among its parents) closes the hole. Corpus content is project-authored today, so this is a contract hole rather than an exposure — but the docstring currently promises more than the check enforces.
**2. Minor**: no requirement that at least one check has `required = true` — a task whose checks are all optional validates but can never fail scoring. Probably belongs to the executor slice; noting so it isn't forgotten.
**3. Nit**: the shared `target` positional now serves daemon/setup/teardown/tasks — the error messages keep it coherent, but this is approaching the point where argparse subparsers would cost less than the growing cross-checks.

### Verdict
Approve — merging. Finding 1 is a small hardening commit whenever corpus authorship opens up beyond the repo.

PR #129, opened 2026-08-13
feat(tasks): preflight mechanical scorers

YoungJinJung:
## Summary
- reject fixture directory symlinks and task manifests without any required scorer
- add `spotter tasks preflight <set.toml>` to execute setup, broken-state checks, a known-good transform, and positive checks in temporary fixture copies
- classify setup/check/timeout/unjudgeable failures, retain bounded command diagnostics, and kill timed-out process groups

## Verification
- `.venv/bin/ruff check .`
- `.venv/bin/mypy src tests`
- `.venv/bin/pytest` (409 passed)

Part of #21. Follow-up to #128.

### DIFF
--- a/README.md
+++ b/README.md
@@ -351,10 +351,11 @@ spotter observability [--session <id>]
 spotter fork --session <id> --step <n>
 spotter prune --repo /path/to/repo  # dry-run unless --apply is supplied
 spotter tasks validate path/to/task-set.toml
+spotter tasks preflight path/to/task-set.toml
 spotter experiment --session <id> --step <n> --guidance "..." --check "..."
 ```
 
-Task-set validation freezes task and fixture hashes and checks the versioned scorer/budget contract; it does not execute setup, prechecks, scorers, or agent arms yet. The experiment harness existing does **not** mean positive intervention advantage has been established. Preflight execution, a mechanically scored corpus, and enough executed runs remain evidence gaps.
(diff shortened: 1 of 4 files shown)


YoungJinJung:
## Review

Part of #21, follow-up to #128 (+318/−24) — and both #128 findings are closed before the new work starts. Verified on the branch: full suite 409/409, `mypy` clean, `ruff` clean.

### #128 findings addressed
- **Directory-symlink hole:** `fixture_digest` now walks with `os.walk(followlinks=False)`, symlink-checks every directory *and* file entry plus the fixture root, and converts walk errors into `TaskCorpusError` — with a test that builds a `dir -> outside` link and proves rejection. The docstring's promise is now enforced.
- **All-optional checks:** `at least one check must be required` is validated at freeze time (better than deferring to the executor, as I'd suggested), tested.

### The preflight executor
The pipeline encodes the right epistemics end to end: setup → precheck must *fail* on the broken fixture → required checks must not all pass on the broken state (the discriminator must exist) → `known_good` transform → required checks must all pass. A task without a `known_good` cannot be READY — you can't certify a scorer you've never seen pass — and every non-READY state is a distinct classification (`SETUP_FAIL`/`TIMEOUT_CHECK`/`CHECK_ERROR`/`UNJUDGEABLE`) with bounded command diagnostics retained. Process-group SIGKILL on timeout (`start_new_session` + `killpg`) is the correct kill shape, and the CLI exits non-zero unless every task is READY.

### Minor (non-blocking)
- After `killpg`, the follow-up `communicate()` has no timeout — a check command that double-forks into a new session can survive the group kill while holding the pipes, hanging preflight. Exotic, and `subprocess.run` has the same flaw, but a `communicate(timeout=…)` + suppress would close it.
- `shell=True` over corpus-declared commands is inherent to preflight's purpose and the corpus is hash-frozen and repo-authored — but one sentence in the README stating that trust boundary ("preflight executes corpus commands; only validate untrusted sets") would make it explicit.
- The shared `target` positional gained `preflight` — the subparsers suggestion from #128 stands, now with seven values on one argument.

### Verdict
Approve — merging.

PR #130, opened 2026-08-13
feat(tasks): freeze initial dev and validation corpus

YoungJinJung:
## Summary
- add hash-frozen dev and validation task sets with localized and multi-file mechanical fixtures
- prove both repository task sets READY through broken-state and known-good scorer preflight in CI
- bound post-timeout pipe cleanup for escaped child sessions and document the preflight command trust boundary

## Verification
- `.venv/bin/ruff check .`
- `.venv/bin/mypy src tests`
- `.venv/bin/pytest` (411 passed)
- `.venv/bin/spotter tasks preflight corpus/dev-v1.toml`
- `.venv/bin/spotter tasks preflight corpus/validation-v1.toml`

Part of #21. Follow-up to #129.

### DIFF
--- a/README.md
+++ b/README.md
@@ -355,7 +355,7 @@ spotter tasks preflight path/to/task-set.toml
 spotter experiment --session <id> --step <n> --guidance "..." --check "..."
 ```
 
-Task-set validation freezes task and fixture hashes and checks the versioned scorer/budget contract without executing commands. Preflight runs setup, the broken-state scorer, a declared known-good transform, and the positive scorer in a temporary fixture copy; it never runs agent arms. The experiment harness existing does **not** mean positive intervention advantage has been established. A frozen multi-task corpus and enough executed runs remain evidence gaps.
+Task-set validation freezes task and fixture hashes and checks the versioned scorer/budget contract without executing commands. Preflight runs setup, the broken-state scorer, a declared known-good transform, and the positive scorer in a temporary fixture copy; it never runs agent arms. Because preflight executes corpus-declared shell commands, use `validate` only for untrusted task sets. The experiment harness existing does **not** mean positive intervention advantage has been established. A frozen multi-task corpus and enough executed runs remain evidence gaps.
 
 ---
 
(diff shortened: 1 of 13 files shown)


YoungJinJung:
## Review

Part of #21, follow-up to #129 (+154/−3) — and both #129 minors are closed. Verified on the branch: full suite 411/411 (which itself validates the frozen hashes and preflights both repo task sets to READY), `mypy` clean, `ruff` clean.

### #129 follow-ups addressed
- **Escaped-child pipe hang:** the post-`killpg` `communicate()` now has a 1s bound; on a second timeout the pipes are closed, `wait` is bounded-and-suppressed, and the result still reports `timed_out` with whatever partial output the `TimeoutExpired` exceptions carried. Preflight can no longer hang on a double-forked scorer.
- **Trust boundary documented:** README now says "use `validate` only for untrusted task sets", and `corpus/README.md` repeats the split.

### The corpus itself
- Two well-chosen first families: a localized one-liner (`split("=")` → `split("=", 1)`) for dev and a multi-file contract change (normalize + render) for validation — different failure surfaces, both with genuinely failing prechecks and inline `known_good` transforms.
- **The freeze is CI-enforced**: `test_repo_corpus_is_frozen_and_preflight_ready` re-validates hashes and re-proves READY on every run, so corpus drift breaks the suite rather than silently rotting.
- `corpus/README.md` states the held-out hygiene rule explicitly — new version + new hashes for any change, never rewrite an observed validation set — and disclaims the synthetic fixtures as harness evidence, not intervention evidence. Consistent with the project's evidence posture.

### Minor (non-blocking)
- `fixtures/profile-contract/profile.py` shadows the stdlib `profile` module. It works — `sys.path[0]` wins from the workspace — but a fixture named after a stdlib module invites confusing failures if a future check imports differently; `user_profile.py` would dodge it for free in the next set version.

### Verdict
Approve — merging. #21's remaining gap is now purely scale (more tasks, executed runs), not machinery.

PR #133, opened 2026-08-13
feat(tasks): run resumable frozen task batches

YoungJinJung:
## Why

#21 now has frozen/preflighted tasks and explicit arm classifications, but there was no way to execute a whole task set from equivalent clean states or resume an interrupted paid batch safely.

Related: #21

## What changed

- add `spotter tasks run <set> --guidance ... --run` with an explicit paid-run opt-in
- preflight first, then run counterbalanced control/guidance arms from independent fixture copies
- enforce each task wall-time budget and classify setup, agent, scorer, timeout, task, and infrastructure outcomes
- persist fsync-backed JSONL rows with frozen set/fixture identity, pair IDs, declared budgets, bounded agent/check diagnostics, and environment provenance
- support `--resume` with torn-tail repair, completed-arm skipping, and refusal when set, environment, guidance, model, sandbox, or row provenance changed
- report arm-level classifications and only compare pairs whose two outcomes are mechanically judgeable
- document trust, lifecycle, current limitations, and the remaining #21 evidence gap

## Validation

- `ruff format --check .`
- `ruff check .`
- `mypy src tests`
- `pytest` (420 passed)
- repository dev and validation corpus validate/preflight commands both return `READY`

## Notes

No paid model run was launched in this PR. Execution behavior is covered with deterministic fake-agent tests; the existing frozen fixtures were preflighted through their real setup/scorer commands. `codex exec` exposes no hard max-turn flag, so this backend enforces wall time and persists the declared max-turn budget for parity/provenance. Corpus expansion and the first reportable real multi-task run remain follow-ups in #21.

### DIFF
--- a/README.md
+++ b/README.md
@@ -352,10 +352,13 @@ spotter fork --session <id> --step <n>
 spotter prune --repo /path/to/repo  # dry-run unless --apply is supplied
 spotter tasks validate path/to/task-set.toml
 spotter tasks preflight path/to/task-set.toml
+spotter tasks run path/to/task-set.toml --guidance "..." --run
+# after an interrupted batch, repeat the same conditions and add:
+# --resume ~/.spotter/experiments/task-batches/<batch>.jsonl
 spotter experiment --session <id> --step <n> --guidance "..." --check "..."
 ```
 
-Task-set validation freezes task and fixture hashes and checks the versioned scorer/budget contract without executing commands. Preflight runs setup, the broken-state scorer, a declared known-good transform, and the positive scorer in a temporary fixture copy; it never runs agent arms. Because preflight executes corpus-declared shell commands, use `validate` only for untrusted task sets. The experiment harness existing does **not** mean positive intervention advantage has been established. A frozen multi-task corpus and enough executed runs remain evidence gaps.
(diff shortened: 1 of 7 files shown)


YoungJinJung:
## Review

Part of #21 (+697/−11): `spotter tasks run` — a resumable batch executor with an explicit paid-run opt-in, preflight gating, counterbalanced control/guidance arms from independent clean fixture copies, fsync-backed provenance-rich JSONL, and condition-checked resume. Verified on the branch: full suite 420/420, `mypy` clean, `ruff` clean, PR CI green.

### Strengths
- **The experimental design is sound where naive harnesses go wrong**: each arm gets its own fixture copy (no cross-arm state leakage), arm order counterbalances deterministically per (task index + run_id) so resume preserves the same order, and `SPOTTER_DISABLE=1` follows the established hook/experiment/reviewer precedent — I verified `hook.py` honors it, so the shadow reviewer can't contaminate its own experiment arms.
- **Resume refuses everything it should**: schema, set hash, guidance, model, sandbox, python, platform, codex version, codex home — plus per-row provenance (fixture hash, budgets, pair-ID shape) re-validated against the *current* frozen manifests. Torn tails repair only at the last line; interior corruption and duplicate arm keys refuse loudly. Interrupted paid batches can be continued without silently changing conditions, which is the whole point.
- `SETUP_FAIL` is now actually assigned (closing the #132 note), classification logic reuses the shared `ArmClassification` rather than growing a fourth copy, and the escaped-child kill pattern from #130 is applied to the agent runner.
- The honesty trail continues: no paid run launched, `codex exec`'s missing max-turn flag documented in three places as "recorded, not enforced", and status.md keeps the corpus question at **Partial** with the real remaining gap named.

### Minor (non-blocking)
- Resume matches `platform.platform()` exactly — an OS patch release will strand a half-finished batch even though nothing meaningful changed. Deliberately strict and stated, but if it bites, matching on `(system, machine)` with the full string kept as provenance would preserve the intent with less friction.
- `max_turns` is now provenance-only for this backend — fine and documented; when a turn-capable backend appears, the field's enforced-vs-declared status should travel into the row rather than the README.
- The overloaded `target` positional reaches eight values; the `--resume`-requires-`run` cross-checks are correct but this is the third PR of manual grammar guards. The subparsers refactor is now cheaper than the next guard.

### Verdict
Approve — merging. With this, #21's machinery is complete end to end: freeze → preflight → classified, resumable execution. What remains is corpus breadth and the first real reportable run.

PR #135, opened 2026-08-13
feat(corpus): freeze heterogeneous v2 task sets

YoungJinJung:
## Why

#21 has a resumable executor, but the v1 dev and validation sets each contained only one task. That is not broad enough for the first meaningful multi-task control/guidance run.

Related: #21

## What changed

- preserve the immutable v1 sets and add disjoint `dev-v2` / `validation-v2` sets with three tasks each
- add four mechanically scored fixtures covering missing validation, regression avoidance, evidence inspection, and a mutation-sensitive multi-file contract
- retain the existing localized parser fix and profile contract, yielding six tasks across five failure families
- freeze fixture and task-manifest hashes and prove every new scorer against both broken and known-good states
- document v2 as the current tuning/held-out split and update status without claiming intervention evidence

## Validation

- `ruff format --check .`
- `ruff check .`
- `mypy src tests`
- `pytest` (423 passed)
- `spotter tasks preflight corpus/dev-v2.toml` (3/3 `READY`)
- `spotter tasks preflight corpus/validation-v2.toml` (3/3 `READY`)

## Notes

No paid agent arms were run. This PR freezes the broader corpus; the first real reportable control/guidance batch remains the final #21 evidence step.

### DIFF
--- a/corpus/README.md
+++ b/corpus/README.md
@@ -1,12 +1,16 @@
 # Spotter task corpus
 
-`dev-v1.toml` is for harness and supervision tuning. `validation-v1.toml` is the frozen held-out split for the first decision-quality measurements. Changing a fixture or task manifest requires a new set version and new hashes; do not rewrite an observed validation set in place. In-place re-freezing is allowed only before the set has been used in a recorded run.
+`dev-v2.toml` is the current harness/supervision tuning set. `validation-v2.toml` is the current frozen held-out split for the first decision-quality measurements. Each contains three disjoint tasks; together they cover localized fixes, missing validation, regression avoidance, evidence inspection, and multi-file contracts. The v1 sets remain immutable for provenance.
+
+Changing a fixture or task manifest requires a new set version and new hashes; do not rewrite an observed validation set in place. In-place re-freezing is allowed only before the set has been used in a recorded run.
 
 Static freeze validation is safe for untrusted input:
 
 ```bash
 spotter tasks validate corpus/dev-v1.toml
 spotter tasks validate corpus/validation-v1.toml
+spotter tasks validate corpus/dev-v2.toml
+spotter tasks validate corpus/validation-v2.toml
 ```
 
 Preflight executes the repo-authored setup and scorer commands in temporary fixture copies:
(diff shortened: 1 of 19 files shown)


YoungJinJung:
## Review

Part of #21 (+284/−8): disjoint `dev-v2`/`validation-v2` sets (3 tasks each), four new mechanically scored fixtures, v1 preserved immutable for provenance, CI-enforced freeze extended with per-set task counts and a dev/validation disjointness test. Verified on the branch: full suite 423/423 (which preflights all four repo sets to READY), `mypy` clean, `ruff` clean, PR CI green.

### Scorer logic — hand-checked all four, all sound
- **cache-regression**: the third assert (`"users: abc "`) is the mutation-sensitive one — it catches an agent that over-normalizes item keys while fixing the namespace, which is exactly what "regression-avoidance" should test. Broken state fails assert 1; known-good preserves item-key case/whitespace.
- **settings-validation**: the invalid loop covers `0/-1/"many"/None`; the broken fixture leaks `TypeError` on `None` (uncaught → check fails, correctly) and the known-good wraps it into the contractual `ValueError`. Also checks the default and endpoint trimming.
- **routing-investigation**: `"/health//"` must stay `None` — the known-good's single `removesuffix("/")` honors that, so an agent that loops slash-stripping fails. Genuine evidence-inspection: the normalization contract is only learnable from the route table.
- **record-contract**: re-asserting `raw` equality *after* the call catches in-place mutation — a contract dimension none of the earlier fixtures tested. Dedup order, blank-discard, and empty-tags rendering all discriminate.

### Minor (non-blocking)
- The `known_good` one-liner `python3 -c` blobs are getting dense (record-contract's is a triple-escaped two-file rewrite). A `fixture.good/` directory convention with a `cp` transform would freeze the same content legibly — worth adopting at v3 rather than retrofitting.
- "Heterogeneous" in the title means failure families, not ecosystems — all six tasks are single-check synthetic Python. The docs already disclaim this correctly ("larger heterogeneous corpus" remains future work), so just noting the title reads slightly bigger than the diff.

### Verdict
Approve — merging. The corpus went from 2 single-task sets to 6 tasks across 5 failure families with the freeze discipline intact; the only remaining #21 item is the first real reportable run.

Issue #42, opened 2026-08-11
Validate replay/fork fidelity and branch coverage


YoungJinJung:
### Question / hypothesis

Are Spotter's forked continuations faithful enough, and branchable early enough, to support causal claims about intervention? The fork instrument needs a measured noise floor before its deltas are interpreted.

### Experiment design

Run four checks:

1. **Identical-arm noise:** fork the same prefix into two neutral/control continuations and measure outcome spread.
2. **Continuation similarity:** fork a prefix from a completed session and compare the continuation with what the original run actually did.
3. **Environment drift:** record failures caused by detached-worktree differences such as missing ignored files, caches, virtualenvs, or absolute paths.
4. **Branch coverage:** report which trajectory steps are forkable, including early read-only/pre-mutation exploration. If patch-only snapshots leave an early blind region, add the smallest checkpoint change needed to make that region reachable.

### Implementation references — study, do not depend on

**Shepherd** is the strongest reference for treating execution traces and reversible environment state as first-class experiment infrastructure:

- runtime substrate: https://github.com/shepherd-agents/shepherd
- experiment repository: https://github.com/shepherd-agents/shepherd-experiments
- live-intervention experiment: https://github.com/shepherd-agents/shepherd-experiments/tree/main/exp/live-intervention

Shepherd's model differs from Spotter—it owns execution and uses its own reversible substrate—but it is useful for three design lessons:

- branch/revert/checkpoint operations must be explicit runtime objects, not ad-hoc directory copies;
- environment state and agent/model continuation state are separate sources of fidelity/error;
- framework-performance/fork mechanics should be benchmarked independently from the intervention experiment that consumes them.

Spotter should reuse those lessons with its existing Git snapshot/worktree/replay machinery, not introduce Shepherd as a dependency.

### Separate three fidelity questions

Do not collapse replay fidelity into one score.

```text
ENVIRONMENT FIDELITY
  Did both branches begin from materially equivalent filesystem/repo/tool state?

CONTEXT FIDELITY
  Did the continued agent receive the intended user goal/history/state available at the prefix?

STOCHASTIC CONTINUATION NOISE
  Even with equivalent state/context, how much do independent model continuations differ?
```

A neutral-vs-neutral disagreement caused by model stochasticity is not a snapshot bug. A missing ignored config file is not model variance. Report them separately.

### Prefix manifest

Every forkable prefix used for experiments should resolve to an immutable/versioned manifest such as:

```text
prefix_id
source_thread/run_id
[...]

PR #136, opened 2026-08-13
experiment(tasks): report first dev v2 batch

### LINKED ISSUE #21: Build a mechanically scored task set

YoungJinJung:
## Question / hypothesis

Can the frozen #21 corpus and resumable runner complete a real multi-task control/guidance experiment with mechanically judgeable outcomes and no infrastructure failures entering task-success counts? A generic evidence-first nudge may preserve or improve success, but this three-task dev run is an instrument qualification rather than an intervention-effect measurement.

Closes #21

## Method

- froze the protocol in the report before execution
- ran `dev-v2` only; `validation-v2` remains held out
- executed one control and one guidance arm for each of three tasks from independent clean copies
- used the generic guidance: inspect the failing check/relevant code, preserve unaffected behavior, make the smallest justified fix, then run the task check
- judged outcomes only with frozen required scorers
- retained the raw Spotter JSONL locally and committed a transcript-free per-arm derived result tied to its SHA-256

## Evidence

- executed/mechanically judged arms: 6/6
- usable pairs: 3/3
- guidance better: 0
- control better: 0
- tie-success: 3
- infrastructure failures: 0
- control/guidance agent-reported tokens: 41,912 / 39,270 (descriptive only)
- control/guidance summed elapsed seconds: 96.72 / 96.77
- all six arms emitted non-fatal MCP shutdown warnings, but agent/setup/check exits were 0
- raw JSONL SHA-256: `948226a9d1fcd768b69023b22d0a2fedc121bc058220e8550bdbc6bf4fb9f434`
- full repository validation: `ruff format --check .`, `ruff check .`, `mypy src tests`, `pytest` (423 passed)
- derived JSON was programmatically checked against all six raw arm rows and passed the repository secret-pattern scan

## Decision impact

#21 is complete as an evaluation foundation: freeze, preflight, isolated execution, classification, resume, and a real multi-task report all work. The observed result is a small dev-set null/tie, not evidence of intervention advantage. Do not change live-intervention defaults or tune the held-out set. Measure replay noise under #42 before interpreting later same-prefix intervention effects.

### DIFF
--- a/docs/experiments/dev-v2-first-run.json
+++ b/docs/experiments/dev-v2-first-run.json
@@ -1,0 +1,119 @@
+{
+  "schema_version": 1,
+  "run_id": "16532470-1be0-41e8-a0ea-b52f6f92b0a8",
+  "task_set": {
+    "id": "spotter-dev",
+    "version": 2,
+    "sha256": "ed6e398fa569b7dc556403cc6ca8c1cbf2b1c31f1eca0f4c76551d830edc0677"
+  },
+  "condition": {
+    "control_suffix": "Continue the task.",
+    "guidance_suffix": "Before editing, inspect the failing check and relevant code. Preserve unaffected behavior, make the smallest justified fix, and run the task check afterward."
+  },
+  "environment": {
+    "requested_model": "codex-config-default",
+    "resolved_rollout_model": "gpt-5.6-sol",
+    "resolved_from_rollout_headers": 6,
+    "codex_version": "codex-cli 0.147.0",
+    "python": "3.12.13",
+    "platform": "macOS-26.5.1-arm64-arm-64bit",
+    "sandbox": "workspace-write"
+  },
+  "started_at": "2026-08-13T10:42:10.078014+00:00",
+  "finished_at": "2026-08-13T10:45:23.578849+00:00",
+  "elapsed_seconds": 193.500835,
+  "summary": {
+    "executed_arms": 6,
+    "mechanically_judged_arms": 6,
+    "usable_pairs": 3,
+    "guidance_better": 0,
+    "control_better": 0,
+    "tie_success": 3,
+    "tie_failure": 0,
+    "infrastructure_failures": 0,
+    "non_fatal_mcp_shutdown_warning_arms": 6,
+    "control_agent_reported_tokens": 41912,
+    "guidance_agent_reported_tokens": 39270,
+    "control_elapsed_seconds": 96.720279,
+    "guidance_elapsed_seconds": 96.772302
+  },
+  "arms": [
+    {
+      "task_id": "fixture/query-parser-001",
+      "arm": "control",
+      "classification": "PASS",
+      "setup_exit": 0,
+      "agent_exit": 0,
+      "required_check_exits": [0],
+      "agent_reported_tokens": 12595,
+      "started_at": "2026-08-13T10:42:10.078468+00:00",
+      "ended_at": "2026-08-13T10:42:38.696471+00:00",
+      "non_fatal_mcp_shutdown_warning": true
+    },
+    {
+      "task_id": "fixture/query-parser-001",
+      "arm": "guidance",
+      "classification": "PASS",
+      "setup_exit": 0,
+      "agent_exit": 0,
+      "required_check_exits": [0],
+      "agent_reported_tokens": 13511,
+      "started_at": "2026-08-13T10:42:38.697484+00:00",
+      "ended_at": "2026-08-13T10:43:04.616268+00:00",
+      "non_fatal_mcp_shutdown_warning": true
+    },
+    {
+      "task_id": "fixture/settings-validation-001",
+      "arm": "guidance",
+      "classification": "PASS",
+      "setup_exit": 0,
+      "agent_exit": 0,
+      "required_check_exits": [0],
+      "agent_reported_tokens": 16401,
+      "started_at": "2026-08-13T10:43:04.619105+00:00",
+      "ended_at": "2026-08-13T10:43:46.717175+00:00",
+      "non_fatal_mcp_shutdown_warning": true
+    },
+    {
+      "task_id": "fixture/settings-validation-001",
+      "arm": "control",
+      "classification": "PASS",
+      "setup_exit": 0,
+      "agent_exit": 0,
+      "required_check_exits": [0],
+      "agent_reported_tokens": 14749,
+      "started_at": "2026-08-13T10:43:46.718189+00:00",
+      "ended_at": "2026-08-13T10:44:25.512717+00:00",
+      "non_fatal_mcp_shutdown_warning": true
+    },
+    {
+      "task_id": "fixture/cache-regression-001",
+      "arm": "control",
+      "classification": "PASS",
+      "setup_exit": 0,
+      "agent_exit": 0,
+      "required_check_exits": [0],
+      "agent_reported_tokens": 14568,
+      "started_at": "2026-08-13T10:44:25.513696+00:00",
+      "ended_at": "2026-08-13T10:44:54.821444+00:00",
+      "non_fatal_mcp_shutdown_warning": true
+    },
+    {
+      "task_id": "fixture/cache-regression-001",
+      "arm": "guidance",
+      "classification": "PASS",
+      "setup_exit": 0,
+      "agent_exit": 0,
+      "required_check_exits": [0],
+      "agent_reported_tokens": 9358,
+      "started_at": "2026-08-13T10:44:54.822417+00:00",
+      "ended_at": "2026-08-13T10:45:23.577865+00:00",
+      "non_fatal_mcp_shutdown_warning": true
+    }
+  ],
+  "raw_result": {
+    "spotter_home_relative_path": "experiments/task-batches/spotter-dev-v2-16532470-1be0-41e8-a0ea-b52f6f92b0a8.jsonl",
+    "sha256": "948226a9d1fcd768b69023b22d0a2fedc121bc058220e8550bdbc6bf4fb9f434",
+    "committed": false
+  }
+}
--- a/docs/experiments/dev-v2-first-run.md
+++ b/docs/experiments/dev-v2-first-run.md
@@ -1,0 +1,98 @@
+# Dev v2 first multi-task run
+
+Date: 2026-08-13
+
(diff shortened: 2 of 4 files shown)


YoungJinJung:
## Review

Closes #21 (+233/−7, docs + derived JSON only): the first real reportable multi-task control/guidance run — three dev-v2 tasks, six paid arms, 6/6 mechanically judged, 3/3 usable pairs, all tie-success. Verified on the branch: full suite 423/423, **`corpus/dev-v2.toml` SHA-256 matches the report's claimed set hash exactly** (`ed6e398f…`), and the derived JSON is internally consistent (per-arm token sums equal the summary totals, arm ordering shows the runner's counterbalancing — task 2 ran guidance-first — and per-arm elapsed sums match total elapsed for sequential execution).

### The epistemics are the best part
- Protocol and decision rule frozen in the report *before* execution; the run is framed as **instrument qualification**, and the decision rule was about coverage and classification integrity, not effect direction.
- `validation-v2` stays held out; the null/tie is explicitly barred from changing live-intervention defaults or tuning the held-out set; #42 (replay noise floor) is named as the prerequisite for interpreting any future effect. Cost numbers are labeled descriptive-only.
- The raw JSONL stays out of the repo (bounded transcripts) with its SHA-256 committed for provenance — the same value-free discipline as the #120 source audit, applied to experiment artifacts.
- The evidence tables move honestly: "reproducible corpus" → **Yes for the synthetic v2 foundation** with breadth as the named gap; "positive intervention advantage" stays ❌; the non-fatal MCP shutdown warnings are surfaced as an operational diagnostic rather than hidden.

### One methodological note for the next run (non-blocking)
3/3 tie-success is a **ceiling effect**: when both arms pass everything, the instrument has zero discriminating power — you'd get the same table whether guidance helps, hurts, or does nothing. The report calls the tasks "easy" but doesn't name the implication: the next corpus iteration needs tasks with a meaningful control-arm failure rate (harder tasks, tighter budgets, or one-shot constraints), or the paired design will keep returning uninformative ties regardless of the true effect. Worth a line in the #42/next-corpus planning.

### Verdict
Approve — merging. #21 closes with the machinery *and* a real run that demonstrates exactly what it claims and nothing more.

PR #137, opened 2026-08-13
feat(replay): persist fork provenance and preflight parity

YoungJinJung:
## Summary
- persist versioned fork manifests with immutable shared-prefix lineage, context limitations, external-effect/gap metadata, and captured environment fingerprints
- prepare both counterfactual forks before execution and refuse both arms when prefix or captured environment parity fails
- record fork manifest, prefix, fingerprint, and preflight results in experiment rows; document captured and explicitly uncaptured state

This is the first bounded implementation slice of #42. Neutral-vs-neutral noise measurement, broader drift categories/resources, and early branch-coverage qualification remain in the issue.

## Validation
- `.venv/bin/ruff check .`
- `.venv/bin/mypy src tests`
- `.venv/bin/pytest` (428 passed)

Related to #42.


YoungJinJung:
## Review

First bounded slice of #42 (+720/−52): durable fork manifests with a `CREATING → READY/FAILED` lifecycle, content-addressed shared-prefix identity, a captured-environment fingerprint with *explicitly declared* non-capture, and pair-parity preflight that refuses both arms before either runs. Verified on the branch: full suite 428/428, `mypy` clean, `ruff` clean.

### Strengths
- **`prefix_id` is earned, not asserted**: it digests canonical identity material including `rollout_prefix_sha256` — the exact rollout bytes before the branch call — so "same prefix" is a hash equality over what the agent will actually resume from, not an assumption from shared session/step numbers.
- **Non-capture encoded as data**: `ignored_state: not_captured_by_git_snapshot` and `environment_variables: not_captured` live *inside* the fingerprint material, and `context_limitations` names the reviewer-state and env-var gaps in the prefix manifest. The instrument documents its own blind spots in the artifact, not just the docs — the #120 discipline applied to replay.
- **Both-forks-first preflight closes the classic sequential-fork bug**: prefix or captured-environment mismatch fails both arms as `INFRA_FAIL` with the drift class in the diagnostic, so a contaminated pair can never contribute to the effect table. The `--binary --no-ext-diff --no-textconv` diff flags also dodge textconv nondeterminism in the fingerprint.
- Manifest writes are `O_EXCL` + fsync + dir-fsync with atomic replace, the failure text passes through `redact_text` before persisting, and a crash leaves a `CREATING` manifest for forensics. Tests cover shared-prefix equivalence, drift classification, and gap/effect limits.

### Minor (non-blocking)
- **`ArmResult` gained four fields without bumping `EXPERIMENT_RESULT_SCHEMA_VERSION`** — the version was introduced in #132 precisely for row evolution. If the contract is "bump on breaking only," one comment saying so ends the ambiguity; otherwise this is the moment for 1→2.
- On a late fork failure (restore succeeded, fingerprint raised) the forked rollout is intentionally kept and referenced by the FAILED manifest — reasonable as a debugging artifact, but worth remembering when #89 does retention, since FAILED manifests now pin rollout files.
- `run_experiment`'s arm loop is now four levels of try/except deep — the next slice that touches it should extract `_run_single_arm` before adding anything.

### Verdict
Approve — merging. The #42 remainder (neutral-vs-neutral noise measurement) now has exactly the provenance substrate it needs: identical `prefix_id`, matched fingerprints, and refusal semantics for everything else.

PR #138, opened 2026-08-13
feat(experiment): measure identical-arm replay noise

YoungJinJung:
## Summary
- add an explicit `spotter experiment --neutral` mode that runs both forks with the exact same control prompt
- report mechanically judged outcome disagreement separately from environment-preflight mismatches and infrastructure failures
- extract single-arm execution from the pair loop and bump experiment result schema to v2

This is the next bounded slice of #42. It adds the measurement path but does not claim a noise bound; representative real-prefix runs and branch-coverage qualification remain.

## Validation
- `.venv/bin/ruff check .`
- `.venv/bin/mypy src tests`
- `.venv/bin/pytest` (431 passed)

Related to #42.


YoungJinJung:
## Review

Next #42 slice (+236/−77) — and both #137 review asks are closed in the same PR: `EXPERIMENT_RESULT_SCHEMA_VERSION` bumped to 2 for the added row fields, and the four-level arm loop is extracted into `_execute_arm` before new behavior was added, exactly as suggested. Verified on the branch: full suite 431/431, `mypy` clean, `ruff` clean.

### Strengths
- **The extraction is behavior-preserving** — I traced every classification path (agent timeout/OSError, nonzero exit, no-check UNJUDGEABLE, check timeout/error/pass/fail) against the old inline flow; identical outcomes, now readable.
- **Neutral mode measures what it claims**: both arms get the byte-identical `CONTROL_PROMPT` (pinned by the prompts assertion), arms are `neutral_a/b` so guidance-mode comparisons can't accidentally consume noise rows, and the summary separates mechanical outcome disagreement, environment-preflight mismatch, and infrastructure failure with explicit denominators — the disagreement rate renders `unknown` when zero pairs are judgeable rather than `0.0%`, the established unknown-not-zero discipline.
- The docs place the epistemics correctly in three files: the rate "becomes evidence only after representative real prefixes are executed" — the measurement path exists, the bound is not claimed.
- Mode travels in both the meta header and every row (`experiment_mode`), so mixed archives stay distinguishable.

### Minor (non-blocking)
- `--neutral --guidance "x"` together is accepted silently: arms run neutral, but the meta header records the guidance text — a neutral-noise artifact carrying a guidance string is confusing provenance. Either `parser.error` on the combination or null out `guidance` in neutral mode; one line either way.
- `summarize` switches to neutral labels if *any* row is neutral-mode — fine for per-experiment use, but a concatenated mixed-mode list would mislabel. Not worth code until someone actually concatenates; noting for the eventual reporting tool.

### Verdict
Approve — merging. #42 now needs only the representative real-prefix executions; the instrument, parity refusal, and honest rate reporting are all in place.

PR #139, opened 2026-08-13
feat(replay): checkpoint session baseline for early forks

YoungJinJung:
## Summary
- pin a deduplicated Git baseline snapshot at Codex `SessionStart` so pre-mutation read-only proposals have a fork anchor
- verify the full `SessionStart -> first read-only proposal -> fork` path and preserve fail-open behavior outside Git repositories
- reject ambiguous `--neutral --guidance` experiments so neutral artifacts cannot carry guidance provenance

This is another bounded #42 slice. Historical sessions whose snapshot objects are already unavailable remain non-recoverable; representative neutral executions and the coverage report still remain.

## Validation
- `.venv/bin/ruff check .`
- `.venv/bin/mypy src tests`
- `.venv/bin/pytest` (435 passed)

Related to #42.

### DIFF
--- a/README.md
+++ b/README.md
@@ -360,7 +360,9 @@ spotter experiment --session <id> --step <n> --guidance "..." --check "..."
 spotter experiment --session <id> --step <n> --neutral --pairs 3 --check "..." --run
 ```
 
-`spotter fork` writes a durable manifest under `~/.spotter/fork-manifests/`. The manifest links the
+When snapshotting is enabled, `SessionStart` pins one baseline Git snapshot so read-only exploration
+before the first mutation has a fork anchor; mutation boundaries continue to add deduplicated
+snapshots. `spotter fork` writes a durable manifest under `~/.spotter/fork-manifests/`. The manifest links the
 source event, snapshot, rollout-prefix digest, external-effect warnings, and captured environment
 fingerprint. Counterfactual pairs are fully prepared and must pass shared-prefix and captured-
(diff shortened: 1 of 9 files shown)


YoungJinJung:
## Review

Another bounded #42 slice (+108/−9) — and the #138 minor is closed in the same PR. Verified on the branch: full suite 435/435, `mypy` clean, `ruff` clean.

### Strengths
- **The pre-mutation blind region is gone at minimal cost**: one baseline snapshot per `SessionStart` (reusing the existing global-lock + tree-dedup path, so an untouched worktree at first mutation dedups against the baseline), and `_source_snapshot`'s reverse scan picks it up naturally — no fork-side changes needed. The E2E test walks the actual promise: `SessionStart` → read-only proposal → `fork` at step 1 → restored worktree carries baseline content.
- **Fail-open is preserved and tested**: a non-repo `cwd` at SessionStart journals with `snapshot: None` instead of blocking, keeping the hook's "unsupervised beats blocked" contract.
- **`--neutral --guidance` is now rejected at both layers** (CLI `parser.error` + library `ValueError`), each with its own test — neutral artifacts can no longer carry guidance provenance, exactly the #138 ask.
- Honest scope note carried into the docs: historical sessions whose snapshot objects were pruned remain non-recoverable; this fixes the future, not the past.

Nothing to object to — small, complete, tested at the boundary that matters.

### Verdict
Approve — merging.

PR #140, opened 2026-08-13
feat(replay): report branch coverage eligibility

YoungJinJung:
## Summary
- add a read-only `spotter fork-coverage --session` report that classifies every proposal by current snapshot, rollout-context, external-effect, and observation-gap eligibility
- report earliest exact fork plus pre-mutation coverage, while conservatively excluding legacy unclassified prefixes
- record the eight-session historical baseline: 0/1,246 clean exact proposals and a NO-GO for neutral execution on that legacy sample

This is another #42 slice. Fresh post-baseline-checkpoint sessions are still required before representative neutral runs can establish a noise bound.

## Validation
- `.venv/bin/ruff check .`
- `.venv/bin/mypy src tests`
- `.venv/bin/pytest` (437 passed)

Related to #42.


YoungJinJung:
## Review

Another #42 slice (+337/−2): a read-only `spotter fork-coverage` classifier with conservative causal-safety semantics, and — the more valuable half — a committed historical baseline reporting **0/1,246 clean exact proposals as a documented NO-GO**. Verified on the branch: full suite 437/437, `mypy` clean, `ruff` clean; the baseline table's arithmetic checks out (candidates 1,246, state 6, context 2, unsafe 1,238 — all row sums match), and `_recorded_repo` is the existing shared helper, reused rather than reimplemented.

### Strengths
- **The causal ordering subtlety is handled exactly right**: the branch point sits *before* a proposal, so a proposal's own possible effect contaminates later points, not its own — commented in code and pinned by both tests (a Class B proposal is judged at its own point; the legacy-unclassified proposal is itself `FORKABLE_EXACT` while everything after it degrades to `UNSAFE_EXTERNAL_EFFECT`).
- **Conservatism is the load-bearing feature**: unclassified legacy prefixes poison forward permanently, snapshot objects are verified *now* via `git cat-file -e` (cached) so pruned repositories read `NOT_FORKABLE_STATE` instead of counting as covered, and the classification priority (gap > effect > context > state) resolves overlaps toward ineligibility.
- **The baseline is a model null-result report**: it explicitly says `UNSAFE_EXTERNAL_EFFECT` means *unknown or unsafe*, "not proof that 1,238 external writes occurred"; it declares NO-GO for neutral execution on the legacy sample rather than fabricating a zero-noise number; and it names the exact next qualification (fresh post-#139 sessions, then `--neutral` only on `FORKABLE_EXACT` prefixes). This closes the loop with #139 cleanly: the blind region fix ships one PR, the instrument that proves the old data was blind ships the next.

### Minor (non-blocking)
- The unclassified-proposal contamination never resets — correct today, but if a future migration back-fills reversibility classes onto legacy journals, the report will still condemn them until this line changes; a half-line comment noting that coupling would help the future migrator.
- The JSON report embeds the full per-point list (600 points for the largest session) — fine for a diagnostic; if it ever feeds dashboards, a `--summary` flag is the obvious trim.

### Verdict
Approve — merging. #42 now has instrument + parity + noise mode + coverage eligibility + an honest baseline saying "not with this data" — everything except the fresh sessions themselves.

Issue #34, opened 2026-08-11
Measure intervention advantage, harm, and recovery


YoungJinJung:
### Question / hypothesis

When Spotter intervenes on a shared trajectory prefix, does it improve the outcome more often than it harms a continuation that would otherwise have succeeded?

### Experiment design

Using the mechanically scored task set (#21) and a replay instrument validated in #42, compare the same prefix under:

- control: continue without guidance
- intervention: `VERIFY` or `NUDGE` first; stronger actions only after Recover exists

Classify each pair as:

- intervention better
- control better (**harm**)
- tied
- incomplete / infrastructure failure

When live delivery exists, also record whether the intervention was delivered, ignored, acted on, and whether the trajectory recovered.

### Implementation references — study, do not depend on

**Shepherd's live-intervention experiment** is the closest public implementation reference for measuring a supervisor rather than merely building one:

- experiment: https://github.com/shepherd-agents/shepherd-experiments/tree/main/exp/live-intervention
- experiment README/conditions: https://github.com/shepherd-agents/shepherd-experiments/blob/main/exp/live-intervention/README.md
- meta-agent implementation: https://github.com/shepherd-agents/shepherd-experiments/blob/main/exp/live-intervention/src/experiment_live_intervention/core/meta_agent_llm.py
- batch/evaluation code: https://github.com/shepherd-agents/shepherd-experiments/tree/main/exp/live-intervention/src/experiment_live_intervention/core

Useful patterns:

- compare named experimental conditions, not anecdotes;
- keep intervention budgets/timing explicit;
- retain raw run data and grade afterward;
- separate execution backend from grading;
- use held-out task sets after prompt/supervisor development.

Spotter needs a stricter causal design than Shepherd's headline managed-vs-peer comparison because this issue asks a **same-prefix intervention advantage** question.

**ECLoop** is also useful as evidence that runtime intervention can improve both SWE success and efficiency, but its aggregate gains do not answer whether a particular Spotter intervention helped a trajectory that would otherwise have succeeded:

- paper: https://arxiv.org/abs/2607.28815

These are references only; no Shepherd/ECLoop dependency should be introduced.

### Define the causal unit as a paired prefix

The unit of analysis should be:

```text
Pair P
  task T
  shared trajectory prefix X
  shared environment/checkpoint E

  control arm      X → continue
  intervention arm X → Spotter action → continue
```

[...]

PR #141, opened 2026-08-13
docs(experiment): report first neutral fork run

YoungJinJung:
## Summary
- report the first fresh exact-prefix identical-arm run: 3/3 judgeable pairs, 0/3 mechanical disagreements, 0/3 environment mismatches, and 0/6 infrastructure failures
- preserve a derived JSON artifact plus the raw local JSONL SHA-256 and per-continuation token/tool/time diagnostics
- keep #42 open: this validates one repeated prefix but exposes a legacy Hook registration gap and is not a representative noise bound

## Validation
- `jq empty docs/experiments/fork-neutral-first-run.json`
- `.venv/bin/ruff check .`
- `.venv/bin/mypy src tests`
- `.venv/bin/pytest` (437 passed)

Related to #42.


YoungJinJung:
## Review

Docs-only #42 report (+138/−1): the first fresh exact-prefix identical-arm run — 3/3 judgeable pairs, 0/3 mechanical disagreements, 0/3 environment mismatches, 0/6 infrastructure failures. Verified on the branch: suite passes, the derived JSON is internally consistent (6 continuations, 3 pairs, arm sets correct), and the pair-first arm order `b, a, b` shows the runner's counterbalancing operating as designed.

### Strengths
- **Scoped exactly right**: "zero observed mechanical disagreement for one repeated prefix, not a global zero-noise claim" — and the token/elapsed variance (102,801–135,073 tokens, 17–25s at identical tool counts) is surfaced as evidence that stochastic variance exists *under* outcome agreement, rather than hidden because the headline number was clean.
- **The run exposed a real operational gap and reported it against itself**: pre-mutation coverage was 0/2 because this machine's legacy Hook registration omits `SessionStart`, so the #139 baseline checkpoint never ran. Naming your own registration drift as the blocker for early-intervention claims — instead of quietly re-registering and rerunning — is the discipline that makes these reports trustworthy.
- Spend discipline: coverage was measured *before* paying for repetitions, and only the one `FORKABLE_EXACT` point was executed. The historical NO-GO doc gains a cross-reference that explicitly does not reopen its decision.

### One methodological note (non-blocking, echoing #136)
6/6 passing arms is again a ceiling regime: disagreement can only manifest when the task has meaningful failure probability at the branch point, so this zero bounds noise only on the always-pass side. The report's own "bound outcome disagreement rather than merely observe zero once" gestures at this; when selecting the next prefixes, branch points where the continuation *sometimes fails* are worth deliberate inclusion — otherwise repeated zeros will accumulate without narrowing the bound that #34/#23 actually need.

### Verdict
Approve — merging. Also worth a maintainer follow-up outside this PR: fixing the legacy Hook registration (`SessionStart` missing) on the measurement machine, since it currently gates all pre-mutation coverage claims.

PR #142, opened 2026-08-13
fix(integration): retain session-start hook

YoungJinJung:
## Bug

Standalone Codex setup removed legacy observation Hooks and retained only `PreToolUse`, so `SessionStart` never reached Spotter. That prevented the baseline checkpoint required for early pre-mutation forks in #42.

Related to #42.

## Root cause

The integration manifest and migration path modeled exactly one owned Hook even though `SessionStart` remains required until #86 proves App Server lifecycle parity.

## Fix

- register and transactionally own both `PreToolUse` and `SessionStart` while preserving unrelated commands
- upgrade schema 1 manifests to schema 2 `owned_hooks` in memory and reconcile them on the next setup
- validate and tear down the complete owned Hook set
- document the temporary lifecycle contract until #86

## Verification

- `.venv/bin/ruff check .`
- `.venv/bin/mypy src tests`
- `.venv/bin/pytest` (438 passed)

### DIFF
--- a/docs/lifecycle.md
+++ b/docs/lifecycle.md
@@ -387,6 +387,7 @@ Plan
   App Server strategy: codex-managed-external
   Runtime mode: managed/login-scoped
   Hooks:
+    add SessionStart
     add PreToolUse
     remove legacy PostToolUse
     remove legacy UserPromptSubmit
@@ -462,7 +463,7 @@ Example manifest:
 
 ```json
 {
-  "schema": 1,
+  "schema": 2,
   "agent": "codex",
   "setup_by": "0.6.0",
   "agent_path": "/opt/homebrew/bin/codex",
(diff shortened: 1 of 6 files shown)


YoungJinJung:
## Review

The root-cause fix for the operational gap the #141 report exposed (+99/−49): standalone setup owned exactly one Hook (`PreToolUse`), so `SessionStart` never reached Spotter and the #139 baseline checkpoint couldn't run. Verified on the branch: full suite 438/438, `mypy` clean, `ruff` clean.

### Strengths
- **Fixed at the model, not the symptom**: the manifest's single `owned_hook` becomes an `owned_hooks` list with a schema 1→2 bump — the fix lands in the ownership contract where every future setup/teardown/doctor flows through it, rather than a one-off re-registration on the measurement machine. This is exactly the migration path the #141 review asked for.
- **The upgrade is reconciling, not just tolerant**: a schema-1 manifest loads with the SessionStart entry synthesized from the existing hook command, and the dedicated test proves the full loop — old manifest + hooks.json missing SessionStart → next `setup` upgrades to schema 2 *and* registers both Hooks. Teardown iterates the list, so upgraded manifests clean up completely (and removing a not-yet-registered SessionStart is a harmless no-op).
- `doctor`'s ownership validation moved to the list shape with its FAIL messages updated, and the docs name the contract honestly: SessionStart is *temporary* until #86 proves App Server lifecycle parity — the hook isn't quietly becoming permanent.

### Minor (non-blocking)
- The ownership comparisons (`_verify`, `check_integration`) are **order-sensitive lists**, and they currently agree only because `_write_hooks` serializes with `sort_keys=True` and `[PreToolUse, SessionStart]` happens to be alphabetical. A future owned event sorting before `PreToolUse` would produce a false "drifted" FAIL with no actual drift. Comparing as sets of (event, matcher, hook) tuples — or sorting both sides — removes the coupling for free.

### Verdict
Approve — merging. With this, the #141 pre-mutation coverage blocker is structurally closed; fresh sessions on standalone setups will get their baseline checkpoint.

PR #143, opened 2026-08-13
fix(integration): ignore owned hook order

YoungJinJung:
## Bug

PR #142 made Hook ownership a list, but verification compared that list in serialization order. A valid manifest order differing from sorted `hooks.json` order could be reported as drift.

Follow-up to #142. Related to #42.

## Root cause

Both setup verification and doctor used order-sensitive list equality for an unordered Hook ownership set.

## Fix

Canonicalize each `(event, matcher, hook)` entry and compare sorted values while retaining duplicate detection.

## Verification

- regression coverage for reversed setup/manifest order and duplicate detection
- `.venv/bin/ruff check .`
- `.venv/bin/mypy src tests`
- `.venv/bin/pytest` (440 passed)

### DIFF
--- a/src/spotter/doctor.py
+++ b/src/spotter/doctor.py
@@ -287,7 +287,9 @@ def check_integration() -> IntegrationInspection:
             False,
         )
     expected = [(entry.get("event"), entry.get("matcher"), entry.get("hook")) for entry in owned]
-    if spotter_hooks != expected:
+    if sorted(json.dumps(entry, sort_keys=True) for entry in spotter_hooks) != sorted(
+        json.dumps(entry, sort_keys=True) for entry in expected
+    ):
         return IntegrationInspection(
             manifest,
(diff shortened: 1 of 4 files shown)


YoungJinJung:
## Review

Closes the #142 review's order-sensitivity finding (+28/−2). Verified on the branch: full suite 440/440, `mypy` clean, `ruff` clean.

The chosen fix is better than the one I suggested: canonicalizing entries via `json.dumps(sort_keys=True)` and comparing **sorted multisets** removes the ordering coupling while *preserving duplicate detection* — a plain `set` comparison would have silently collapsed duplicates, weakening the drift check. Both comparison sites (setup `_verify` and doctor `check_integration`) are fixed symmetrically, with regression tests for reversed order at each layer plus the existing duplicate-detection test still holding.

Approve — merging.

PR #144, opened 2026-08-13
fix(integration): retain observation hooks before parity

YoungJinJung:
## Bug

Standalone setup removed `UserPromptSubmit` and `PostToolUse` even though no App Server endpoint was configured and #86 has not proven responsibility parity. Applying setup would therefore reduce observation coverage while trying to enable fresh #42 sessions.

Related to #86 and #42.

## Root cause

The integration lifecycle implemented the target minimal Hook set before its documented measurement gate had passed.

## Fix

- own all four current observation/enforcement Hooks until #86 explicitly removes the redundant three
- bump the integration manifest to schema 3 and reconcile schema 1/2 installations to the complete Hook set
- keep setup, teardown, doctor fixtures, lifecycle docs, and current status aligned

## Verification

- schema 1 and schema 2 reconciliation regressions
- setup/setup/teardown/setup coverage preserves unrelated commands
- `.venv/bin/ruff check .`
- `.venv/bin/mypy src tests`
- `.venv/bin/pytest` (441 passed)

### DIFF
--- a/docs/lifecycle.md
+++ b/docs/lifecycle.md
@@ -388,9 +388,9 @@ Plan
   Runtime mode: managed/login-scoped
   Hooks:
     add SessionStart
+    add UserPromptSubmit
     add PreToolUse
-    remove legacy PostToolUse
-    remove legacy UserPromptSubmit
+    add PostToolUse
   Legacy plugin: migrate
   Existing data: preserve
   Agent config backup/fingerprint: required
@@ -463,7 +463,7 @@ Example manifest:
 
 ```json
 {
-  "schema": 2,
+  "schema": 3,
   "agent": "codex",
   "setup_by": "0.6.0",
   "agent_path": "/opt/homebrew/bin/codex",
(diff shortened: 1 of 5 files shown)


YoungJinJung:
## Review

The deeper sequencing correction behind #142 (+81/−21): standalone setup was *removing* `UserPromptSubmit` and `PostToolUse` — implementing #86's target end-state before #86's parity measurement had run, which meant applying setup reduced observation coverage on a system with no App Server endpoint configured. Now all four current Hooks are owned until #86 explicitly removes the redundant three. Verified on the branch: full suite 441/441, `mypy` clean, `ruff` clean.

### Strengths
- **The root cause is named honestly**: "the integration lifecycle implemented the target minimal Hook set before its documented measurement gate had passed" — this walks back part of #110's legacy-hook removal, and correctly so; the removal was gated on evidence (#86) that doesn't exist yet. Aligning implementation with the documented gate rather than the aspiration is precisely this project's discipline.
- **The schema 1/2 → 3 upgrade path is unified and idempotent**: schema 1 first normalizes to a list, then one shared path appends whichever of the three observation entries are missing (derived from the first valid owned hook, deduplicated by full-entry equality). Both reconciliation regressions are tested — schema-1 and schema-2 manifests with stripped hooks.json each converge to the complete four-Hook set on next setup.
- Matcher semantics stay consistent (tool events `".*"`, lifecycle events `null`), and the tests moved to set comparisons — leaning on #143's order-insensitive verification rather than re-encoding an order.
- The lifecycle migration diagram now states the truth in the right tense: *retain* observation hooks until parity is measured; *remove* them under #86.

### Minor (non-blocking)
- Three schema bumps in three consecutive PRs (#142→2, here→3) is fine while unreleased, but once real installations exist, batching contract changes will matter — worth remembering when Homebrew distribution lands.

### Verdict
Approve — merging. Setup no longer trades observation coverage for tidiness it hasn't earned.

PR #145, opened 2026-08-13
fix(replay): correlate Code Mode fork calls

YoungJinJung:
Part of #42.

Summary:
- correlate Codex 0.147 Code Mode Hook execution IDs with persisted model call IDs only when the complete sequences align
- cut fork rollouts before the model call instead of at a later execution-completion event
- reject incomplete correlations and restore exact pre-mutation branch coverage

Measured result:
- real record-contract session improved from 0/2 to 2/2 forkable pre-mutation candidates
- total exact points improved from 1/5 to 4/5; the remaining point is excluded by an earlier Class C external effect

Verification:
- ruff format --check .
- ruff check .
- mypy src tests
- pytest: 443 passed

### DIFF
--- a/src/spotter/replay.py
+++ b/src/spotter/replay.py
@@ -185,14 +185,17 @@ def branch_coverage(session_id: str, codex_home: Path | None = None) -> BranchCo
     records = StepJournal.load(journal_path({"session_id": session_id}))
     try:
         rollout = find_rollout(session_id, codex_home)
+        rollout_records = _read_rollout(rollout)
         rollout_ids = {
             identifier
-            for number, line in enumerate(rollout.read_text(encoding="utf-8").splitlines(), 1)
-            for identifier in _record_ids(_rollout_record(line, number))
+            for record in rollout_records
+            for identifier in _record_ids(record)
             if isinstance(identifier, str)
         }
+        code_mode_ids = _code_mode_call_ids(records, rollout_records)
     except (OSError, ReplayError):
         rollout_ids = set()
+        code_mode_ids = None
     mutation_steps = [
         record.step
         for record in records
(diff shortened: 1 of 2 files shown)


YoungJinJung:
## Review

Part of #42 (+174/−13): fixes the root cause of the #141 report's "three proposals without rollout correlation" — Codex 0.147 Code Mode gives Hooks an `exec-…` execution id while the persisted `custom_tool_call` carries a different `call_…` id, so the old first-mention matching could never correlate them. Verified on the branch: full suite 443/443, `mypy` clean, `ruff` clean.

### Strengths
- **The correlation is order-based but refuses to guess**: a mapping exists only when the Hook-proposal sequence and the rollout's `custom_tool_call` sequence align one-to-one — any count mismatch or broken `proposal_number` continuity rejects the *entire* mapping (empty dict → `NOT_FORKABLE_CONTEXT` in coverage, `ReplayError` in fork) rather than overstating coverage. Both rejection layers are tested.
- **The cut got strictly better**: forks now truncate *before the model call* instead of at whatever line first mentioned the id — the old `ponytail:` comment about completion-event cuts is retired with its history preserved in the docstring, and the manifest records the model-level call id (pinned by test).
- Legacy rollouts (no `custom_tool_call` records) fall back to the old id matching unchanged, and fork gains an earlier, clearer error when the id has no rollout correlation.
- The measured impact is stated concretely: the real record-contract session went 0/2 → 2/2 pre-mutation forkable and 1/5 → 4/5 exact, with the last point correctly excluded by a Class C external effect.

### One theoretical note (non-blocking)
The completeness check catches any single-sided loss, but a *paired* loss — the Hook missing one proposal while the rollout independently drops one `custom_tool_call` — would leave equal-length sequences that misalign silently. That requires two independent failures in one session, and `proposal_number` continuity already rules out journal-side gaps, so this is acceptably improbable — but the docstring could name it as the known residual assumption in one clause.

### Verdict
Approve — merging. With #142/#144 restoring the Hooks and this restoring correlation, fresh sessions should now produce the `FORKABLE_EXACT` pre-mutation prefixes #42's noise measurement has been waiting for.

PR #146, opened 2026-08-13
docs(experiment): report early-prefix replay noise

YoungJinJung:
Part of #42.

Summary:
- report the first identical-arm experiment at an exact prefix before repository inspection
- publish coverage, environment parity, mechanical outcome, timing, tool, and token measurements
- update current research/status totals across both fresh neutral runs

Result:
- 3/3 judgeable pairs, 6/6 passing arms
- 0/3 mechanical disagreements and environment mismatches
- 0/6 infrastructure failures
- source pre-mutation coverage: 2/2 exact

Verification:
- jq report validation
- ruff format --check .
- ruff check .
- mypy src tests
- pytest: 443 passed


YoungJinJung:
## Review

Docs-only #42 report (+134/−2): the second fresh identical-arm run, this time at the **earliest pre-mutation prefix** (step 2, before any repository inspection) — the exact capability the #141→#142→#144→#145 chain was built to restore. Verified on the branch: suite passes, derived JSON internally consistent (6 continuations, 3 pairs, counterbalance `b, a, b`), and the coverage numbers (4/5 exact, 2/2 pre-mutation) match #145's measured claim exactly — good cross-PR consistency.

### Strengths
- **The chain closes with evidence**: 0/2 pre-mutation forkable in the first run → 2/2 here, with the report explicitly attributing the change to baseline capture (#139/#142) and Code Mode correlation (#145). Fix, then measure the fix — in that order, with both measurements committed.
- **The interesting outlier is surfaced, not smoothed**: pair 2's `neutral_a` took 6 tool calls and 154k tokens against 4 calls and ~108k everywhere else, and the report draws the right conclusion — "identical mechanical outcomes do not imply identical continuations." That single continuation is the best evidence yet that behavioral variance exists under outcome agreement, which is precisely what a noise instrument must not hide.
- Cumulative framing stays honest: 0/6 disagreements / 0/12 infra across both fresh runs, still declared "not a representative global noise bound," with the next dimensions named (failure families, labeled intervention points, explicit drift cases).

### Standing note
Third run, third 6/6-pass ceiling (#136, #141, here) — the named next step "prefixes near labeled intervention opportunities" is the right one, since disagreement can only register where continuations sometimes fail. Flagging once more only so the next run's prefix selection treats it as a requirement rather than a preference.

### Verdict
Approve — merging.

PR #147, opened 2026-08-13
docs(experiment): report routing-decision noise

YoungJinJung:
Part of #42.

Summary:
- report five identical-arm pairs at a scored routing implementation decision
- publish exact coverage, environment parity, outcome, timing, tool, and token measurements
- record that this third failure family still produced a 10/10-pass ceiling and direct the next sample toward observed failures

Result:
- 5/5 judgeable pairs, 10/10 passing arms
- 0/5 mechanical disagreements and environment mismatches
- 0/10 infrastructure failures
- cumulative fresh result: 0/11 disagreements across three tasks

Verification:
- jq report validation
- ruff format --check .
- ruff check .
- mypy src tests
- pytest: 443 passed


YoungJinJung:
## Review

Third fresh identical-arm run, this time at a scored implementation decision (step 8, post-inspection / pre-patch). Verified locally on the branch: 443 passed, mypy clean, ruff clean; CI green. Docs/JSON only — no code changes.

**What holds up:**

- **Internal consistency checks out.** 5 pairs / 10 arms in the JSON, pair indices complete (0–4 × 2 arms), and the cumulative claim is arithmetically right: 3+3+5 = 11 pairs for disagreements/mismatches, 6+6+10 = 22 arms for infrastructure failures. `research.md` and `status.md` both carry the same numbers.
- **The ceiling caveat is now the headline, not a footnote.** "Further ceiling-task repetition is low-value; the next sample needs observed mechanical failures near an intervention opportunity" is exactly the right conclusion — a third 10/10-pass family adds coverage breadth but nothing to the noise bound where #34/#23 need it. The status.md next-step column was updated to match ("Find non-ceiling tasks with observed failures"), which keeps the single-source-of-truth discipline.
- **Provenance is complete as usual:** prefix ID, source snapshot, raw JSONL SHA-256, pinned model, per-continuation session IDs.

**Notes (non-blocking):**

- Three tasks, three ceilings. The instrument has still never been pointed at a prefix where any continuation fails, so "0/11 disagreements" measures agreement among successes only — the docs say this plainly, but it's worth keeping loud until a failure-family sample lands.
- The hidden-negative-case design (double trailing slash must be rejected) was a good attempt to break the ceiling; it just didn't. The next escalation probably needs a task where the source session itself struggled, not a harder assertion on a task the model solves cleanly.

No blockers. Merging.

PR #148, opened 2026-08-13
feat(experiment): pin reasoning effort

YoungJinJung:
Part of #42.

Summary:
- add an experiment-only `--reasoning-effort` option
- forward the pinned effort to Codex as `model_reasoning_effort`
- persist the selected effort beside model/runtime provenance
- document reproducible lower-effort neutral calibration runs

Verification:
- ruff format --check .
- ruff check .
- mypy src tests
- pytest: 443 passed

### DIFF
--- a/README.md
+++ b/README.md
@@ -358,7 +358,8 @@ spotter tasks run path/to/task-set.toml --guidance "..." --run
 # --resume ~/.spotter/experiments/task-batches/<batch>.jsonl
 spotter experiment --session <id> --step <n> --guidance "..." --check "..."
 # repeat the same neutral continuation to measure replay/model outcome noise:
-spotter experiment --session <id> --step <n> --neutral --pairs 3 --check "..." --run
+spotter experiment --session <id> --step <n> --neutral --pairs 3 --check "..." \
+  --model gpt-5.6 --reasoning-effort low --run
 ```
 
 When snapshotting is enabled, `SessionStart` pins one baseline Git snapshot so read-only exploration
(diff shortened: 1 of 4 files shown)


YoungJinJung:
## Review

Small, well-scoped knob: an experiment-only `--reasoning-effort` option forwarded to Codex as a `--config model_reasoning_effort="…"` override, persisted in result-meta provenance next to the model pin. Verified locally on the branch: 443 passed, mypy clean, ruff clean; CI green.

**What holds up:**

- **Provenance symmetry with `--model`.** `reasoning_effort or "codex-config-default"` mirrors the existing model sentinel exactly, so a result file always says what effort produced it — no silent "whatever the config had that day" ambiguity. That's the property that matters for comparing noise runs across time.
- **Both arms get the same value**, so neutral-pair parity is preserved by construction.
- **Test coverage hits the three layers that matter:** CLI forwarding, meta persistence, and the exact argv shape handed to `codex exec` (including the quoted TOML value).

**Notes (non-blocking):**

- No validation of the effort value — a typo like `--reasoning-effort lwo` reaches Codex and fails there. That's acceptable fail-visible behavior for an experiment-only flag, and hardcoding Codex's accepted set (`minimal/low/medium/high`) would just rot; fine as is.
- One methodological caveat worth remembering when using this for cheap calibration runs: a noise floor measured at `low` effort bounds noise *at low effort*, not at the effort the real supervised sessions run with. Effort changes the continuation distribution, so cross-effort extrapolation of the disagreement rate should be flagged in any report that mixes them — same clock-domain-style discipline as elsewhere in the docs.
- `result_schema_version` stays at 2 with a new additive meta key — reasonable, since readers that ignore unknown keys keep working.

No blockers. Merging.

PR #155, opened 2026-08-13
Trigger Homebrew updates after releases

Bogyie:
## Why

Publishing a verified Spotter release should immediately request the matching Homebrew Formula update instead of relying on an unrelated polling schedule.

## What changed

- add a post-publication release job that sends the exact tag to `spotter-agent/homebrew-spotter` as a `spotter-release-published` repository dispatch;
- keep the dispatch in a separate job so a failed cross-repository request can be retried without republishing release assets;
- fail visibly when `HOMEBREW_TAP_TOKEN` is not configured;
- document the cross-repository token and release-to-tap contract.

## Required setup

Create the `HOMEBREW_TAP_TOKEN` repository secret before the next release. Use a fine-grained token scoped only to `spotter-agent/homebrew-spotter` with repository **Contents: Read and write** permission.

Merge spotter-agent/homebrew-spotter#4 first so the tap recognizes the dispatch event.

## Validation

- `ruff format --check .`
- `ruff check .`
- `mypy src tests`
- `pytest` — 492 passed
- release workflow YAML parsed successfully
- dispatch JSON payload rendered successfully

## Out of scope

The tap continues to own Formula validation, commits, and pull requests; the Spotter release workflow only requests an update for the published tag.


youngjinjung-linq:
## Review

Clean, well-shaped change: a separate `update-homebrew` job after `publish` that sends the exact published tag to the tap as a `spotter-release-published` repository dispatch. CI is green (checks + CodeQL/Analyze). The design choices are right:

- **Separate job, `needs: publish`** — a failed cross-repo dispatch is retriable from the Actions UI without republishing assets. That preserves the release workflow's existing no-partial-release guarantee.
- **Fail-visible on missing `HOMEBREW_TAP_TOKEN`** — an unset secret produces a loud failed job instead of a silently skipped tap update, matching the project's "degraded rather than silently healthy" discipline.
- **Exact-tag payload, tap keeps ownership** — the dispatch carries only `release_tag`; validation, Formula commits, and PRs stay in the tap. Push-vs-poll with the same trust boundary as before.
- **`jq -n --arg` for payload construction** — no string interpolation into JSON; safe even though `GITHUB_REF_NAME` is workflow-controlled here.

**Why I'm not merging this yet (ordering, not content):**

1. **spotter-agent/homebrew-spotter#4 is still open.** Until the tap listens for `spotter-release-published`, dispatches are accepted by the API but ignored — and this PR's `docs/releasing.md` edit removes the "hourly" polling description, so the docs would describe dispatch-driven behavior that doesn't exist yet. Merge tap#4 first, as the PR body itself says.
2. **`HOMEBREW_TAP_TOKEN` must be created by a repo admin** (fine-grained, `spotter-agent/homebrew-spotter` only, Contents: read/write) before the next release, or that release's `update-homebrew` job will fail. Not destructive — assets are already published and the job is retriable — but it should be in place before this lands to avoid a red release run.

Once tap#4 is merged and the secret exists, this is merge-ready as-is — no code changes requested.

One optional hardening idea for a follow-up, not this PR: `gh api` exits non-zero on transient 5xx; a single retry (`for i in 1 2; do … && break; sleep 10; done`) would cover the common flake without hiding real failures, since the job status still reflects the final attempt.

PR #156, opened 2026-08-13
docs(experiment): report recovery-prefix noise

YoungJinJung:
Part of #42.

Summary:
- report three identical-arm pairs from a source session that made and recovered from a mechanically failing patch
- pin and persist `gpt-5.6-sol` at `low` reasoning effort for source and continuations
- publish exact coverage, outcome, timing, tool, and token measurements
- keep the low-effort 0/3 result separate from the prior default-effort 0/11 cohort

Result:
- 3/3 judgeable pairs, 6/6 passing arms
- 0/3 mechanical disagreements and environment mismatches
- 0/6 infrastructure failures

Verification:
- jq report validation and raw-result SHA-256 verification
- ruff format --check .
- ruff check .
- mypy src tests
- pytest: 443 passed

### DIFF
--- a/docs/experiments/fork-neutral-recovery-prefix.json
+++ b/docs/experiments/fork-neutral-recovery-prefix.json
@@ -1,0 +1,83 @@
+{
+  "schema_version": 1,
+  "measured_at": "2026-08-14",
+  "experiment_id": "4a0ae646-68f7-489d-a5d0-d998b405bce2",
+  "source_session": "019ffc08-9b3d-7f83-bfeb-6cf3f833671a",
+  "branch_step": 14,
+  "prefix_id": "20a086ad71aff70f39a174f5600baeaff2982c6b0d24a7b33e0ae2ab73bc3928",
+  "source_snapshot": "96c6509e1265225210fc16ee1c823663c1e2f848",
+  "model": "gpt-5.6-sol",
+  "reasoning_effort": "low",
+  "scorer": "python3 check.py",
+  "raw_result_sha256": "3a863f28a6cc1460d456e2d6b62034bdf3915bf528cfbd8afe75bc955ab1b2dd",
+  "source_failure": {
+    "failure_step": 20,
+    "exception": "TypeError",
+    "recovered": true
+  },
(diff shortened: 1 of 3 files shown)


youngjinjung-linq:
## Review

Fourth fresh identical-arm run, and the first that actually breaks the source-success ceiling: the source session (pinned `gpt-5.6-sol` at `low` effort) shipped an incomplete patch, watched the scorer fail with a `TypeError`, and recovered — and the branch point (step 14) sits after the bad patch but before the failing scorer run. That is exactly the "task where the source session itself struggled" that the #147 review asked for. Verified locally on the branch: full suite green, mypy clean, ruff clean; CI green.

**What holds up:**

- **Stratification discipline.** The 0/3 low-effort result is kept out of the default-effort 0/11 cohort, with the pooled 0/14 explicitly labeled descriptive-only — precisely the cross-effort caveat from the #148 review, applied before anyone could misread a pooled number. The "defensible strata are still 0/11 and 0/3" sentence is the right shape.
- **Provenance is the strongest yet:** experiment ID scoping excludes incomplete/prepare-only attempts in the same raw JSONL, effort is persisted next to the model pin (the #148 knob doing its job), and the SHA-256/raw-path chain is intact.
- **Arithmetic checks out:** 3 pairs / 6 arms, pair indices complete; 11+3=14 pairs and 22+6=28 arms match the descriptive totals.
- **Honest interpretation.** All six continuations found and repaired the same `None` failure, so the report claims recovery-path *stability*, not a noise bound — and keeps #42 open for failures that survive to final outcomes. The elapsed-time spread (23s to 86min) with identical outcomes is a nice concrete illustration of why outcome agreement ≠ trajectory agreement.

**Non-blocking note:**

- `docs/status.md`'s Fork/replay row still says "three fresh exact prefixes produced 0/11" — true for the default-effort stratum, but the row's next-step ("find non-ceiling tasks with observed failures") is now partially satisfied by this run. Worth a one-line refresh in the next status touch; not worth a re-spin of this PR.

No blockers. Merging.

PR #157, opened 2026-08-14
fix(experiment): recheck arm environment

YoungJinJung:
Part of #42.

Summary:
- fingerprint each fork worktree again immediately before its model continuation
- compare late drift against the immutable fork-manifest environment
- refuse the arm and persist the classified environment mismatch in result rows and summaries
- refresh the status row with the recovery-prefix cohort and remaining qualification work

Verification:
- ruff format --check .
- ruff check .
- mypy src tests
- pytest: 494 passed


YoungJinJung:
## Review

Closes a real instrument gap that #156 made concrete: environment parity was checked once at pair-preparation time, but that run's continuations ran up to ~86 minutes each — plenty of window for a worktree to drift between preparation and the actual model run. Now each arm re-fingerprints immediately before `codex exec` and refuses to run on drift. Verified locally on the branch: full suite green (494), mypy clean, ruff clean; CI green.

**What holds up:**

- **Right check at the right boundary.** The recheck happens inside `_execute_arm`, after the pair-level preflight and immediately before `_run_arm` — the last moment Spotter controls. The unit test pins exactly the property that matters: on drift, `_run_arm` is never invoked.
- **Classified, not just refused.** A cheap SHA comparison short-circuits the happy path; only on mismatch does it load the immutable manifest and produce a named drift detail via `compare_environments`, with distinct sentinels for a missing manifest environment and for preflight errors (`ENVIRONMENT_PREFLIGHT_ERROR:` wraps OSError/ReplayError/ValueError rather than crashing the batch). Fail-visible in the established taxonomy.
- **Persistence keeps the summary honest.** Overwriting the row's `environment_preflight` with the arm-level diagnostic means `summarize` buckets late drift under environment mismatches — the same stratum the reports already publish — instead of burying it in generic infra failures. The integration test asserts the full path down to the `environment mismatches=1/1` summary line.
- **Status row refresh** picks up the #156 review note: default-effort 0/11 and low-effort 0/3 now listed as separate strata, next step reworded to final-outcome failures + explicit drift qualification.

**Nitpick (non-blocking):** the `ENVIRONMENT_FINGERPRINT_MISMATCH` branch for a manifest-less plan is unreachable from `_execute_arm` (which gates on `plan.manifest`) — harmless defensive coding, fine to leave.

No blockers. Merging.

PR #158, opened 2026-08-14
feat(replay): capture agent config provenance

YoungJinJung:
Part of #42.

Summary:
- derive replay-prefix agent configuration from the latest pre-branch Codex `turn_context`
- persist effort, approval policy, personality, sandbox mode, and collaboration mode as canonical JSON
- exclude nested instructions, writable roots, and other potentially sensitive configuration values
- retain the explicit `not_captured` sentinel for legacy rollouts without turn context

Verification:
- ruff format --check .
- ruff check .
- mypy src tests
- pytest: 494 passed

### DIFF
--- a/README.md
+++ b/README.md
@@ -370,9 +370,10 @@ snapshots. `spotter fork` writes a durable manifest under `~/.spotter/fork-manif
 source event, snapshot, rollout-prefix digest, external-effect warnings, and captured environment
 fingerprint. Counterfactual pairs are fully prepared and must pass shared-prefix and captured-
 environment parity checks before either agent arm runs, then rechecks each arm against its immutable
-manifest immediately before model continuation. Ignored files, environment variables, and
-agent configuration are explicitly marked as not captured rather than assumed equal. Experiments
-can pin both the Codex model and reasoning effort, and persist both values in result provenance.
+manifest immediately before model continuation. Ignored files and environment variables are
+explicitly marked as not captured rather than assumed equal. Prefixes preserve the rollout model,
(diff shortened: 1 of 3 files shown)


YoungJinJung:
## Review

Upgrades the fork manifest's `agent_config` from a permanent `not_captured` sentinel to a secret-free subset of the source session's actual turn configuration, derived from the latest `turn_context` record before the branch cut. Verified locally on the branch: 494 passed, mypy clean, ruff clean; CI green.

**What holds up:**

- **Allowlist, not blocklist.** Only five known-safe fields are extracted (`approval_policy`, `effort`, `personality`, `collaboration_mode.mode`, `sandbox_policy.type`); everything else — including `writable_roots` and `developer_instructions` — never enters the manifest. The end-to-end test plants decoy `"secret"` values in exactly those nested fields and asserts they don't appear in the serialized config. That's the right hygiene model: a new Codex field is excluded by default rather than leaked by default.
- **Latest-pre-branch semantics are correct.** The scan breaks at the branch call, and later `turn_context` records overwrite earlier ones, so the config that actually governed the branched turn wins — matters when settings change mid-session.
- **The sentinel survives where it should.** Legacy rollouts without any `turn_context` still get `not_captured` — unknown stays unknown instead of degrading to an empty object that reads as "captured and empty".
- **This closes the loop on the #148 review caveat.** With source-side `effort` now in prefix provenance and continuation-side effort in result meta, an effort mismatch between source and continuation is detectable from the artifacts alone rather than from operator memory. The #156-style stratification can now be enforced by inspection.
- Canonical JSON keeps the value stable for comparison/digesting.

No blockers. Merging.

PR #159, opened 2026-08-14
fix(experiment): enforce source config parity

YoungJinJung:
Part of #42.

Summary:
- compare explicit continuation model and reasoning-effort pins with source-prefix provenance
- refuse both arms before model execution when source and continuation config differ
- fail visibly when a requested pin cannot be proven from legacy source provenance
- persist the mismatch as a setup diagnostic without misclassifying it as environment drift

Verification:
- ruff format --check .
- ruff check .
- mypy src tests
- pytest: 495 passed

### DIFF
--- a/README.md
+++ b/README.md
@@ -373,7 +373,8 @@ environment parity checks before either agent arm runs, then rechecks each arm a
 manifest immediately before model continuation. Ignored files and environment variables are
 explicitly marked as not captured rather than assumed equal. Prefixes preserve the rollout model,
 runtime, and a secret-free subset of turn configuration; experiments can pin both the Codex model
-and reasoning effort and persist both values in result provenance.
+and reasoning effort and persist both values in result provenance. When those values are pinned and
+source provenance is available, a mismatch is rejected before either arm starts.
 Neutral mode gives both arms the exact control prompt and reports mechanically judgeable outcome
 disagreement, environment-preflight mismatch, and infrastructure-failure rates separately. The
 command provides the measurement path; a representative executed corpus is still required before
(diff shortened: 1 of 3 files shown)


YoungJinJung:
## Review

Completes the three-PR chain: #148 added the model/effort pins, #158 captured the source's actual config in prefix provenance, and this PR makes the two agree or refuse — a pinned continuation that doesn't match proven source config never reaches `codex exec`. Verified locally on the branch: 495 passed, mypy clean, ruff clean; CI green.

**What holds up:**

- **The invariant is the right one for #42.** Neutral replay measures noise only if source and continuation run the same configuration; the #156/#148 reviews' cross-effort caveat is now enforced by the instrument instead of by operator discipline. The stratification that the reports do by hand (0/11 default vs 0/3 low) can no longer be accidentally polluted.
- **Correct failure taxonomy.** Mismatches land as `SETUP_FAIL` with named diagnostics (`SOURCE_MODEL_MISMATCH`, `SOURCE_REASONING_EFFORT_MISMATCH`, `…_UNAVAILABLE`, `SOURCE_CONFIG_ERROR:`) rather than being folded into environment drift or infra failures — each stratum in the summaries keeps its meaning.
- **Unknown provenance refuses rather than assumes.** Pinning against a legacy source whose manifest lacks config yields `…_UNAVAILABLE` instead of silently running an unprovable comparison — consistent with the `not_captured` discipline in #158.
- **Zero behavior change for unpinned runs** (`model is None and reasoning_effort is None` → MATCHED), and the check runs once per pair before either arm, so both arms fail together instead of producing a half-executed pair.
- Test pins the property that matters: on mismatch, `_run_arm` is never invoked and both rows persist the diagnostic.

**Note (non-blocking):** this deliberately forecloses running pinned continuations at a *different* config than the source (e.g. low-effort continuations of a default-effort source). For neutral noise measurement that's correct — such a run measures config delta, not replay noise. If a deliberate config-delta mode is ever wanted, it should arrive as an explicit override flag with its own result labeling, not by relaxing this check.

No blockers. Merging.

PR #160, opened 2026-08-14
fix(experiment): reject shared arm worktrees

YoungJinJung:
Part of #42.

Summary:
- require the two counterfactual arms to use distinct resolved worktree paths
- refuse both arms before model execution when fork planning aliases the same mutable workspace
- persist `SHARED_ARM_WORKTREE` in the established preflight diagnostic field

Verification:
- ruff format --check .
- ruff check .
- mypy src tests
- pytest: 496 passed

### DIFF
--- a/README.md
+++ b/README.md
@@ -369,8 +369,9 @@ before the first mutation has a fork anchor; mutation boundaries continue to add
 snapshots. `spotter fork` writes a durable manifest under `~/.spotter/fork-manifests/`. The manifest links the
 source event, snapshot, rollout-prefix digest, external-effect warnings, and captured environment
 fingerprint. Counterfactual pairs are fully prepared and must pass shared-prefix and captured-
-environment parity checks before either agent arm runs, then rechecks each arm against its immutable
-manifest immediately before model continuation. Ignored files and environment variables are
+environment parity checks before either agent arm runs. The arms must use distinct worktrees, and
+each is rechecked against its immutable manifest immediately before model continuation. Ignored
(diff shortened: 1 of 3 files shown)


YoungJinJung:
## Review

Small diff, real bug class. If fork planning ever aliases both arms onto the same mutable workspace, the existing pair preflight would actually *pass* — a shared worktree trivially produces identical environment fingerprints, which is exactly what the parity check wants to see. The failure mode is silent: two "independent" continuations racing in one directory, scored as a clean pair. This turns that into a loud `SHARED_ARM_WORKTREE` refusal before either arm runs. Verified locally on the branch: 496 passed, mypy clean, ruff clean; CI green.

**What holds up:**

- **Right location and ordering** — the check sits in `_pair_environment_preflight` after the provenance-availability gate and before the prefix/fingerprint comparisons, so the degenerate case is named specifically instead of masquerading as MATCHED.
- **`.resolve()` on both sides** catches symlink aliasing, not just string-equal paths.
- **Reuses the established taxonomy** — `INFRA_FAIL` with the diagnostic in the same preflight field the summaries already stratify; no new plumbing.
- Test pins the property that matters: both arms refused, `_run_arm` never invoked, diagnostic persisted on both rows.

This rounds out a solid instrument-hardening sequence (#157 late drift, #159 config parity, now workspace aliasing) — the neutral-replay preconditions are increasingly enforced by the tool rather than assumed by the operator.

No blockers. Merging.

PR #161, opened 2026-08-14
docs(experiment): publish fork fidelity v1

YoungJinJung:
Part of #42.

Summary:
- consolidate the historical coverage baseline and four fresh identical-arm runs into versioned fidelity artifact v1
- preserve unpinned-effort 0/11 and pinned-low 0/3 as separate strata
- publish raw coverage/failure counts and clearly label naive pair-level upper bounds as non-inferential across repeated prefixes
- record an explicit NO-GO for representative causal use by #34/#23 and the evidence needed to qualify a later version

Verification:
- jq report validation and aggregate arithmetic assertions
- ruff format --check .
- ruff check .
- mypy src tests
- pytest: 496 passed


YoungJinJung:
## Review

The consolidation artifact the last two weeks of #42 work has been building toward: a versioned fidelity qualification that downstream experiments cite instead of assuming replay validity. Verified locally on the branch: 496 passed, mypy clean, ruff clean; CI green.

**Independently verified the arithmetic:**

- Fresh coverage totals match the four underlying reports exactly: candidates 5+5+6+9=25, exact 1+4+5+5=15, pre-mutation 2+2+3+3=10 with 0+2+3+3=8 forkable.
- Strata counts match (11 pairs/22 arms; 3 pairs/6 arms), and the one-sided 95% zero-event upper bounds are computed correctly: 1−0.05^(1/11)=0.2384 and 1−0.05^(1/3)=0.6316.
- Those bounds are labeled "orientation only, not valid cross-task" with the right reason (pairs repeat prefixes, so they aren't independent Bernoulli draws) — publishing the naive number *with* its invalidity condition is much better than either hiding it or citing it.

**What makes this a good qualification document:**

- **NO-GO is the correct call and correctly scoped.** 28/28 passing arms means the failure region is uncalibrated — zero observed disagreement is not a small noise bound, and the report says so plainly. Development use stays allowed; representative causal citation by #34/#23 is blocked until final-outcome failures, more task families, and drift cases exist. That's the "separate implementation from evidence" rule applied to the instrument itself.
- **"Default-effort" honestly renamed to "unpinned/legacy".** Effort wasn't persisted for the first three runs, so "default" was an assumption; the rename plus the blocking-gap entry makes the provenance hole explicit rather than papering over it.
- **The instrument-controls checklist maps one-to-one onto the merged hardening sequence** (#157 late recheck, #159 config parity, #160 distinct worktrees, plus the pre-existing preflight/exclusion/baseline machinery) — a reader can audit each claimed control against a real enforcement point in the code.
- Versioning (`fork-fidelity-v1`, `qualification_version`) gives later runs a clean supersession path instead of edit-in-place drift.

No blockers. Merging.

PR #162, opened 2026-08-14
feat(metrics): persist App Server arrival order

YoungJinJung:
## Why
App Server events need durable receipt ordering when source timestamps are equal or unavailable; otherwise later latency and conformance analysis can only infer order from file position.

Related: #33

## What changed
- persist a connection-epoch-scoped `arrival_seq` on normalized App Server Trace IR
- recover the last sequence from journals so same-epoch ingestion resumes monotonically
- report arrival-order coverage in `spotter metrics` and document the timestamp contract

## Validation
- `ruff format --check .`
- `ruff check .`
- `mypy src tests`
- `pytest` (497 passed)

## Notes
Legacy events remain explicitly uncovered (`arrival_order=0/N`) rather than being backfilled with inferred ordering.


YoungJinJung:
## Review

Adds a durable, connection-epoch-scoped `arrival_seq` to normalized App Server Trace IR, so events with equal or missing source timestamps keep a provable receipt order across journal reloads instead of relying on file position. Verified locally on the branch: 497 passed, mypy clean, ruff clean; CI green.

**What holds up:**

- **Epoch-scoped is the right scope.** Sequences are per connection epoch, matching the existing rule that nothing (turn durations, operation correlation, now ordering) crosses epoch boundaries. An ordering claim that spanned reconnects would be manufactured, and this design can't make one.
- **Recovery is monotonic, and regression is loud.** `_remember` replays the per-epoch max from journals, so a restarted ingestor continues at seq 3 (pinned by the restart test); an explicitly supplied sequence that fails to advance raises `IngestionError` rather than silently reordering. The `event.arrival_seq or previous_seq + 1` assignment can't be fooled by a supplied 0 either — 0 fails the `<= previous` check first, and deserialization rejects non-positive values back to None.
- **Legacy stays uncovered, not backfilled.** `arrival_order=0/N` for pre-existing events keeps the unknown-not-zero discipline; inferring historical order from file position would have been exactly the kind of manufactured provenance the metrics doc warns against. The coverage denominator (epoch-bearing events only) makes Hook-era events ineligible rather than counted as failures.
- **Serialization is symmetric and strict** — `arrival_seq` joins the trace metadata allowlist, and the reader validates int/not-bool/positive the same way `connection_epoch` does.
- Docs updated in the same PR (timestamp contract in architecture, #33 roadmap row, status row) per the contract-change rule.

No blockers. Merging.

PR #163, opened 2026-08-14
feat(metrics): report token field coverage

YoungJinJung:
## Why
Main token cost was represented only by a cumulative total, hiding input/output/cache/reasoning detail and making partial provider coverage look complete.

Related: #33

## What changed
- project the latest cumulative token breakdown per session without adding repeated cumulative updates
- report input, cached-input, cache-write-input, output, reasoning-output, and total coverage independently
- preserve unavailable fields as unknown and document the accounting contract

## Validation
- `ruff format --check .`
- `ruff check .`
- `mypy src tests`
- `pytest` (498 passed)

## Notes
Provider `total` values remain labeled `cumulative/unknown-scope`; this PR does not infer call- or turn-level scope from field names.


youngjinjung-linq:
## Review

Replaces the single cumulative Main-token scalar with a field-level breakdown (`input`, `cached_input`, `cache_write_input`, `output`, `reasoning_output`, `total`), each carrying its own per-session coverage denominator. Verified locally on the branch: 498 passed, mypy clean, ruff clean; CI green. Also confirmed the ingestion normalizer's token-field allowlist already existed on main — the ingestion test change exercises existing behavior rather than smuggling in a normalizer change.

**What holds up:**

- **The stated problem is real and the fix is shaped correctly.** A lone total made partial provider coverage look complete; now `cache_write_input=unknown (0/2 sessions)` says precisely which fields which fraction of sessions actually reported. Unknown-not-zero, with the denominator printed.
- **Latest-snapshot-per-session is the coherent aggregation.** Taking the session's latest cumulative `total` mapping wholesale (rather than merging fields across different updates) avoids both double-counting repeated cumulative updates and stitching an incoherent breakdown out of values observed at different times. The dedup test pins it: totals 5→14 in one session contribute 14 once, not 19.
- **Scope humility is preserved.** Values stay labeled `cumulative/unknown-scope`; the PR explicitly refuses to infer call- or turn-level scope from field names, which is the trap this kind of breakdown usually falls into.
- **Backward-compatible surface** — `cumulative_main_tokens` remains as the breakdown's total value, so existing consumers keep working.
- `_token_count` validation (int, not bool, non-negative) matches the established strictness pattern, and docs (architecture accounting contract, roadmap #33, status row) move in the same PR.

No blockers. Merging.

PR #164, opened 2026-08-14
feat(metrics): break down semantic actions

YoungJinJung:
## Why
Main action totals did not show which operation families consumed work or produced failures, limiting efficiency diagnosis. Counting raw Hook/App Server observations would also double-count lifecycle events.

Related: #33

## What changed
- classify already-correlated semantic actions as command, file change, or tool
- report action counts and classified failure denominators per family and observation surface
- keep Hook and App Server surfaces separate and document the projection

## Validation
- `ruff format --check .`
- `ruff check .`
- `mypy src tests`
- `pytest` (499 passed)

## Notes
Repeated-equivalent-action detection remains owned by #28; this PR does not add a second heuristic.


YoungJinJung:
## Review

Breaks Main semantic actions down by operation family (`command` / `file_change` / `tool`) per observation surface, each with its own classified-failure denominator. Verified locally on the branch: 499 passed, mypy clean, ruff clean; CI green.

**What holds up:**

- **No second heuristic.** Family classification piggybacks on the already-correlated action identity (`_action_key`) — the same key that dedupes start/result observations into one action also carries its family, so lifecycle events can't double-count into family totals. The explicit note deferring repeated-equivalent-action detection to #28 is the right scope boundary; two overlapping heuristics for "same action" would have been the classic mistake here.
- **Denominator discipline per family.** `file_change failed=1/1 classified` vs `tool` having no classified outcomes keeps the blind-spot-vs-failure separation the outcomes work established — a family with unclassifiable results shows a small denominator, not a fabricated success rate. And every key in `classified`/`failed` provably has a family, since both are populated in the same branch that records the outcome family.
- **Surfaces remain unsummed** during migration, per the architecture rule; the family line renders under each surface separately.
- Counter-based merge across a surface's journals is boring and correct; render sorts families for stable output.

**Nit (non-blocking):** an operation whose start and outcome arrive under different kinds (say `tool_started` + `command_result` on one operation id) would be counted under different families in the action vs outcome maps. No current producer does that, and the mismatch would be visible in the family denominators rather than silent — fine to leave until a real producer exhibits it.

No blockers. Merging.

PR #165, opened 2026-08-14
feat(metrics): count unique action resources

YoungJinJung:
## Why
Semantic action totals show volume but not how broadly a session touched files or external tools. Raw lifecycle observations can also duplicate the same resource.

Related: #33

## What changed
- count unique file, classified Hook, and MCP server/tool resources from explicit normalized fields
- deduplicate resource identities across correlated action lifecycle events and report action-level coverage
- render unavailable resource coverage as unknown instead of zero and document the projection boundary

## Validation
- ruff format --check .
- ruff check .
- mypy src tests
- pytest (499 passed)

## Notes
Command strings and opaque tool arguments are intentionally not parsed to invent resource identity.


YoungJinJung:
## Review

Adds unique-resource counts per observation surface (files, classified Hook resources, MCP `server/tool` identities), deduplicated across a correlated action's lifecycle events, with an action-level declaration denominator. Verified locally on the branch: 499 passed, mypy clean, ruff clean; CI green.

**What holds up:**

- **Identity comes only from explicit normalized fields.** The refusal to parse command strings or opaque tool arguments is the load-bearing decision — a `grep src/a.py` command touching a file does not get a fabricated `file:` resource, it just shows up in the declaration denominator as an action that declared nothing. That's "do not invent observed outcomes" applied to resources.
- **Namespaced keys (`file:`, `resource:`, `tool:server/tool`)** prevent a file named `workspace` from colliding with a classified resource named `workspace` — cheap and correct.
- **The two-level dedup is right:** per-action sets collapse repeated lifecycle observations (the test's duplicated `files` entry counts once), then the surface-level union collapses the same file touched by different actions. Volume and breadth stay distinguishable — `resource_actions` tells you how many actions declared anything at all.
- **Unknown-not-zero again where it matters:** zero declaring actions renders `resources=unknown (0/0 actions declared)` rather than `0 unique` — an instrument that saw no declarations doesn't claim the session touched nothing.
- Surfaces stay separate per the migration rule, and the architecture doc gains the projection-boundary paragraph in the same PR.

No blockers. Merging.

PR #166, opened 2026-08-14
feat(replay): verify declared environment resources

YoungJinJung:
## Why
Two fork arms can lose the same ignored fixture file and still match each other, hiding source-to-fork environment drift. #42 requires explicit resource coverage before replay fidelity can be qualified.

Related: #42

## What changed
- add repeatable --environment-resource declarations for fork and experiment commands
- hash only declared relative non-secret files and compare source state with each restored fork
- classify missing ignored/untracked files and block both arms before any agent run
- persist the source preflight in fork manifest schema v2 while keeping v1 readable

## Validation
- ruff format --check .
- ruff check .
- mypy src tests
- pytest (503 passed)

## Notes
No paid agent arms were executed. Undeclared ignored files and environment variables remain explicit coverage gaps.


youngjinjung-linq:
## Review

This targets exactly the right gap — fidelity v1's "ignored files remain uncaptured" blocker, and specifically the failure mode pair parity cannot see: both arms losing the same ignored fixture and matching *each other* perfectly. The design is sound: opt-in declaration, hash-only capture, path-traversal/symlink rejection, ignored/untracked/tracked classification, schema v2 with v1 kept readable, and the source→fork preflight persisted in both plan and manifest. The pair-level blocking test and the manifest migration test are good.

**Blocker — the immediate pre-run arm recheck falsely refuses every declared-resource run:**

In `_arm_environment_preflight`, when the manifest declares resources:

1. The first `current = fingerprint_environment(worktree)` is computed **without** declared resources, so its `fingerprint_sha256` can never equal `plan.environment_fingerprint` (whose material includes the non-empty `declared_resources` list). The early `return "MATCHED"` is unreachable for any declared-resource plan.
2. Control falls through, recomputes `current` with resources, and calls `compare_environments`. If nothing drifted, `comparison.drift` is `()` — but the function then hits `",".join(()) or "UNKNOWN_ENVIRONMENT_DRIFT"` and unconditionally returns `ENVIRONMENT_MISMATCH:UNKNOWN_ENVIRONMENT_DRIFT`.

Net effect: any `--environment-resource … --run` experiment INFRA_FAILs both arms at the #157 recheck even when the environment is pristine. Fail-closed rather than fail-open, so nothing silent — but the feature's happy path cannot execute. Fix is small: either fetch the declared paths and compute the resource-inclusive fingerprint **before** the sha equality check, or add `if comparison.equivalent: return "MATCHED"` after the comparison. Either way, please add the missing test: arm recheck with declared resources and an unchanged worktree must return `MATCHED` and reach `_run_arm`.

**Non-blocking notes:**

- Declaring a file that is missing in the *source* yields `UNKNOWN_ENVIRONMENT_DRIFT` at fork time (missing→missing falls to the else branch). Arguably that should be a loud declaration error instead — the operator typo'd a path — but drift-visible is acceptable.
- `_fingerprint_resources` reads file bytes to hash them; the manifest stores only path/state/sha256, which matches the "non-secret" contract. Worth keeping that invariant in mind if anyone later adds content echoing to diagnostics.

Holding the merge for the blocker; everything else is merge-ready as-is.

YoungJinJung:
## Re-review after 72f8e5bd

The blocker is fixed, and belt-and-suspenders at that: the resource-inclusive fingerprint is now computed before the sha equality check, *and* `comparison.equivalent` returns `MATCHED` as a second guard — so a pristine declared-resource worktree can no longer fall through to the unconditional mismatch return. The new `test_unchanged_declared_resource_recheck_reaches_agent_run` pins exactly the property I asked for: declared resources + unchanged worktree → `MATCHED` → `_run_arm` invoked.

Re-verified locally on the branch: full suite green (504), mypy clean, ruff clean; CI green; MERGEABLE.

With the recheck fixed, the full chain now holds end to end: declaration → source fingerprint → restore → source→fork drift classification (pair-level block, both arms) → immediate pre-run recheck including declared files → schema-v2 persistence with v1 readable. This closes the last operationally addressable blocking gap from fidelity v1's list short of actually executing drift cases, which the status row correctly keeps as the #42 next step.

Merging.

PR #167, opened 2026-08-14
docs(experiment): record declared resource drift

YoungJinJung:
## Question
Does the declared-resource guard catch the #42 failure mode where both fork arms lose the same ignored file and still match each other?

Related: #42

## What changed
- run one induced ignored-file source-to-fork drift case through the real experiment CLI
- record the scoped machine-readable result and raw artifact hash
- update current status while keeping fidelity v1 NO-GO

## Result
Both neutral arms were classified INFRA_FAIL with SOURCE_ENVIRONMENT_MISMATCH:MISSING_IGNORED_FILE before agent execution. Zero Codex continuations or paid model calls started.

## Validation
- jq empty docs/experiments/fork-declared-resource-drift.json
- ruff format --check .
- ruff check .
- mypy src tests
- pytest (504 passed)

## Limitations
The drift was deliberately induced after the historical source run. Final-outcome failures, natural environment drift, undeclared resources, and environment variables remain unqualified.


YoungJinJung:
## Review

Operational qualification of the #166 declared-resource guard: an induced ignored-file drift (`.fixture-config` + exact `.gitignore` rule) run through the real CLI with `--run`, blocked at pair preflight with `SOURCE_ENVIRONMENT_MISMATCH:MISSING_IGNORED_FILE` on both arms, zero agent starts, zero paid calls. Verified locally on the branch: 504 passed, mypy clean, ruff clean; JSON internal consistency checked with jq (2 arms, both `agent_started=false`, 2 infra failures, 0 judgeable pairs); CI green.

**What holds up:**

- **This is the end-to-end proof the #166 review chain needed.** The unit tests proved the logic; this run proves the wiring — real fork manifests (schema v2), real preflight, real result persistence, real Git-aware cleanup. And notably, the diagnostic is the *named* `MISSING_IGNORED_FILE`, not a fallback `UNKNOWN_ENVIRONMENT_DRIFT` — empirical confirmation that the arm-recheck fix from 72f8e5bd behaves correctly in the execution path, not just under monkeypatched tests.
- **Claims stay inside the evidence.** "Proves the operational guard, not replay fidelity or a model noise rate" is exactly right — the drift was induced, so this cannot count toward the natural-drift qualification, and the doc says so in both the prose and the machine-readable `limitations`/`induced_fixture: true` fields.
- **Source hygiene is documented:** temporary fixture and `.gitignore` rule removed after the run, prior tracked/untracked state preserved, raw journal hash recorded post-append.
- **status.md moves precisely as far as the evidence allows** — the Fork/replay row gains the proof link while fidelity v1 stays NO-GO, and the next step narrows to final-outcome failures + *natural* drift.

Cost discipline note worth appreciating: the experiment design guaranteed `--run` could not reach a model call if the guard worked, so the qualification was free either way — the guard failing would have been caught by the check exit, not by a wasted continuation.

No blockers. Merging.

PR #169, opened 2026-08-14
feat(metrics): join objective outcomes with costs

YoungJinJung:
## Why

#33 needs mechanically scored experiment outcomes reported beside their costs without labeling ordinary user sessions as pass or fail.

## What changed

- project versioned counterfactual and task-batch result journals by stable run/pair/arm identity
- report judgeable arms, guidance and neutral pair outcomes, agent-reported token coverage, and per-arm elapsed coverage in global `spotter metrics`
- reject unsupported newer result schemas and tolerate only a torn final JSONL append
- document the separate objective-outcome accounting boundary

## Validation

- `ruff format --check .`
- `ruff check .`
- `mypy src tests`
- `pytest` — 506 passed
- exercised `spotter metrics` against retained local experiment/task-batch artifacts

Part of #33.


YoungJinJung:
## Review

Adds the objective-outcome half of #33: `spotter metrics` (global mode only) projects versioned counterfactual and task-batch result journals, joins each mechanical classification to the costs carried on the same durable run/pair/arm row, and reports guidance/neutral pair outcomes beside per-arm token and elapsed coverage. Verified locally on the branch: 506 passed, mypy clean, ruff clean; CI green.

**What holds up:**

- **The accounting boundary is the headline design decision, and it's right.** Objective outcomes stay outside user-session telemetry — the join runs only for global metrics and never labels an unscored session pass/fail. The architecture doc states the boundary in the same PR.
- **Schema handling refuses to guess.** Unknown classifications, missing schema versions, and *newer* schemas all raise `ObjectiveOutcomeError` and abort metrics visibly (exit 1) instead of silently projecting a format the reader doesn't understand. Only a torn final JSONL row is tolerated — matching the interrupted-append semantics the journals already promise — while mid-file corruption still fails loudly.
- **Identity is durable, not positional.** Pair grouping keys off `run_id:experiment_pair_id` / `experiment_id:pair` + arm, with duplicate-identity conflict detection, so re-reads and multi-file layouts can't double-count or mispair arms.
- **Coverage discipline carries through:** judgeable denominators (PASS/TASK_FAIL only) for both pair analyses, `agent_reported_tokens=… (2/4 arms)` with unknown-not-zero when absent, and elapsed computed strictly within one row's own started/ended stamps — no cross-clock subtraction.
- The stderr `tokens used` fallback is a heuristic, but a well-scoped one: it parses Codex's own summary line, is labeled *agent-reported*, takes the last match, and yields None rather than a guess when absent.

**Note (non-blocking):** a single unreadable/corrupt artifact under `~/.spotter/experiments` now aborts *all* of `spotter metrics` rather than just the objective section. That's the project's fail-visible stance and I think it's correct today; if the experiments directory ever accumulates unrelated debris, a per-artifact error listing might become friendlier than a hard abort — decide when it actually happens.

No blockers. Merging.

PR #170, opened 2026-08-14
feat(signals): detect resource failure streaks

YoungJinJung:
## Why

#28 needs cheap event-driven evidence candidates before reviewer scheduling can move away from fixed cadence. This PR establishes the incremental and durable signal boundary with one conservative family.

## What changed

- add an incremental failure-streak SignalEngine over runtime-neutral Trace IR
- require stable event IDs and explicit normalized file, resource, test, or MCP tool identity; opaque shell outcomes remain unknown
- journal active candidates, cooldown suppressions, and resolved/stale transitions with thread, turn, epoch, state version, evidence, and resource provenance
- restore cooldown state after restart and backfill a derived candidate when a crash lands between source-outcome and candidate appends
- connect the engine to the daemon-owned App Server ingestion path

Reviewer scheduling and the remaining #28 signal families intentionally remain out of scope.

## Validation

- `ruff format --check .`
- `ruff check .`
- `mypy src tests`
- `pytest` — 513 passed

Part of #28.


YoungJinJung:
## Review

First #28 signal family — an incremental failure-streak `SignalEngine` wired into the daemon's App Server ingestion path, journaling active/cooled-down/resolved/stale candidates as derived Trace IR. This moves Event-driven detection from ❌ to 🟡 and it earns it. Verified locally on the branch: 513 passed, mypy clean, ruff clean; CI green.

**What holds up:**

- **"A signal is a hypothesis, not a verdict" is enforced, not just quoted.** Candidates carry evidence IDs, resource scope, and a deterministic magnitude — no semantic judgment. Opaque shell outcomes stay unknown (the repeated `rg` probe test pins that no candidate fires), keeping the #165 no-invented-identity rule intact at the detection layer.
- **Emit-once with journaled suppression is the right anti-spam shape.** Streak threshold fires one `active` candidate; continued failures journal `cooled_down` records instead of re-arming, so downstream scheduling sees one actionable candidate while the streak's continuation remains durable evidence. Dedup by event ID prevents replay inflation.
- **Recovery is genuinely crash-safe.** Deterministic derived event IDs (`spotter:signal:{id}:{sha(status:source)}`) make hydration idempotent: replaying the journal reconstructs cooldown state without duplicating candidates, and the pending-map backfill covers the exact crash window between source-outcome append and candidate append — both proven by tests including a real two-process journal round-trip.
- **No recursion hazard:** derived candidate events re-enter `_record` but their kinds are neither outcome nor terminal, so the engine no-ops on its own output; hydration explicitly skips them.
- **Boundary discipline carries over:** epoch comparison only when both sides are known, turn changes stale the streak, evidence bounded at 20 IDs, and the periodic reviewer explicitly remains the trigger — scheduling stays out of scope as stated.

**Notes (non-blocking):**

- One streak per thread means interleaved failures on two resources (A, B, A, B…) stale each other and never fire — literally "consecutive" per the doc, correct for family v1, but worth remembering when the alternating-failure pattern shows up in real traces; a per-key streak map is the natural upgrade path.
- `_seen_event_ids` grows with thread history — same O(history) shape as the ingestion recovery ponytail note, so it should ride along with the #89 checkpoint work.
- `test_result` without explicit resources folds into a synthetic `validation` scope — a mild widening of the explicit-identity rule that seems justified for test outcomes; the doc's "explicit … test … scope" wording covers it.

No blockers. Merging.

PR #171, opened 2026-08-14
feat(signals): queue signal-driven reviews

YoungJinJung:
## Why

Issue #28 now emits durable signal candidates, but model review is still triggered only by fixed Hook cadence. This adds the scheduling boundary needed to target reviews from evidence without putting semantic work on the synchronous path.

Related: #28

## What changed

- deduplicate each active signal lifecycle into one durable reviewer job
- bind the job to the exact immutable ThreadState version and turn/epoch target
- discard jobs when the target is no longer proven active
- replay pending jobs and backfill interrupted queue/discard appends on daemon recovery
- expose the partial queue state in doctor and current-state documentation

## Validation

- `ruff format --check .`
- `ruff check .`
- `mypy src tests`
- `pytest` (517 passed)

## Notes

Reviewer model execution, bounded input construction, candidate merge/budget policy, and additional signal families remain intentionally out of scope.


YoungJinJung:
## Review

The scheduling boundary between #170's durable candidates and future reviewer execution: one deduplicated, restart-safe reviewer job per active signal lifecycle, bound to the exact immutable `ThreadState` snapshot the candidate names. Verified locally on the branch: 517 passed, mypy clean, ruff clean; CI green.

**What holds up:**

- **The snapshot binding is cross-validated, not assumed.** `_job` refuses to construct a job unless the candidate's payload identity, the event identity, and the snapshot's thread/version/turn/epoch all agree — a candidate that names a state it can't prove raises `ReviewSchedulingError` instead of queuing a review against the wrong world. Since candidates are self-produced, loud failure on mismatch is the right call.
- **Conservative liveness fencing in both directions.** A candidate whose turn was never proven active is discarded at queue time (`target_not_active`), and pending jobs are discarded the moment the target turn/epoch changes or a terminal event lands (`target_changed`) — with discards journaled as durable events, so "we chose not to review" is as auditable as "we queued a review". This is the stale-control-fencing discipline applied to review targeting.
- **One job per signal lifecycle, permanently.** `_seen_signal_ids` means a discarded job's signal can't re-queue on replayed or repeated candidates — the test pins that a post-discard duplicate candidate yields nothing. Cooldown suppressions never reach the scheduler at all since only `active` status queues.
- **Recovery mirrors the #170 pattern and goes one step further:** hydration replays the journal through a private `ThreadStateStore`, indexes snapshots by (thread, version) so each candidate re-binds to the *exact* state it originally named, and backfills only missing queue/discard transitions via deterministic event IDs — proven idempotent by the double-hydrate test. The `_append_derived` path keeps recovery from re-running producers out of order.
- **Scope honesty throughout:** no model calls, no input construction, no merge/budget policy; the doctor line and three status rows all say precisely what exists ("signal-driven jobs are durable; reviewer execution is not connected") — the periodic reviewer remains the only thing spending tokens.

The #170→#171 sequence is textbook boundary construction: detection emits hypotheses, scheduling binds them to immutable evidence, and the expensive step stays unbuilt until the cheap steps are durable. No blockers. Merging.

PR #172, opened 2026-08-14
feat(reviewer): build bounded signal inputs

YoungJinJung:
## Why

PR #171 made signal-driven reviewer jobs durable, but each job still retained only the full internal ThreadState snapshot. A future asynchronous reviewer needs a small explicit input contract rather than an unbounded transcript or ad hoc state access.

Related: #28

## What changed

- build an immutable bounded reviewer input from the exact snapshot named by a signal candidate
- include candidate features/evidence references, goal and constraints, touched and unvalidated files, recent failures, validation state, and known coverage gaps
- cap both item counts and text lengths, and record input coverage/truncation on the durable queue event
- deterministically reconstruct the same input during queue hydration
- update current-state documentation without claiming model execution

## Validation

- `ruff format --check .`
- `ruff check .`
- `mypy src tests`
- `pytest` (518 passed)

## Notes

Prompt rendering/model execution, candidate merge/budget policy, and additional signal families remain intentionally out of scope.


YoungJinJung:
## Review

Third step of the #28 chain: each queued reviewer job now carries an immutable, explicitly bounded input slice of the exact snapshot it binds — candidate features/evidence, goal, constraints, touched/unvalidated files, recent failures, validation status, and coverage gaps — instead of a raw ThreadState or an unbounded transcript. Verified locally on the branch: 518 passed, mypy clean, ruff clean; CI green.

**What holds up:**

- **Truncation is a recorded fact, not a silent one.** Every cap (600-char goal, 300-char texts, 20/50/20/10 item limits) that fires lands in `truncated_fields`, and the full `input_coverage` dict — counts shown, goal presence, truncation flags — is persisted on the durable `review_job_queued` event. A future reviewer, and anyone auditing its verdicts, can see exactly what the input omitted. That's the degraded-not-silently-healthy rule applied to prompt construction, which is precisely where it's usually violated.
- **Recency-biased bounds (`values[-limit:]`)** keep the most recent constraints/files/failures when capping — the right default for reviewing a live streak.
- **Deterministic reconstruction:** hydration rebuilds byte-identical inputs from the replayed snapshot (pinned by the equality assertion in the hydration test), so restart cannot produce a review against a subtly different input than the one originally queued.
- **The input contract is defined before the consumer exists.** Prompt rendering and model execution stay out of scope, which means the expensive step will be built against a frozen, tested contract rather than ad hoc state access — same boundary-first ordering as #170→#171.
- The stress test is thorough: overlong goal, 25 constraints, 60 files, 12 gaps, all verified to cap at their limits with the correct truncation ledger, and the queued event's coverage payload matched against the input.

**Nit (non-blocking):** `_texts` and `_strings` are currently identical implementations — fine to leave as separate names if they're expected to diverge (e.g. different per-item caps), otherwise one of them can go.

No blockers. Merging.

PR #173, opened 2026-08-14
feat(reviewer): run signal reviews asynchronously

YoungJinJung:
## Why

PR #171 made signal-driven review jobs durable and PR #172 bounded their immutable input, but no daemon consumer executed those jobs. This increment connects the queue to actual model-backed shadow review without enabling live intervention.

Related to #28.

## What changed

- add explicit `reviewer.on_signals` opt-in and carry custom config paths into `spotterd`
- run bounded reviewer inputs through a cancellation-safe async Codex subprocess outside the observation hot path
- reuse the existing session/day spend caps and serialize paid model calls
- durably record inference start, decision, error/cap, token/timing, and stale lifecycle events
- keep every verdict shadow-only; no steering or delivery is introduced

Candidate merge/priority policy and the remaining signal families stay scoped to later #28 increments. Live freshness-fenced delivery remains #22.

## Validation

- `.venv/bin/ruff check .`
- `.venv/bin/mypy src tests`
- `.venv/bin/pytest` (525 passed)


YoungJinJung:
## Review

The step where the #170→#171→#172 chain finally touches money: opted-in signal jobs now run through a real async Codex subprocess, shadow-only, under the existing budget ledger. This is the riskiest increment in the sequence and the safety engineering is commensurate. Verified locally on the branch: 525 passed, mypy clean, ruff clean; CI green.

**Spend-safety review (the part that matters most here):**

- **Off by default, and config failure fails toward not spending.** `on_signals` is a validated boolean defaulting to false; an unreadable/invalid config logs to stderr and disables signal reviews rather than falling back to spending. The manifest now records `config_path` at setup so the daemon reads the same config the operator wrote.
- **Budget is charged through the existing reserve/settle ledger *before* inference**, caps produce durable `reviewer_capped` events, ledger failures produce `reviewer_error` — a refused review is as auditable as an executed one. Paid calls are serialized under one lock, so a burst of candidates can't fan out into parallel spend.
- **Freshness is checked twice and honesty wins over optics:** once before reserving (don't spend on a dead target), and again after inference — a verdict whose turn ended mid-flight is journaled with `stale: true` instead of being dropped or presented as actionable. The scheduler distinguishes `review_job_stale` (money possibly in flight) from `review_job_discarded` (never started), and hydration reconstructs that distinction correctly, including the started-then-turn-completed crash window.
- **Cancellation can't leak spend or processes:** daemon shutdown settles the reservation at 0, journals the cancellation, and re-raises; the subprocess runs in its own session with TERM→KILL process-group teardown on timeout/cancel.

**Two details I particularly want to endorse:**

- `SPOTTER_DISABLE=1` on the reviewer subprocess — without it, the reviewer's own `codex exec` would be observed by Spotter's hooks and could recursively generate failure streaks about itself. Easy to miss, ugly to debug later.
- The fence-stripping test plants `<<<END-SPOTTER-TRAJECTORY-DATA>>>` inside the goal text and asserts the prompt contains exactly the wrapper's two fence occurrences — prompt-injection defense at the input boundary, pinned by test rather than asserted in prose. The `spec_drift`-without-goal guardrail also carries over from the periodic reviewer.

Shadow-only is enforced by construction (nothing here can reach `turn/steer`), the docs/status/doctor rows all say exactly what changed, and merge/priority policy stays scoped out. No blockers. Merging.

PR #174, opened 2026-08-14
feat(metrics): measure supervision lifecycle latency

YoungJinJung:
## Why

Signal-driven reviews now execute asynchronously, but the durable metrics surface only exposed queue/inference samples and could not reconstruct the full evidence-to-decision lifecycle without crossing unrelated clocks. #33 requires latency attribution and explicit coverage rather than invented zeroes.

Related to #33.

## What changed

- stamp every newly journaled event with a process-local monotonic receipt timestamp and clock-domain ID alongside existing receipt wall time
- correlate evidence, active signal, queued job, inference start, and terminal events through stable IDs
- report same-clock signal delay, queue delay, inference duration, and detection-to-decision latency with coverage denominators
- expose queued/started/decided/error/cap/discard/stale job counts
- record monotonic queue and failed/cancelled inference timing in the async executor
- leave cross-restart and legacy durations unknown instead of subtracting across clock domains

Control dispatch/adoption timing remains scoped to live delivery work.

## Validation

- `.venv/bin/ruff check .`
- `.venv/bin/mypy src tests`
- `.venv/bin/pytest` (528 passed)


YoungJinJung:
## Review

Adds the third clock domain the metrics layer needed: every journal append now carries a process-local monotonic receipt timestamp plus a per-process clock-domain ID, and the new supervision-lifecycle latencies (evidence→signal, queue, inference, detection→decision) are derived **only within one domain**. Verified locally on the branch: 528 passed, mypy clean, ruff clean; CI green.

**What holds up:**

- **The clock-domain rule is enforced structurally, not by convention.** `_monotonic_elapsed_ms` refuses to subtract across differing clock IDs or backwards values, so a daemon restart shrinks a coverage denominator instead of manufacturing a duration. The lifecycle test proves it the hard way: a cross-domain evidence event plus a wrapped monotonic value yields `signal_delay (1/2)` — one real sample, one honest gap — rather than a garbage number. This is the "source wall time is never subtracted from receipt wall time" rule generalized to a third clock, with the same discipline.
- **Correlation is by stable IDs across the whole chain** (evidence event IDs → signal ID → review job ID → terminal), so latencies attach to the exact lifecycle they measure. The full terminal taxonomy from #173 — decided/errored/capped/discarded/stale — now has visible counts, and the comprehensive test exercises every one of those paths in a single journal.
- **The executor's timing fix matters beyond metrics:** queue/inference timing moved from wall-clock diffs to `perf_counter_ns`, and failed/cancelled inference now records its timing too — previously a failed paid call had spend but no duration, which would have biased inference stats toward successes.
- **Sensible fallback layering** — monotonic pairs preferred, payload-reported `queue_ms`/`inference_ms` as fallback — keeps legacy and cross-restart records contributing where they can without ever crossing clocks.
- The kwargs-ification of the `RuntimeCostReport` constructor was overdue at this field count; positional construction was one reorder away from a silent field swap.

Control delivery/adoption timing staying scoped to #22 is the right line — that leg needs the delivery mechanism to exist first. No blockers. Merging.

PR #175, opened 2026-08-14
feat(replay): fingerprint declared directories

YoungJinJung:
## Summary

- fingerprint declared directory trees with deterministic path, type, and content hashes
- reject nested symlinks and catch ignored-directory loss during source-to-fork preflight
- persist resource type in fork manifest schema v3 while retaining v1/v2 compatibility

Related to #42.

## Verification

- `.venv/bin/ruff check .`
- `.venv/bin/mypy src tests`
- `.venv/bin/pytest` (531 passed)


YoungJinJung:
## Review

Extends the #166 declared-resource guard from single files to directory trees. Verified locally on the branch: 531 passed, mypy clean, ruff clean; CI green.

**What holds up:**

- **The tree digest is deterministic by construction:** children sorted by relative POSIX path, each contributing (type, relpath, content-hash) into canonical JSON — so hash equality means tree equality including empty directories, and the content-change test pins that a nested edit flips the digest.
- **Symlinks are rejected, not followed.** A nested symlink raises instead of silently hashing its target — the traversal-safety property from #166 extended into tree walks, where following links would have been the classic escape hatch out of the worktree. Unsupported entry types (sockets, fifos) also fail loudly rather than being skipped.
- **`kind` joins the drift comparison**, so a file→directory swap at the same path can't slip through on a hash coincidence, and the v2-compat default (`kind="file"`) is exactly right since v2 could only ever have declared files.
- **Schema v3 keeps the compatibility ladder intact** — the migration test now exercises both v2 (kind stripped) and v1 (resources stripped) readbacks, matching the durable-state rule that old manifests stay readable.
- The end-to-end test mirrors #167's proof shape: an ignored `.fixture-cache/` directory lost in restore blocks the fork at source preflight before any arm could run.

**Nit (non-blocking):** a lost ignored *directory* is classified `MISSING_IGNORED_FILE` — the enum name is now slightly narrower than what it covers. Renaming would break recorded values, and the fingerprint's `kind`/`state` fields disambiguate, so leaving it is the right call; just worth remembering when reading drift reports.

No blockers. Merging.

PR #176, opened 2026-08-14
feat(replay): fingerprint environment variables

YoungJinJung:
## Summary

- fingerprint explicitly declared non-secret environment variables by name, presence, and value hash
- classify declared variable drift and recheck it immediately before each continuation
- persist schema v4 provenance while keeping fork manifest schemas v1 through v3 readable

Related to #42.

## Verification

- `.venv/bin/ruff check .`
- `.venv/bin/mypy src tests`
- `.venv/bin/pytest` (533 passed)


YoungJinJung:
## Review

Closes the last declarable gap from fidelity v1's blocking list: explicitly declared non-secret environment variables are now fingerprinted by name, presence, and value hash, checked source→fork and re-checked immediately before each continuation. Verified locally on the branch: 533 passed, mypy clean, ruff clean; CI green.

**What holds up:**

- **Raw values never touch disk.** Only name + set/missing state + SHA-256 of the value are persisted, and both new tests assert the plaintext is absent from the serialized fingerprint *and* the on-disk manifest. Same hash-only contract as declared files, applied to the more secret-prone input class.
- **Presence is part of the contract, both directions.** A declared-but-missing variable records `missing`, so a variable that *appears* in the continuation environment drifts just as loudly as one that vanishes or changes — state and hash are compared together.
- **The #157 late-drift window is covered for variables too:** the arm recheck extracts declared variable names from the immutable manifest and re-fingerprints against the live environment right before `codex exec`. Notably, the recheck now takes a single code path (resources and variables always passed, possibly empty) — the branch shape that caused the #166 false-refusal bug is gone entirely, and the updated happy-path test pins both tuples reaching the fingerprint call.
- **POSIX name validation** rejects malformed declarations up front rather than producing an unmatchable fingerprint, and sorted-name ordering keeps comparison deterministic.
- **Schema v4 with the full ladder tested** — v3 (variables stripped), v2 (kind stripped), v1 (resources stripped) all remain readable in one test.

One semantic worth stating for future readers: the fingerprint reads the *orchestrating process's* environment — which is exactly what the continuation subprocess inherits, so it verifies the right thing — but a wrapper that injects env vars into the child directly would be invisible to this check. That's consistent with the declared-inputs-only philosophy; just worth remembering if such a wrapper ever appears.

With files, directories, and variables all declarable and rechecked, the "undeclared X remains uncaptured" limitation is now a deliberate opt-in boundary rather than a hole. No blockers. Merging.

PR #177, opened 2026-08-14
feat(replay): report signal branch coverage

YoungJinJung:
## Summary

- link each active detector signal to its first subsequent proposal while the signal remains active
- report trigger, follow-up, and exact-forkable follow-up counts with per-trigger provenance
- document detector-trigger coverage as a distinct #42 qualification dimension

Related to #42.

## Verification

- `.venv/bin/ruff check .`
- `.venv/bin/mypy src tests`
- `.venv/bin/pytest` (534 passed)


YoungJinJung:
## Review

The bridge between the two big tracks of the last week: `fork-coverage` now links each active detector signal (the #28 side) to its first subsequent proposal and reports whether that follow-up point is exactly forkable (the #42 side). Verified locally on the branch: 534 passed, mypy clean, ruff clean; CI green.

**What holds up:**

- **This is the missing precondition for fidelity v1's "labeled intervention opportunities" requirement.** Until now, "prefixes near intervention opportunities" was operator judgment; active signal candidates *are* the machine-labeled opportunities, and this report measures whether the replay instrument can actually anchor at them. The three counters (triggers / follow-ups / exact-forkable follow-ups) expose both failure modes separately — a trigger with no follow-up action, and a follow-up that exists but can't be forked.
- **The linkage semantics are conservative and correct:** the scan stops at the signal's own resolved/stale transition, so a proposal that arrives after the streak died is never attributed to it. The test pins both branches — a live trigger linking to a forkable proposal, and a resolved-before-proposal trigger counting as no-follow-up rather than grabbing the next unrelated step.
- **Per-trigger provenance** (signal ID/type, trigger step, source event ID, proposal step, coverage status) makes each row auditable back to the journal rather than being an aggregate-only claim, and JSON serialization is covered.
- **Reuses the existing per-proposal coverage verdicts** instead of re-deriving forkability — one classifier, one truth.
- Kwargs-ifying the growing `BranchCoverageReport` constructor follows the same hygiene as #174's report; docs move in all four places with the status next-step gaining "labeled-opportunity coverage" explicitly.

No blockers. Merging.

PR #178, opened 2026-08-14
feat(signals): detect repeated equivalent tool calls

YoungJinJung:
## Summary
- add an incremental repeated-equivalent-tool-call signal over normalized tool resource and arguments
- emit once at the threshold, journal suppressed growth during cooldown, and reset on mutation, another outcome, failure, target change, or terminal gaps
- keep unsupported argument shapes unknown and preserve signal-specific reviewer features

Advances #28.

## Verification
- `.venv/bin/ruff check .`
- `.venv/bin/mypy src tests`
- `.venv/bin/pytest` (537 passed)

### DIFF
--- a/src/spotter/signals.py
+++ b/src/spotter/signals.py
@@ -1,6 +1,7 @@
 """Cheap incremental candidates derived from runtime-neutral Trace IR."""
 
 import hashlib
+import json
 from collections.abc import Iterable, Mapping
 from dataclasses import dataclass, replace
 from enum import StrEnum
@@ -24,6 +25,7 @@ _EVIDENCE_LIMIT = 20
 
 class SignalType(StrEnum):
     FAILURE_STREAK = "failure_streak"
+    REPEATED_EQUIVALENT_TOOL_CALL = "repeated_equivalent_tool_call"
 
 
 class SignalStatus(StrEnum):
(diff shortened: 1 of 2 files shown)


YoungJinJung:
## Review

Second #28 signal family: repeated equivalent tool calls — the "spinning" detector that was explicitly deferred back in the replay-era ponytail note. Verified locally on the branch: 537 passed, mypy clean, ruff clean; CI green.

**What holds up:**

- **The reset semantics are what make this a spinning detector rather than a repetition counter.** Any intervening mutation or other outcome re-arms the tracker, so read→edit→read-read-read only counts the post-edit reads — the mutation-rearms test pins exactly the case that separates legitimate re-inspection after a change from actual no-progress looping. Failures also reset, which cleanly hands the repeated-*failing*-call case to the failure-streak family: two detectors, disjoint jurisdictions, no double-firing on the same pathology.
- **Equivalence stays inside the no-guessing rule:** exactly one explicit `tool:server/name` resource plus canonical-JSON-hashed arguments (sorted keys, `allow_nan=False`); key-order-insensitive equality is pinned by the test, and unserializable argument shapes yield no signal rather than a fabricated equivalence class (`object()` test).
- **The family-1 machinery is reused, not duplicated:** same emit-once/cooldown/stale lifecycle, evidence cap, per-thread dedup, and — importantly — `_signal_id` now includes the signal type, so the two families can never collide on IDs even for the same key at the same event.
- **Features become properly per-type** (`consecutive_equivalent_calls` vs `consecutive_failures`), carried on the candidate instead of hardcoded at serialization — which also means the #172 bounded reviewer input transports the right feature name downstream with zero changes.
- Default threshold of 3 (vs 2 for failures) is the right asymmetry — benign repeats are more common than benign failure pairs.
- Hydration idempotence carries over for free since the family lives inside the existing `update()`/`hydrate()` replay.

The known per-thread single-tracker limitation from family 1 applies here identically (interleaved A/B/A/B repeats reset each other) — consistent, documented behavior, same upgrade path when it matters. No blockers. Merging.

PR #180, opened 2026-08-14
feat(reviewer): merge related signal jobs

YoungJinJung:
## Summary
- batch signal candidates derived from the same source state before review scheduling
- merge their signal IDs, types, evidence, resources, and namespaced features into one bounded immutable reviewer job
- preserve single-candidate job IDs for journal compatibility and rebuild identical merged jobs during hydration

Advances #28.

## Verification
- `.venv/bin/ruff check .`
- `.venv/bin/mypy src tests`
- `.venv/bin/pytest` (540 passed)

### DIFF
--- a/src/spotter/review_scheduler.py
+++ b/src/spotter/review_scheduler.py
@@ -24,6 +24,9 @@ class ReviewerJob:
     created_at: float | None
     snapshot: ThreadState
     reviewer_input: ReviewerInput
+    signal_ids: tuple[str, ...]
+    signal_types: tuple[str, ...]
+    candidate_event_ids: tuple[str, ...]
 
     def queued_event(self, trigger: TraceEvent) -> TraceEvent:
         return TraceEvent(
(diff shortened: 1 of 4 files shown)


YoungJinJung:
## Review

First half of the deferred #28 merge/priority policy: signal candidates derived from the same source event now batch into one bounded reviewer job instead of racing each other into separate paid reviews. With two families live as of #178, the same failing tool call could plausibly trip both — this makes the natural dedup structural. Verified locally on the branch: 540 passed, mypy clean, ruff clean; CI green.

**What holds up:**

- **The merge invariant is exactly right and loudly enforced:** every merged candidate must share the identical thread/turn/state_version/epoch target, or `_job` raises "merged signals do not share one immutable target". Merging is only ever a same-instant, same-state operation — no cross-state smearing of evidence.
- **The merged payload composes cleanly:** evidence and resources union with order-preserving dedup, features get namespaced by family (`failure_streak.consecutive_failures` vs `repeated_equivalent_tool_call.consecutive_equivalent_calls`) so identically-named features can never collide, severity takes the max, and the synthetic `merged:` signal ID is derived deterministically. The #172 bounded-input builder consumes the merged payload unchanged.
- **Journal compatibility is preserved where it matters:** single-candidate jobs keep the original `{signal_id}:{event_id}` job-ID derivation, so pre-existing journaled jobs re-derive identically on hydration; plural `signal_ids`/`candidate_event_ids` fields are additive on the queue/discard/stale events.
- **Any member signal resolving discards the whole merged job** — conservative and correct: the review was scoped to a state where all members were active, and one resolving invalidates that premise. Re-queuing the remainder can come later if it proves worth the spend.
- **Hydration grouping mirrors the live batch boundary** (consecutive candidates sharing a state_version, suppressions skipped, missing snapshot a hard error), and the double-hydrate test proves merged evidence survives recovery byte-identically. The runtime test closes the loop: two candidates → one queued job → one executor submission.

The remaining half — cross-state prioritization when multiple *pending* jobs compete for budget — stays open, which the docs already track. No blockers. Merging.

PR #181, opened 2026-08-14
feat(reviewer): prioritize pending signal jobs

YoungJinJung:
## Summary
- replace submission-order lock contention with one severity-prioritized pending review queue
- preserve FIFO ordering for equal severity and never preempt an in-flight paid review
- recheck freshness before budget reservation and journal the chosen priority on started/capped outcomes

Advances #28.

## Verification
- `.venv/bin/ruff check .`
- `.venv/bin/mypy src tests`
- `.venv/bin/pytest` (541 passed)

### DIFF
--- a/src/spotter/review_executor.py
+++ b/src/spotter/review_executor.py
@@ -1,6 +1,7 @@
 """Asynchronous shadow execution for durable signal-driven reviewer jobs."""
 
 import asyncio
+import math
 import time
 from collections.abc import Awaitable, Callable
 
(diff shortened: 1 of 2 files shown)


YoungJinJung:
## Review

Second half of the #28 merge/priority policy: pending signal jobs now compete for constrained budget by severity instead of submission order, via a single worker over an `asyncio.PriorityQueue`. Verified locally on the branch: 541 passed, mypy clean, ruff clean; CI green.

**What holds up:**

- **The priority semantics are exactly what a budget-capped shadow reviewer should have:** higher severity first, strict FIFO within equal severity (monotonic sequence tiebreak), and an in-flight paid review is never preempted — the worker finishes what it started. The test pins the whole contract in one scenario: with budget for 2, `low` (already inflight) completes, `high` jumps `medium` for the last slot, and `medium` gets a durable `reviewer_capped` with its priority recorded.
- **Freshness moved to the right moment.** Under the old lock, freshness was checked when the task acquired the lock; now it's checked when the job is dequeued — after however long it sat behind higher-priority work — so a job that went stale while waiting costs nothing.
- **The choice itself is auditable:** priority is journaled on started, capped, and budget-error events, so a postmortem can answer "why did this review run and that one didn't" from the journal alone. That's the #38 precision/miss-rate measurement's raw material being laid down ahead of time.
- **`_priority` is defensively typed** — non-numeric, boolean, NaN, and infinite severity hints all collapse to 0.0 rather than poisoning the heap ordering.
- **Simpler under the hood:** replacing per-job tasks + a shared lock with one worker loop deletes net ~100 lines, and the worker self-heals (submit restarts it if it ever died). Shutdown semantics are preserved — in-flight cancellation still settles the reservation and journals; still-queued jobs are dropped exactly as pre-lock tasks were cancelled before.

With #180 (same-state merge) and this (cross-job priority), the merge/budget policy line item from the #173 scope list is now closed; what remains for #28 is broader signal coverage and the comparative-evidence work against periodic review. No blockers. Merging.

PR #182, opened 2026-08-14
feat(signals): detect reads without frontier growth

YoungJinJung:
## Summary

- add an incremental `repeated_read_no_frontier` signal for successful read tools with explicit resources
- treat new resources, read scopes, result evidence, semantic state changes, and observation gaps as frontier resets
- bound remembered observations and keep unsupported or ambiguous reads unknown
- share the existing signal lifecycle state instead of adding another duplicate streak implementation

Advances #28.

## Verification

- `.venv/bin/ruff check .`
- `.venv/bin/mypy src tests`
- `.venv/bin/pytest` (547 passed)

### DIFF
--- a/src/spotter/signals.py
+++ b/src/spotter/signals.py
@@ -20,12 +20,37 @@ _TERMINAL_KINDS = {
     "thread_deleted",
     "turn_completed",
 }
+_FRONTIER_RESET_KINDS = {
+    "command_result",
+    "diff_updated",
+    "evidence_invalidated",
+    "file_edit",
+    "file_read",
+    "hypothesis",
+    "reasoning_summary",
+    "search",
+    "test_result",
+    "verification_satisfied",
+}
+_READ_VERBS = ("describe", "fetch", "find", "get", "list", "open", "read", "search", "view")
+_READ_RESOURCE_KEYS = {
+    "file": "file",
+    "file_path": "file",
+    "path": "file",
+    "repository": "repository",
+    "resource": "resource",
+    "scope": "resource",
+    "uri": "uri",
+    "url": "url",
+}
 _EVIDENCE_LIMIT = 20
+_FRONTIER_LIMIT = 128
 
 
 class SignalType(StrEnum):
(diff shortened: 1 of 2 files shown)


YoungJinJung:
## Review

Third #28 signal family: successful reads that don't grow the observation frontier. This covers the pathology the first two families can't see — an agent cycling through *slightly different* read calls (different lines, re-listings, re-fetches) that all return already-known information. Verified locally on the branch: 547 passed, mypy clean, ruff clean; CI green.

**What holds up:**

- **The frontier definition is information-theoretic, not call-syntactic:** an observation key hashes (resources, evidence, arguments) together, so a re-read after an edit produces different evidence → a new frontier entry → streak reset, without needing any special mutation handling. The reset taxonomy in the tests is thorough — new resource, new scope (line 1→2), new evidence, ambiguous read, and observation gap each re-arm correctly, and the gap case properly clears the whole frontier (post-gap knowledge claims would be unverifiable).
- **Every classification failure fails toward silence.** Non-read tools never claim the signal even with resource-shaped arguments (the `execute` test), unrecognized tool names produce no observation, missing result/contentItems evidence produces no observation, and unserializable payloads bail — the read-verb prefix heuristic can only cause false *negatives*, which is the correct direction for a hypothesis generator.
- **The `_SignalStreak` consolidation landed at exactly the right moment** — a third copy of the streak dataclass would have cemented the duplication; instead type+feature became data and `_candidate`/`_finish`/`_target_changed` collapsed to single implementations. Family 4 will now be mostly free.
- **The frontier is bounded (128 keys)** so memory can't grow with session length; eviction by sorted-hash order is arbitrary but deterministic — fine as a ceiling, worth a `ponytail:`-style note if eviction-order artifacts ever show up in practice.
- Lifecycle mechanics (emit-once, cooldown, stale-on-target-change/terminal, hydration replay) are inherited from the shared machinery rather than re-implemented, so the #170/#171 recovery guarantees extend to this family with no new code.

No blockers. Merging.

PR #183, opened 2026-08-14
feat(metrics): report detected repeated actions

YoungJinJung:
## Why

#33 requires repeated-action cost to be visible without creating a second action-equivalence heuristic. The durable #28 signal lifecycle now provides that semantic source.

## What changed

- project repeated-equivalent actions and reads without frontier expansion from durable signal features
- deduplicate active, cooldown, stale, and resolved updates by stable signal lifecycle ID and use the maximum observed count
- report feature coverage explicitly; old or partial records remain unknown instead of looking like zero repetition
- refresh the status surface for the third signal family and merged priority scheduling

## Verification

- `.venv/bin/ruff check .`
- `.venv/bin/mypy src tests`
- `.venv/bin/pytest` (548 passed)
- `.venv/bin/spotter metrics` against the retained local corpus

Part of #33.


YoungJinJung:
## Review

The clean landing of a note from the #164 review: repeated-action cost is now visible in `spotter metrics` **without a second action-equivalence heuristic** — the #28 signal lifecycles are the single semantic source, and the metrics layer only projects their durable features. Verified locally on the branch: 548 passed, mypy clean, ruff clean; CI green.

**What holds up:**

- **One classifier, one truth, again.** The projection reads `consecutive_equivalent_calls` / `reads_without_frontier_expansion` off journaled candidate/suppression events, deduplicates by stable signal lifecycle ID, and takes the max observed count per lifecycle (cooldown updates carry the growing tally). No re-derivation of equivalence anywhere in the metrics path.
- **The waste arithmetic is subtle and correct:** equivalent-call lifecycles subtract one (`first_is_not_repeated=True`) because the first call was legitimate work, while no-frontier reads don't — every read counted by family 3 was already redundant by definition. A count of 5 equivalent calls reports 4 wasted actions; 4 no-frontier reads report 4. The test pins both.
- **Coverage discipline holds at the lifecycle granularity:** a lifecycle without the expected feature (older records, partial journals) widens the eligible denominator without contributing a fabricated zero — `equivalent_actions=4 (1/2 signal lifecycles)` says exactly what's known and what isn't, and the empty case renders `unknown (0/0)`.
- **Failure streaks are deliberately excluded** from repetition counts — they measure failure cost, not repetition cost, and conflating them would have double-counted a repeated failing call against both axes.
- Status rows catch up on the family-3 and merged-priority state in the same PR.

With this, the #33 line has action families, resources, tokens, lifecycle latency, and now repetition waste all sourced from durable records with explicit coverage — the "efficiency diagnosis" goal that started the metrics series is substantially in place, pending only delivery/adoption telemetry. No blockers. Merging.

PR #184, opened 2026-08-14
feat(replay): report labeled opportunity coverage

YoungJinJung:
## Summary
- report whether current human-labeled gate and reviewer intervention opportunities have an exact replay branch point
- correlate gate labels by tool-use identity and reviewer labels with the next proposal, while keeping stale labels explicit
- fail the CLI clearly on corrupt label evidence and update the #42 status

## Verification
- `.venv/bin/ruff check .`
- `.venv/bin/mypy src tests`
- `.venv/bin/pytest` (549 passed)

Advances #42.


YoungJinJung:
## Review

The human-labeled complement to #177's machine-labeled triggers: `fork-coverage` now maps gate and reviewer labels to the proposal where an intervention could actually branch, closing the "labeled-opportunity coverage" next-step from the status row. Verified locally on the branch: 549 passed, mypy clean, ruff clean; CI green.

**What holds up:**

- **The direction semantics are correct per opportunity type.** A gate label points *backwards* — the branch point is the proposal that would have been blocked, correlated by `tool_use_id`; a reviewer label points *forwards* — the intervention would shape the next proposal after the verdict. Getting these directions confused would have quietly measured the wrong prefixes, and the test pins both mappings (`proposal_step` 0 for the gate label at step 1, 3 for the reviewer label at step 2).
- **Legacy correlation is adjacency-only, with the reasoning in the comment:** gate events that predate the correlation ID are recovered only when the immediately preceding record is a proposal — "only adjacency is strong enough evidence" is exactly the no-guessing bar applied to historical data.
- **Stale labels stay in the report but out of the mapping.** A label whose fingerprint no longer matches the journal counts toward `labeled_opportunities` but not `labeled_opportunities_current`, and gets no branch point — the deliberately-corrupted-fingerprint test proves stale evidence can't smuggle in a branch claim. This mirrors the label-staleness discipline the audit ledger already uses.
- **Corrupt label files fail the CLI loudly** (`LabelError` now caught and reported at the `fork-coverage` entry point) rather than producing a partial report.
- Reuses the per-proposal coverage verdicts from the existing classifier — same one-truth pattern as #177.

With machine triggers (#177) and human labels (#184) both mapped to forkable branch points, the #42 sample-selection question — "which prefixes are worth spending neutral pairs on" — is now answerable from a single report instead of operator archaeology. No blockers. Merging.

PR #185, opened 2026-08-14
feat(signals): detect recurrence after gate blocks

YoungJinJung:
## Summary
- carry exact Hook thread, turn, and tool identity through the daemon gate request
- project deterministic rejections into the live App Server timeline only when the active turn and connection epoch still match
- emit a bounded, cooldown-aware candidate on the second semantically equivalent blocked attempt while leaving unknown and different-resource attempts unclassified

## Verification
- `.venv/bin/ruff check .`
- `.venv/bin/mypy src tests`
- `.venv/bin/pytest` (553 passed)

Advances #28.


youngjinjung-linq:
## Review

Fourth #28 signal family, and the deepest integration in the sequence: deterministic gate rejections from the synchronous Hook path are projected into the App Server signal timeline, so an agent retrying a semantically equivalent blocked action becomes a durable candidate. Verified locally on the branch: 553 passed, mypy clean, ruff clean; CI green.

**What holds up:**

- **The hot path stays untouched.** The gate response is written and the connection closed *before* `record_gate_decision` runs, and an observation failure degrades daemon health visibly instead of affecting the gate verdict — the "bounded and deterministic PreToolUse" rule survives contact with the signal system, which was the main thing to get wrong here.
- **Live correlation is exact-match-or-nothing.** Projection requires exactly one control-ready thread state whose external thread *and* turn match the Hook identity, with a known epoch; ambiguity, a wrong turn, or a missing epoch drops the observation silently rather than attributing a block to the wrong timeline. The test drives all three outcomes (wrong-turn dropped, matching projected, second attempt firing).
- **The equivalence key is the smartest reuse in the family set:** for command gates it's the gate's *rule*, not the command text — the shell-aware gate already proved `git reset --hard` and `bash -c 'git reset --hard'` are the same operation, so the signal layer inherits that classification instead of growing a second command parser. The comment says exactly why, and the test pins the wrapper-equivalence property. File gates key on the explicit file set. Unknown shapes stay unclassified.
- **Threshold 2 is right for this family** — a *second* equivalent blocked attempt is already evidence the agent didn't adapt to the rejection, unlike benign repeats elsewhere. Per-(thread, key) tracking with a bounded map (128, stale-evict), deterministic projected event IDs (dedup on replay via ingestion), and the shared emit/cooldown/stale/hydration machinery all carry over.
- The status row now names the four families and — usefully — pre-declares the next ones (touched-scope, edit-without-validation, stale-hypothesis, budget-anomaly), which keeps #28's remaining scope concrete.

No blockers. Merging.

PR #186, opened 2026-08-14
feat(signals): detect touched scope growth

YoungJinJung:
## Summary
- derive a bounded explicit path scope from the active user request without treating unscoped tasks as narrow
- emit a cooldown-aware candidate when successful edits accumulate across three new files outside that declared scope
- keep broad scopes, failed edits, oversized inputs, unsupported paths, and new user goals conservative and explicit

## Verification
- `.venv/bin/ruff check .`
- `.venv/bin/mypy src tests`
- `.venv/bin/pytest` (556 passed)

Advances #28.


YoungJinJung:
## Review

Fifth #28 signal family: scope creep — successful edits accumulating across distinct new files outside the path scope the user's request explicitly declared. Verified locally on the branch: 556 passed, mypy clean, ruff clean; CI green.

**What holds up:**

- **The load-bearing conservatism is "unscoped ≠ narrow."** A prompt with no recognizable path tokens produces no declared scope and the family stays silent — "refactor the project" can never generate scope-creep candidates. Broad scopes work as directories (`src/` covers `src/pkg/b.py`), failed edits don't count, and oversized prompt text falls back to explicit `expected_paths` only. Each of those is pinned by the unscoped/broad test.
- **A new user prompt both re-derives the scope and stales the active growth** — the goal changed, so the prior out-of-scope claim is no longer valid evidence. That's the same target-invalidates-hypothesis rule the scheduler applies to jobs, enforced at the detection layer.
- **Path extraction is bounded and traversal-safe:** 4KB text limit, token allowlist regex, rejection of absolute/`~`/`..`/URL forms, punctuation stripping for the realistic `` `src/api.py` `` case. Severity is the count of *distinct* out-of-scope files (deduped against the streak's resource set), which composes naturally with the #181 priority ordering — wider creep reviews first.
- **Honest caveat worth recording:** this is the first family whose equivalence input is user prose, and partial path extraction fails toward a possible false positive (a path the user mentioned in an unrecognized format counts as out-of-scope). That's acceptable *because* candidates are hypotheses feeding a shadow reviewer rather than verdicts, the threshold requires three distinct files, and total extraction failure fails silent — but it makes this family the natural first subject for the #38 precision measurement once comparative evidence work starts.
- Lifecycle, hydration, terminal/target staling, and the status-row family census (5 down, 3 named remaining) all follow the established pattern.

No blockers. Merging.

PR #187, opened 2026-08-14
feat(signals): detect edits without validation

YoungJinJung:
## Why

Issue #28 requires cheap evidence-driven detection when distinct edits accumulate without proven validation, without treating one narrow test as validation for unrelated packages.

Related: #28

## What changed

- emit active and cooled-down candidates as distinct successful file edits accumulate
- clear and re-arm only the file or directory scopes explicitly proven by `validated_paths`
- keep failed edits, failed tests, unknown validation scope, and unsupported paths from creating false certainty
- preserve scope-aware `edits_since_validation` in ThreadState and document the Trace IR relation

## Validation

- `.venv/bin/ruff check .`
- `.venv/bin/mypy src tests`
- `.venv/bin/pytest` (560 passed)

## Notes

Stale-hypothesis reuse and budget-anomaly detection remain as the final two initial #28 signal families.


YoungJinJung:
## Review

Sixth #28 signal family — distinct successful edits accumulating without proven validation — but the most important change here is actually to ThreadState itself. Verified locally on the branch: 560 passed, mypy clean, ruff clean; CI green.

**The ThreadState fix is a real correctness improvement, not just signal plumbing:**

Previously a passing `test_result` cleared *all* of `edits_since_validation`. Now clearing is scope-matched: only paths within explicit `validated_paths` are cleared, and — the sharper edge — **a pass with no declared scope clears nothing**. A narrow unit test can no longer masquerade as global validation, and unknown validation coverage stays unknown instead of being optimistically zeroed. That's the unknown-not-zero rule applied to the one place in live state where it had been silently violated, and both directions are pinned by the thread-state tests. The architecture doc records the contract change in the same PR.

**The signal family itself holds up:**

- **Partial validation resolves and re-arms precisely:** validating `src/pkg` RESOLVEs the streak and seeds a *fresh* streak (new signal ID) from the surviving files, with evidence remapped to each remaining file's original edit event via the per-resource source map — the re-arm test asserts the surviving evidence is exactly the unvalidated files' edit events, not a stale mixed list.
- **Nothing creates false certainty:** failed edits don't count, failed tests don't clear, unscoped passes don't clear, unsupported paths are dropped by the same traversal-safe normalization as family 5. If the streak survives above threshold after partial validation, it continues as a cooldown update rather than re-firing active.
- Severity is the count of distinct unvalidated files — monotone with risk, composing with #181 priority ordering.
- The scope-growth tests isolating this family via a high threshold is the right way to keep the two file-oriented families from cross-contaminating each other's assertions.

**Nit (non-blocking):** `_validation_scopes`/`_within_scope` in thread_state.py near-duplicate `_normalized_paths`/`_within_scope` in signals.py. Module isolation justifies it today (thread_state shouldn't import signals); if a third consumer appears, that's the moment for a shared `paths` helper.

Six families down, two named to go. No blockers. Merging.

PR #188, opened 2026-08-14
feat(signals): detect stale hypothesis reuse

YoungJinJung:
## Why

Issue #28 requires a stale-hypothesis reuse candidate, but history containing an old claim must not itself be treated as proof that a later action relied on it.

Related: #28

## What changed

- retain an incremental index of transitively stale hypothesis IDs in ThreadState
- emit and cool down candidates only for action events with explicit `hypothesis_ids` matching that index
- keep missing, unknown, oversized, and still-fresh references silent instead of guessing causality
- include only the related bounded stale claim text in the immutable reviewer input
- replay ThreadState during SignalEngine hydration so stale-reuse cooldown survives restart

## Validation

- `.venv/bin/ruff check .`
- `.venv/bin/mypy src tests`
- `.venv/bin/pytest` (563 passed)

## Notes

Budget-anomaly detection remains as the final initial #28 signal family.


YoungJinJung:
## Review

Seventh #28 family — actions explicitly linked to stale hypotheses — with two quiet infrastructure improvements that matter beyond the family itself. Verified locally on the branch: 563 passed, mypy clean, ruff clean; CI green.

**What holds up:**

- **The causality bar is stated and enforced:** "history containing an old claim must not itself be treated as proof that a later action relied on it." Only actions carrying explicit `hypothesis_ids` that intersect the stale index fire; unlinked actions on the same resources, unknown IDs, oversized lists, and references made *while the hypothesis was still fresh* all stay silent — every one pinned in the first test. This is the no-guessing rule applied to causal attribution, the place it's hardest to resist inferring.
- **Transitive staleness lives in ThreadState, incrementally:** invalidation records existing stale hypotheses into `stale_hypothesis_ids`, and a *new* hypothesis depending on a stale one is born STALE and indexed — so the chain claim→dependent-claim→action is caught without any retroactive graph walk. The h1→h2 propagation test pins it.
- **Immediate-fire semantics are right for this family:** one action explicitly relying on an invalidated claim is already reportable (unlike benign repeats), so the first linked action goes ACTIVE and later ones cool down with growing count/resources — per-(thread, hypothesis) so distinct stale claims produce distinct candidates.
- **Hydration quietly got more correct for every family:** the approximate per-thread version counter is replaced with a real `ThreadStateStore` replay, so hydrated candidates now carry exact state versions — closing a small drift between live and recovered state_version that predated this PR.
- **The reviewer input carries exactly the relevant claims:** only stale hypotheses referenced by the candidate's `hypothesis:` resources, bounded at 20, with the count in the coverage dict — the end-to-end test proves the claim text reaches the queued job's bounded input.
- `update()` grows an optional `state` parameter with a None default, so existing callers and the derived-event no-op path are untouched.

Seven down, budget-anomaly to go. No blockers. Merging.

PR #189, opened 2026-08-14
feat(signals): detect relative budget anomalies

YoungJinJung:
## Why

Issue #28 requires budget-anomaly candidates without one absolute tool, token, or time threshold for every repository and task.

Related: #28

## What changed

- emit context-pressure candidates only from explicit `totalTokens / modelContextWindow` capacity
- learn bounded same-turn duration medians per semantic operation/resource and emit only after three baseline samples exceed the configured relative factor
- keep missing capacity, unstable turn identity, unscoped duration, unsupported values, and unrelated operations unknown instead of guessing
- cool down repeated anomalies, resolve on proven recovery, stale on target boundaries, and hydrate both relative baselines and cooldown state
- record all eight initial #28 signal families as implemented while leaving comparative evidence to #38/#24/#33

## Validation

- `.venv/bin/ruff check .`
- `.venv/bin/mypy src tests`
- `.venv/bin/pytest` (566 passed)

## Notes

Repeated-action and no-frontier families already cover tool-call inefficiency; this family adds the directly observable relative token-capacity and operation-duration pressure without parsing opaque commands.


youngjinjung-linq:
## Review

Eighth and final initial #28 family: budget anomalies that are relative by construction — no global token, tool, or time threshold pretending to fit every repository. Verified locally on the branch: 566 passed, mypy clean, ruff clean; CI green.

**What holds up:**

- **Both sub-detectors derive their reference from the observation itself.** Context pressure fires only when *both* `totalTokens` and `modelContextWindow` are explicit — capacity is never guessed — and duration anomalies compare against a same-operation/resource median built from at least three same-turn samples. Missing capacity, missing duration, missing resource identity, or an unstable turn all stay unknown, exactly as the architecture note now records.
- **The baseline can't be poisoned by the anomalies it detects:** anomalous durations are excluded from the median window, while sub-threshold samples both feed the baseline (bounded at 20) and RESOLVE any active streak. Without that asymmetry, a run of slow operations would normalize itself out of detection within a few samples.
- **RESOLVED-on-recovery is the right lifecycle here** — unlike streak families, budget pressure has a directly observable exit condition (utilization dropping below threshold), and this family uses it instead of only staling at boundaries.
- **Severity is the magnitude itself** (utilization percent / duration-vs-median percent, clamped), so the #181 priority queue naturally reviews a 95%-full context ahead of a mild scope creep, and the per-candidate features carry the raw ingredients (window, total, baseline samples, median) for the reviewer to reason from.
- Config validation rejects nonsensical thresholds (≤100% duration factor, out-of-range window percent), numeric parsing is strict (non-bool, finite, non-negative), and hydration/target-staling extend to both streaks and baselines.

**The milestone:** all eight initial families are now implemented, and the roadmap/status rows correctly pivot the #28 next step from "build detectors" to "measure them" — comparative precision/miss/delay/cost against periodic-only review in #38/#24/#33. That's the "separate implementation from evidence" rule applied at the moment it becomes tempting to skip: eight working detectors prove the mechanism, not the benefit. No blockers. Merging.

PR #190, opened 2026-08-14
docs(experiment): report edit-decision replay noise

YoungJinJung:
## Summary

- run one real identical-neutral replay pair from the retained record-contract first-edit decision prefix
- publish the scoped machine-readable result and interpretation without committing transcript-derived raw JSONL
- update fork-fidelity v1 aggregation while retaining the NO-GO decision for representative causal use

## Result

- 1/1 judgeable pair
- 2/2 scorer passes
- 0/1 mechanical disagreements
- 0/1 environment mismatches
- 0/2 infrastructure failures

This extends the legacy unpinned-effort stratum to 12 pairs across four prefixes and three tasks. It does not add a final-outcome failure, so #42 remains open.

## Verification

- `.venv/bin/ruff check .`
- `.venv/bin/mypy src tests`
- `.venv/bin/pytest` (566 passed)
- `jq empty docs/experiments/fork-neutral-edit-decision.json docs/experiments/fork-fidelity-v1.json`

Related to #42.


youngjinjung-linq:
## Review

**Summary:** Sixth fresh identical-arm run — one neutral pair from the retained `record-contract` session at step 8, immediately before the first `apply_patch`. 1/1 judgeable, 2/2 passing arms, 0 disagreements/mismatches/infra failures. Extends the legacy unpinned-effort stratum to 12 pairs across 4 prefixes, and fork-fidelity-v1 retains NO-GO.

**Verified:**
- Arithmetic checks out everywhere: 11→12 pairs, 22→24 arms, 4 prefixes; new bound 0.2209 = 1 − 0.05^(1/12) (recomputed); total neutral arms 28→30 in the v1 narrative.
- Honest accounting of what this does and doesn't add: tasks stay at 3 because the step-8 prefix reuses the `record-contract` task family, and the report explicitly says the new prefix "must not tighten a representative causal bound." Effort stays in the unpinned stratum because the source didn't persist it, consistent with the parity contract (no false pinning claim).
- The blocking-gap rewording ("undeclared ignored files…") correctly reflects that declared resources/directories/env-vars are now capturable — only undeclared ones remain a gap.
- The prose fix "three and one independent task families" (was "prefixes") is the right correction now that prefixes (4) and task families (3) diverge in the legacy stratum.
- Raw JSONL provenance recorded with SHA-256 and matching `raw_result_sha256` in the derived JSON.
- Local worktree: pytest, mypy (79 files), ruff all clean; jq validation of both experiment JSONs (pair/arm/bound fields) passes. PR CI green.

**One note (non-blocking):** this is the second in-place update to the versioned fork-fidelity-v1 artifact. Extending the same qualification's aggregation while the decision is unchanged is reasonable, but if a future run ever flips or materially reframes the decision, that should land as a v2 supersession rather than another in-place edit.

No blockers. Merging.

PR #191, opened 2026-08-14
feat(replay): capture task batch sources

YoungJinJung:
## Summary

- add an explicit `--capture-replay-sources` mode to frozen task batches
- initialize captured fixtures as local Git baselines and retain their source workspaces under `SPOTTER_HOME/task-sources`
- run Hooks in capture-only mode so journals and snapshots are recorded without daemon gating or shadow-review spend
- persist per-arm replay source session IDs or explicit capture errors, and fence resume on the capture setting

## Why

#42 needs final-outcome failures as replay prefixes, but task batches previously set `SPOTTER_DISABLE=1` for every child arm. A failed arm therefore had a mechanical outcome but no Hook journal, Git snapshot, or durable workspace to fork. This closes that source-capture gap without changing the default isolated batch behavior.

## Verification

- `.venv/bin/ruff check .`
- `.venv/bin/mypy src tests`
- focused test proves a captured arm can be forked later from its retained baseline

Related to #42.


YoungJinJung:
## Review

**Summary:** Adds `--capture-replay-sources` to `tasks run` so batch arms can become fork sources for #42. Capture mode initializes each fixture copy as a local Git baseline, runs the child with `SPOTTER_CAPTURE_ONLY=1` (journals + snapshots on, daemon gating and shadow review off), retains workspaces under `SPOTTER_HOME/task-sources`, and records per-arm either a verified replay-source session ID or an explicit capture error. Resume is fenced on the capture setting. This closes the real gap where `SPOTTER_DISABLE=1` meant a failed arm had a mechanical outcome but nothing to fork.

**What I like:**
- The capture-only hook contract is enforced in depth: `run_hook` forces `observation_only=True` + reviewer off, `_gate_over_ipc` short-circuits to the local Gate with an explicit `capture_only` status record, and the test proves the daemon is never called while the shadow-block journal trail (`tool_proposal` → `gate_shadow_block` → `gate_ipc`) stays intact. No supervision spend, no recursion, still diagnosable.
- Capture verification is honest end-to-end: `_replay_source` doesn't just grab `thread.started` — it checks the Spotter journal exists with a snapshot and the Codex rollout is findable, and each failure mode gets a distinct named error instead of a silent null. The main test then actually `fork()`s from the captured baseline and asserts fixture content, which proves the whole point of the PR.
- Env hygiene both ways: capture mode pops an inherited `SPOTTER_DISABLE`, normal mode pops a stale `SPOTTER_CAPTURE_ONLY` — and both directions are tested.
- `TimeoutExpired` still attempts `_replay_source` from partial stdout, so timed-out arms (a likely failure class worth forking) can be captured too.
- Resume fencing on `capture_replay_sources` prevents mixing captured and uncaptured rows in one batch file.

**Non-blocking notes:**
1. Capture mode retains *every* arm's workspace (including `PASS`), and `task-sources` has no retention/pruning story yet — fine for now given #42 needs the sources, but worth folding into the #89 retention lifecycle before large batches.
2. The `if capture_replay_source: … else: …` duplication around the two `_run_task_agent` calls could collapse to a single call passing the flag; the branch only exists so older monkeypatched fakes without the kwarg keep working.
3. In `_replay_source`, a `thread.started` with a malformed `thread_id` breaks out and reports "did not report a thread.started session" — slightly imprecise wording for that case, though still fail-visible.

**Verified locally** in a clean worktree: pytest, mypy (79 files), ruff all pass. PR CI fully green (checks ×2, CodeQL, Analyze ×3). README/status.md updated with the contract change.

No blockers. Merging.

PR #192, opened 2026-08-14
fix(replay): enable hooks in captured task workspaces

YoungJinJung:
## Bug

`tasks run --capture-replay-sources` initialized each fixture as a fresh scratch repository, but Codex did not trust configured Hooks in that repository. Agent arms and scorers completed while Spotter journals were never created, so every arm reported a missing replay source.

## Root cause and fix

The capture child used `--skip-git-repo-check` but did not opt into the Codex automation flag that permits already-vetted configured Hooks. Capture mode now adds `--dangerously-bypass-hook-trust`; normal task arms remain unchanged and continue with Spotter disabled.

This path is already gated by frozen task validation, preflight, explicit `--capture-replay-sources`, and explicit paid-run consent.

## Verification

- focused task-corpus tests
- `.venv/bin/ruff check .`
- `.venv/bin/mypy src tests`
- real `validation-v1` batch: control `PASS`, guidance `TASK_FAIL`, replay sources `2/2 captured`

Related to #42.

### DIFF
--- a/src/spotter/task_corpus.py
+++ b/src/spotter/task_corpus.py
@@ -480,7 +480,7 @@ def _run_task_agent(
         args += ["--model", model]
     args += ["--skip-git-repo-check", "--sandbox", sandbox]
     if capture_replay_source:
-        args.append("--json")
+        args += ["--dangerously-bypass-hook-trust", "--json"]
     args.append(prompt)
     env = {**os.environ}
     if capture_replay_source:
--- a/tests/test_task_corpus.py
+++ b/tests/test_task_corpus.py
@@ -406,10 +406,12 @@ def test_task_agent_capture_mode_enables_json_hooks_without_supervision(
 
     capture_args, capture_env = calls[0]
     assert "--json" in capture_args
+    assert "--dangerously-bypass-hook-trust" in capture_args
     assert capture_env["SPOTTER_CAPTURE_ONLY"] == "1"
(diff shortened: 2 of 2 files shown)


YoungJinJung:
## Review

**Summary:** Two-line fix on top of #191: capture-mode child arms now pass `--dangerously-bypass-hook-trust` so Codex actually runs the configured Spotter Hooks inside the freshly `git init`-ed scratch workspace. Without it, arms completed but no journal was ever written, so every arm reported a missing replay source. Normal (non-capture) arms are unchanged.

**Assessment:**
- This is exactly the class of gap the #191 test suite couldn't catch — the capture test monkeypatches `_run_task_agent`, so real Codex hook-trust behavior was never exercised. Finding it via a real `validation-v1` batch (2/2 captured, including a genuine `TASK_FAIL` guidance arm — the first captured failed outcome for #42) is the right kind of evidence, and the PR reports it honestly.
- Scoping is correct: the bypass flag is added only in the `capture_replay_source` branch, and the test now asserts both its presence in capture mode and its absence in normal mode.
- Risk is acceptable: the flag trusts the user's own already-configured Hooks inside a Spotter-created scratch copy of a frozen fixture, and the path is fenced behind task-set validation, preflight, the explicit `--capture-replay-sources` flag, and paid-run consent.

**Verified locally** in a clean worktree: pytest, mypy (79 files), ruff all pass. PR CI fully green.

No blockers. Merging.

PR #193, opened 2026-08-14
feat(tasks): pin replay-source reasoning effort

YoungJinJung:
## Why

#42 now captures mechanically failed task arms as replay sources, but task batches could pin only the model. A source rollout created under an implicit reasoning effort cannot prove source/continuation parity when the later neutral experiment pins effort.

## What changed

- allow `tasks run --reasoning-effort <value>`
- pass the pin to Codex as `model_reasoning_effort`
- persist it in task-batch provenance and fence resume on changes
- treat a missing field in schema-1 batches as `codex-config-default` for backward-compatible resume
- document model/effort-pinned source capture

## Verification

- focused task-corpus tests
- `.venv/bin/ruff check .`
- `.venv/bin/mypy src tests`

## Next

After merge, rerun the captured `TASK_FAIL` source at pinned low effort and execute neutral replay pairs from its exact pre-mutation prefix.

Related to #42.


YoungJinJung:
## Review

**Summary:** Adds `--reasoning-effort` to `tasks run`, passed to Codex as `model_reasoning_effort` config, recorded in task-batch provenance, and fenced on resume. This closes the provenance gap that kept every fork-fidelity source in the unpinned-effort stratum: captured replay sources can now be created with pinned effort, so later neutral pairs can prove source/continuation effort parity instead of falling back to `codex-config-default`.

**Verified:**
- The effort pin uses the same `--config model_reasoning_effort="..."` mechanism as the fork experiment side, so source and continuation pins are directly comparable — exactly what the parity check needs.
- Resume fencing is complete and backward compatible: a schema-1 batch header missing the field reads as `codex-config-default`, the test simulates an old header by popping the field and resumes cleanly with no pin, and resuming with a different pin raises a named error. Mixing effort within one batch file is impossible.
- Provenance lands in the header (`reasoning_effort: <value|codex-config-default>`), asserted in tests, and every `_run_task_agent` fake across the suite was updated to the new signature — including the exact-args assertion that `--config model_reasoning_effort="low"` reaches the Codex command line.
- CLI plumbing is minimal reuse of the existing `--reasoning-effort` argument (help text widened), consistent with the experiment command.
- README/status.md updated with the model/effort-pinned capture contract.

**Verified locally** in a clean worktree: pytest, mypy (79 files), ruff all pass. PR CI fully green.

The stated next step — rerunning the captured `TASK_FAIL` source at pinned low effort and replaying neutral pairs from its pre-mutation prefix — is the right sequencing: it would produce the first pinned-effort stratum entry sourced from a mechanically failed outcome, which is what #42's qualification has been missing.

No blockers. Merging.

PR #195, opened 2026-08-14
feat(replay): classify submodule drift

YoungJinJung:
## Why

Issue #42 defines submodule divergence as a distinct replay-environment drift category, but the existing fingerprint comparison collapsed it into generic tracked-state drift.

## What changed

- report submodule-status differences as `SYMLINK_OR_SUBMODULE_MISMATCH`
- retain `TRACKED_STATE_MISMATCH` for ordinary Git state changes and report both when both differ
- document the classifier and existing declared-resource symlink rejection

## Validation

- `.venv/bin/ruff check .`
- `.venv/bin/mypy src tests`
- `.venv/bin/pytest` (571 passed)

Related to #42.


YoungJinJung:
## Review

**Summary:** Splits `submodule_status_sha256` out of the generic tracked-state fingerprint tuple so submodule divergence is reported as its own `SYMLINK_OR_SUBMODULE_MISMATCH` drift category, matching the taxonomy #42 defines. Ordinary Git state changes keep `TRACKED_STATE_MISMATCH`, and when both differ, both are reported.

**Verified:**
- The refactor is a pure reclassification: the union of compared fields is unchanged, so nothing that previously counted as drift can now pass silently — the same divergence just gets a more precise name. This preserves the fail-visible property of the preflight.
- The test covers exactly the two interesting cases: submodule-only drift yields only the new category, and combined drift yields both, in deterministic order.
- The enum's symlink half is honest: declared-resource trees containing symlinks were already rejected before continuation, and the README now documents that alongside the new classifier instead of implying new symlink detection.
- Named-category drift beats a lumped mismatch for diagnosis — an experiment INFRA_FAILing on submodule drift now says so directly rather than pointing at generic tracked state.

**Verified locally** in a clean worktree: pytest (571), mypy (79 files), ruff all pass. PR CI fully green. README/status.md updated.

No blockers. Merging.

PR #196, opened 2026-08-14
feat(replay): detect absolute path drift

YoungJinJung:
## Why

Issue #42 requires detached-worktree drift to distinguish hard-coded absolute paths. Declared resources and environment variables were hashed, but a copied value that still pointed at the source worktree could pass source-to-fork preflight unchanged.

## What changed

- record a secret-free boolean when a declared resource or environment variable contains its current worktree path
- classify source/fork reference changes as `ABSOLUTE_PATH_MISMATCH` and deduplicate the category across declared inputs
- persist the field in fork manifest schema v5 while keeping schemas v1-v4 readable
- document the new preflight behavior and current qualification status

## Validation

- `.venv/bin/ruff check .`
- `.venv/bin/mypy src tests`

Related to #42.


YoungJinJung:
## Review

**Summary:** Detects hard-coded absolute-path drift in declared resources and environment variables. Each fingerprint now records a secret-free boolean — "does this value contain the *current* worktree's absolute path" — computed against the source worktree on the source side and the fork worktree on the restored side. A value copied verbatim from source therefore flips True→False and is classified `ABSOLUTE_PATH_MISMATCH`, even when its content hash is identical. Fork manifest schema goes to v5 with v1–v4 kept readable.

**What makes this correct:**
- The ordering is the crux and it's right: the boolean comparison precedes the sha256-equality short-circuit in `_declared_resource_drift`. That's exactly the previously-undetectable case — identical bytes, identical hash, but still pointing at the source worktree — and it can now no longer pass preflight silently.
- The relative-vs-absolute semantics fall out cleanly: a value with no worktree reference matches (False/False), and a value that was *regenerated* to reference the fork's own worktree also matches (True/True) — only stale cross-worktree references trip the check. Direction-agnostic mismatch is also caught.
- Secret discipline preserved: only a boolean is persisted, never the value; consistent with the existing hash-only capture.
- Env-var drift now reports categories additively (`ABSOLUTE_PATH_MISMATCH` alongside `ENVIRONMENT_VARIABLE_MISMATCH` when both apply) instead of collapsing to one, and `fork()` deduplicates categories across resources and variables while preserving order.
- Schema v5 migration is by the book: version accepted set extended, missing field defaults to False for older manifests, and the v1–v4 readability test now explicitly strips the new field to prove v4 loads.
- The end-to-end test covers both channels at once (a JSON fixture and an env var carrying the source path), asserts the composite `SOURCE_ENVIRONMENT_MISMATCH:ABSOLUTE_PATH_MISMATCH` preflight, and checks the manifest's fork-side booleans.

**One non-blocking note:** in `_declared_resource_drift`, a source resource whose content referenced the worktree (True) that is entirely *missing* after restore (default False) now classifies as `ABSOLUTE_PATH_MISMATCH` rather than `MISSING_IGNORED_FILE`/`MISSING_UNTRACKED_FILE`, since the boolean check takes precedence. Both outcomes block the pair, so nothing is lost in safety — but "missing" is arguably the more diagnostic label there. Worth a precedence tweak if it ever shows up in practice.

**Verified locally** in a clean worktree: pytest (572), mypy (79 files), ruff all pass. PR CI fully green. README/lifecycle/status docs updated with the new classifier and schema version.

No blockers. Merging.

PR #197, opened 2026-08-14
fix(replay): prioritize missing resource drift

YoungJinJung:
## Why

PR #196 added absolute-worktree-path drift detection. Its review noted that a declared ignored or untracked resource containing such a path would be labeled `ABSOLUTE_PATH_MISMATCH` when the resource was entirely missing after restore, hiding the more diagnostic missing-resource category.

## What changed

- classify missing ignored/untracked resources before comparing worktree-path references
- keep `ABSOLUTE_PATH_MISMATCH` for resources that exist on both sides but retain the wrong worktree reference
- extend the source-to-fork regression case with an ignored file that contains the source worktree path

## Validation

- `.venv/bin/ruff check .`
- `.venv/bin/mypy src tests`

Follow-up to #196. Related to #42.

### DIFF
--- a/src/spotter/replay.py
+++ b/src/spotter/replay.py
@@ -576,7 +576,11 @@ def _declared_resource_drift(
         return (EnvironmentDrift.UNKNOWN_ENVIRONMENT_DRIFT,)
     drift: list[EnvironmentDrift] = []
     for source, restored in zip(left, right, strict=True):
-        if source.worktree_path_reference != restored.worktree_path_reference:
+        if source.state == "ignored" and restored.state == "missing":
+            category = EnvironmentDrift.MISSING_IGNORED_FILE
+        elif source.state == "untracked" and restored.state == "missing":
+            category = EnvironmentDrift.MISSING_UNTRACKED_FILE
+        elif source.worktree_path_reference != restored.worktree_path_reference:
             category = EnvironmentDrift.ABSOLUTE_PATH_MISMATCH
         elif (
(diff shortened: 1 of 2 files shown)


YoungJinJung:
## Review

**Summary:** Follow-up to the precedence note from the #196 review: missing ignored/untracked resource classification now runs before the worktree-path-reference comparison, so a declared resource that vanishes after restore is labeled `MISSING_IGNORED_FILE`/`MISSING_UNTRACKED_FILE` — the more diagnostic category — rather than `ABSOLUTE_PATH_MISMATCH`. Resources present on both sides with a stale worktree reference still classify as `ABSOLUTE_PATH_MISMATCH`, including the hash-equal copied-value case #196 exists to catch (the boolean check remains ahead of the sha256 short-circuit).

**Verified:**
- The reorder is safe for every other branch: a missing restored resource has `sha256 = None`, so the sha-equality short-circuit never applied to it anyway — no previously-passing case can newly drift, and no previously-drifting case can newly pass. Only the label changes, in the direction of better diagnosis.
- The regression test is the exact adversarial setup: an ignored `.env` whose content is the source worktree's absolute path (asserted `worktree_path_reference is True` on the source fingerprint), lost after restore — and the preflight reports `MISSING_IGNORED_FILE`, proving the missing category wins.

**Verified locally** in a clean worktree: pytest (572), mypy (79 files), ruff all pass. PR CI fully green.

No blockers. Merging.

PR #198, opened 2026-08-14
docs(experiment): report validation v2 replay noise

YoungJinJung:
## Why

Issue #42 still lacks representative natural-failure noise evidence. The previously held-out validation-v2 corpus provided three more mechanically scored task prefixes for pinned-low replay qualification.

## What changed

- record a six-arm replay-source capture run and coverage for all six retained sessions
- report one exact-prompt neutral pair from each task control source at its earliest pre-mutation exact prefix
- add bounded machine-readable provenance, raw SHA-256 references, continuation diagnostics, and aggregate outcomes
- update fidelity-v1 coverage from 5/27/17 to 11/59/40 sessions/candidates/exact and add the passing validation-v2 stratum
- retain the NO-GO decision because all sources and continuations passed, so natural failure-region noise remains unobserved

The source batch guidance suffix repeated the neutral control sentence but was not byte-identical to the control prompt. It is used only for capture/coverage; neutral outcome counts come exclusively from `--neutral` exact-prompt pairs.

## Validation

- frozen corpus validate/preflight: 3/3 READY
- source batch: 6/6 PASS, replay sources 6/6 captured
- neutral replay: 3/3 judgeable, 6/6 PASS, 0 disagreements, 0 environment mismatches, 0 infrastructure failures
- raw batch/result SHA-256 and derived classifications cross-checked
- `jq empty` on both JSON artifacts
- `.venv/bin/ruff check .`
- `.venv/bin/mypy src tests`

Related to #42.


YoungJinJung:
## Review

**Summary:** Runs the previously held-out `validation-v2` corpus as a pinned-low capture batch (6/6 arms passed, 6/6 replay sources captured), measures coverage for all six retained sessions, and replays one exact-prompt neutral pair from each task's earliest pre-mutation control prefix: 3/3 judgeable, 6/6 passing, 0 disagreements, 0 environment mismatches, 0 infrastructure failures. A fourth stratum (`pinned-low-effort-validation-v2-pass`) lands in fidelity v1; NO-GO is retained.

**Verified — the arithmetic is consistent at every level:**
- Fresh coverage 5→11 sessions (+6), 27→59 candidates (+32), 17→40 exact (+23), 11→25 (+14) and 9→22 (+13) pre-mutation — each delta exactly matches the reported `source_coverage` block.
- The per-arm coverage table sums check out: 32 candidates, 23 exact, 13/14 pre-mutation across the six rows; the three control-arm rows match the per-run coverage blocks in the JSON verbatim.
- New stratum bound 0.6316 = 1 − 0.05^(1/3), consistent with the other 0/3 strata.

**What I like:**
- The exact-prompt discipline: the source batch's guidance suffix repeated the control sentence but wasn't byte-identical, so those pairs are explicitly excluded from neutral evidence (`exact_prompt_pairs: false`) and used only for capture/coverage. That's the right line to hold — "almost the same prompt" is not a neutral pair.
- This is the first neutral evidence with one-pair-per-prefix breadth across three distinct task prefixes rather than repeat precision on one prefix, and the report frames it exactly that way instead of overclaiming.
- Honest negative-evidence accounting continues: all sources passed, so the run contributes nothing to failure-region calibration, and the blocking gaps say so ("added three passing prefixes but no naturally occurring failure", "repeated task families").
- Excluded later proposals (post-external-effect) are named as exclusions rather than silently absorbed into coverage.
- Provenance: batch and all three neutral raw JSONLs pinned by SHA-256, with a stated reason for not committing raw JSONL.

**Verified locally** in a clean worktree: pytest (572), mypy (79 files), ruff pass; jq validation of the new stratum and the run JSON passes. PR CI fully green.

No blockers. Merging.

PR #199, opened 2026-08-14
feat(replay): classify missing venv/cache drift

YoungJinJung:
## Summary

- add an explicit `--environment-venv-or-cache` task contract for fork and experiment runs
- persist declared resource purpose in fork manifest schema v6 and retain v1-v5 compatibility
- classify source-to-fork loss as `MISSING_VENV_OR_CACHE` while preserving generic ignored/untracked classifications
- retain the declaration through experiment provenance and the immediate pre-continuation preflight

Part of #42.

## Verification

- `ruff format --check .`
- `ruff check .`
- `mypy src tests`
- `pytest` (574 passed)


YoungJinJung:
## Review

**Summary:** Adds an explicit `--environment-venv-or-cache` declaration for fork/experiment runs. Declared paths carry a `purpose` field ("venv_or_cache" vs "resource") persisted in fork manifest schema v6 (v1–v5 stay readable, defaulting to "resource"), and a declared venv/cache missing after restore classifies as `MISSING_VENV_OR_CACHE` instead of the generic missing-ignored/untracked categories.

**What I like:**
- The no-guessing discipline is preserved where it would have been tempting to cheat: purpose comes only from explicit declaration — Spotter does not infer "this is a venv" from a path named `.venv`, and the lifecycle doc states that outright.
- The classification is a strict refinement, not a relaxation: every case that now reports `MISSING_VENV_OR_CACHE` previously reported some other blocking drift (missing-ignored/untracked or unknown), so nothing that used to block can now pass — the label just names the #42 category directly.
- The identity check in `_declared_resource_drift` now compares `(path, purpose)` pairs, so a purpose re-declaration between source and recheck surfaces as `UNKNOWN_ENVIRONMENT_DRIFT` rather than being silently coerced.
- The pre-continuation recheck reconstructs the declaration split from the immutable manifest by purpose (filtering venv paths out of the generic resource list), so the immediately-before-continuation preflight verifies exactly what was declared at fork time — and the recheck test asserts the split reaches `fingerprint_environment` correctly.
- Merging the two declaration lists as a sorted set keeps fingerprint ordering deterministic and lets a doubly-declared path resolve to the more specific purpose.
- Test coverage hits every layer: end-to-end fork preflight (`SOURCE_ENVIRONMENT_MISMATCH:MISSING_VENV_OR_CACHE` for a gitignored `.venv` lost in restore), manifest purpose persistence, v5 backward readability (field stripped and reloaded), experiment provenance JSONL, per-fork pass-through, and CLI plumbing.

**Verified locally** in a clean worktree: pytest (574), mypy (79 files), ruff all pass. PR CI fully green. README/lifecycle/status updated with the new contract and schema version.

No blockers. Merging.

PR #200, opened 2026-08-14
feat(metrics): record counterfactual arm costs

YoungJinJung:
## Summary

- retain parsed agent token totals and same-process monotonic agent duration in counterfactual experiment schema v3 rows
- project those fields through the existing stable run/pair/arm objective-cost join while keeping v1/v2 results readable
- avoid persisting raw agent stderr solely for cost extraction

This closes a missing experiment/scorer cost-coverage path from #33; control delivery/adoption telemetry remains separate follow-up work.

## Verification

- `ruff format --check .`
- `ruff check .`
- `mypy src tests`
- `pytest` (576 passed)


YoungJinJung:
## Review

Reviewed the full diff (experiment.py, runtime_metrics.py, docs, tests).

- `TimeoutExpired.stderr` is raw bytes even with `text=True` on POSIX; `_reported_tokens` handles the bytes branch correctly, so token recovery from a timed-out arm works.
- `_agent_elapsed` falls back to the v1/v2 `started_at`/`ended_at` wall-clock path only when `agent_elapsed_ms` is absent or negative — older rows stay readable, and the schema gate now accepts {1, 2, 3} while still failing visibly on unknown newer schemas.
- `monotonic_ns` start is taken after the environment preflight, so INFRA_FAIL preflight rows correctly carry no duration.
- Rows persist only the parsed token total, not raw agent stderr, matching the stated goal; the test asserts `agent_stderr` is absent.

**Minor (non-blocking):**
- `_TOKENS_USED_RE` is now duplicated between `experiment.py` and `runtime_metrics.py`. Fine for now since the metrics side must keep parsing legacy `agent_stderr` rows independently, but worth consolidating if the format ever changes.
- The 8-element execution tuple in `_execute_arm`/`run_experiment` is getting unwieldy; a small frozen dataclass would improve readability next time the shape grows.

Checks green (CI + CodeQL), 576 tests passing per the PR description, docs updated with the contract change. LGTM — merging.

PR #201, opened 2026-08-14
feat(metrics): measure supervision lead and lag

YoungJinJung:
## Why

Issue #33 requires supervision lead/lag to be reportable without manufacturing durations across process clocks. Existing metrics stopped at detection-to-decision latency and could not show whether a reviewer decision was ready before the target Main turn boundary.

## What changed

- join signal-driven reviewer decisions to the matching thread, turn, and connection-epoch completion boundary
- report same-clock monotonic decision lead and lag with explicit coverage denominators
- leave missing identity, missing boundaries, and cross-restart clock domains unknown
- document the expanded runtime-metrics contract and current project status

This advances #33. Control delivery/adoption telemetry remains a separate follow-up after the live delivery path exists.

## Validation

- `ruff format --check .`
- `ruff check .`
- `mypy src tests`
- `pytest` (577 passed)


YoungJinJung:
## Review

Reviewed the full diff (runtime_metrics.py, docs, tests).

- The `_monotonic_elapsed_ms` → `_monotonic_delta_ms` split is the right refactor: signed deltas are exactly what lead/lag needs, and the elapsed wrapper still rejects negative durations, so all existing callers keep their previous behavior.
- Lead/lag sign convention checks out: decision-before-boundary yields a positive delta (lead), decision-after yields negative (lag), and both report `abs(delta)` — matches the test expectations (200ms lead / 300ms lag).
- Clock-domain safety is preserved: `_monotonic_delta_ms` returns None on mismatched `monotonic_clock_id`, so the cross-restart case stays uncovered instead of manufacturing a duration, consistent with the architecture doc's contract. The restart test pins this.
- Boundary map is fully built during the record scan and lead/lag is computed after it, so a decision record appearing before its `turn_completed` record in journal order still joins correctly.
- `setdefault` keeps the first completion per (thread, turn, epoch) as the boundary — sensible and deterministic.

**Minor (non-blocking):**
- `delta >= 0` counts an exact tie as lead. Fine either way, but worth a comment if the tie semantics ever matter.
- Lead and lag samples both render against the `jobs_decided` denominator, so the two fractions overlap rather than partition. That's consistent with the repo's explicit-coverage style, just noting it's intentional.

Checks green (CI + CodeQL), 577 tests passing, architecture/roadmap/status docs updated with the contract change. LGTM — merging.

PR #202, opened 2026-08-14
feat(metrics): report paired arm costs

YoungJinJung:
## Why

Issue #33 requires experiment costs to appear next to paired benefit/harm outcomes. The objective report joined costs to durable arm rows but only rendered one aggregate total, hiding control-versus-guidance and neutral-arm cost differences.

## What changed

- form complete guidance and neutral pairs using their stable run/pair identity before aggregating costs
- report token and elapsed coverage separately for each paired arm role beside its outcome summary
- keep orphan arms in overall coverage without pairing them by role alone
- document the expanded objective cost-reporting contract and current project status

This advances #33. Control delivery/adoption telemetry remains a separate follow-up after the live delivery path exists.

## Validation

- `ruff format --check .`
- `ruff check .`
- `mypy src tests`
- `pytest` (578 passed)


YoungJinJung:
## Review

Reviewed the full diff (runtime_metrics.py, docs, tests).

- Per-role cost aggregation only draws from complete pairs (`{control, guidance}` or `{neutral_a, neutral_b}` by stable pair identity), so a stray control row from pair 0 and a guidance row from pair 1 never get compared as if they were a pair — and the new cross-pair-id test pins exactly that.
- Orphan arms still count toward the overall `reported_tokens`/`token_arms`/`elapsed` coverage while being excluded from role aggregation, matching the stated contract ("orphan arms are never paired by role alone").
- Rendering is coverage-aware: token and elapsed denominators use the paired-arm count per role, and the whole line is omitted when no pairs of that type exist, so empty reports don't print misleading zero rows.
- `ObjectiveArmCost` is frozen and the report's new field is a plain mapping built in one place; no other construction sites of `ObjectiveOutcomeReport` to update.

**Minor (non-blocking):**
- In `_append_paired_costs`, the `if cost is None: continue` branch is effectively dead since complete pairs always populate both roles together; harmless as defensive coding.

Docs (architecture/roadmap/status) updated with the contract change, 578 tests passing per the description. CI is still running — will merge once checks are green.

PR #205, opened 2026-08-15
feat(metrics): measure rater agreement

YoungJinJung:
## Why

Issue #38 requires double-labeling a subset of judgments and reporting rater agreement. Existing labels discarded rater identity at write time, so independent judgments and same-rater corrections were indistinguishable.

## What changed

- record a stable rater identity in label schema v2, with CLI `--rater` override and OS-account default
- retain v0/v1 readability and expose unattributed legacy labels in agreement coverage
- project the latest judgment per rater, treating same-rater corrections as one judgment
- report exact agreement only after the existing minimum sample threshold, with double-label, stale, and unattributed coverage
- document the new labeling and agreement contract

This advances #38. Sampling and labeling unflagged cases for miss-rate estimates remains follow-up work.

## Validation

- `ruff format --check .`
- `ruff check .`
- `mypy src tests`
- `pytest` (594 passed)


YoungJinJung:
## Review

Reviewed the full diff (labels.py, metrics.py, cli.py, docs, tests).

- Latest-per-(step, rater) projection is the right semantics: same-rater corrections collapse to one judgment (append order wins), while independent raters stay separate — and the test pins exactly that three-row alice/alice/bob case.
- Legacy v0/v1 rows deserialize with `rater=""` and are excluded from the agreement verdict list but still surface as `unattributed_labels` coverage, so old data can never silently inflate or deflate agreement. All unattributed rows per step collapse to one `(step, "")` key, which is fine since they're never compared anyway.
- Staleness is checked on the rater's *latest* judgment against the current fingerprint — a stale correction counts as stale rather than reviving the rater's earlier matching verdict, consistent with the append-only correction contract in `load_labels`.
- Exact agreement over 3+ raters requires unanimity (`len(set) == 1`), and the MIN_SAMPLES gate keeps the report honest below 5 double-labeled targets.
- `Label` field order (`rater` before `version`, defaulted) keeps positional construction and v0/v1 JSON readability intact; the newer-schema refusal test now tracks the label schema constant instead of the journal one — good catch.

**Minor (non-blocking):**
- `getuser()` can raise `OSError` on daemonized/no-login environments; if labeling ever runs non-interactively, wrapping the default in a LabelError would fail cleaner. Not a concern for the current CLI-only path.
- `load_label_history` is now read twice per session in `_metrics_main` (via `tally_session` → `load_labels` and via `agreement_session`); negligible for label-file sizes.

Checks green (CI + CodeQL), 594 tests passing, architecture/roadmap/status docs updated. LGTM — merging.

PR #206, opened 2026-08-15
feat(metrics): measure gate misses

YoungJinJung:
## Why

Issue #38 cannot estimate detector misses by labeling positives alone. Gate precision was measurable, but proposals that received no gate flag had no label vocabulary or coverage denominator.

## What changed

- allow correlation-proven unflagged `tool_proposal` rows to be labeled `miss`, `tn`, or `unclear`
- reject labels when a proposal has a correlated flag or lacks the identity needed to prove silence
- report coverage-aware gate miss rate while surfacing uncorrelatable proposals as blind spots
- preserve proposal labels as exact replay branch points
- bump the backward-readable label contract to schema v3 and document the measurement boundary

This advances #38. Negative sampling for candidate and reviewer silence remains follow-up work.

## Validation

- `ruff format --check .`
- `ruff check .`
- `mypy src tests`
- `pytest` (596 passed)


YoungJinJung:
## Review

Reviewed the full diff (labels.py, metrics.py, cli.py, replay.py, docs, tests).

- The eligibility gate is strict in the right way: `miss|tn` labels are only accepted when a `tool_use_id` exists AND no correlated gate event shares it — proposals without a correlation id are rejected at label time and surfaced as blind spots in the report instead of silently entering the denominator.
- Moving the CONTINUE check under an explicit `reviewer_decision` kind guard is a small correctness improvement over the old shape, which would have rejected any labelable target whose payload happened to carry `decision: continue`.
- Verdict-vocabulary validation now happens after target resolution, so each target kind gets its own vocabulary (`tp|fp` on flags, `miss|tn` on proven-unflagged proposals) and the cross-vocabulary test pins the rejection.
- No cross-contamination between tallies: `Tally.plus` maps `miss`→positive / `tn`→negative, but `tally_session` only visits gate/reviewer records and `tally_unflagged_proposals` only visits eligible proposals, so the shared vocabulary extension can't leak between P3 rates.
- Coverage semantics are right: eligible-but-unlabeled proposals count in `total` (7 total / 5 labeled in the test), keeping the miss-rate honest about coverage.
- Replay branch points for proposal labels map to the proposal's own step — the natural exact branch point — and the updated replay test covers forkability.

**Minor (non-blocking):**
- `tally_unflagged_proposals` calls `unflagged_proposal_eligibility` per proposal, each scanning all records — O(proposals × records) per session. Fine at current journal sizes; precomputing the flagged `tool_use_id` set would make it linear if journals grow.

596 tests passing, docs updated with the v3 label contract. CI still running — will merge once checks are green.

PR #207, opened 2026-08-15
feat(metrics): measure signal precision

YoungJinJung:
## Why

Issue #38 needs coverage-aware precision evidence for emitted detector candidates, stratified by signal type, before active intervention can be justified.

## What changed

- allow `tp|fp|unclear` labels only on canonical active signal candidates with stable identity and type
- report deduplicated candidate precision and coverage by signal type, with malformed candidates shown as blind spots
- map signal labels to their next proposal for replay branch coverage
- advance the label schema to v4 with v0-v3 read compatibility and update implementation-status docs

Negative candidate and reviewer-silence sampling remains follow-up work in #38; this PR does not claim detector recall.

## Validation

- `ruff format --check .`
- `ruff check .`
- `mypy src tests`
- `pytest` (598 passed)


YoungJinJung:
## Review

Reviewed the full diff (labels.py, metrics.py, cli.py, replay.py, docs, tests).

- Label eligibility for signal candidates is properly canonical: only the *first* active occurrence of a `signal_id` accepts a label (the `records[:step]` scan rejects repeats), which lines up exactly with `tally_signal_candidates` deduplicating on first-active-occurrence — a label can never land on a row the tally would skip.
- The three rejection paths (non-active status, missing `signal_id`, missing `signal_type`) are each pinned by tests, and malformed active candidates surface as blind-spot counts instead of silently entering any denominator, consistent with the repo's coverage-honest style.
- Stratification by `signal_type` with per-type `Tally` merge across sessions is straightforward and the mixed-type test verifies both strata and the unattributed count.
- Replay branch mapping treats a signal label like a reviewer decision — next proposal after the labeled step — which matches the signal-trigger-followup semantics already in `branch_coverage`, and the updated replay test covers both the forkable and tail (no-following-proposal) cases.
- Schema bump to v4 with v0–v3 readability follows the established pattern; no loader changes were needed since `rater` was the last field addition.

**Minor (non-blocking):**
- Dedup in the tally keys on `signal_id` alone; if a duplicated id ever arrived with a different `signal_type`, the first occurrence's type silently wins. Not reachable with well-formed emitters, just noting the tie-break.

Checks green (CI + CodeQL), 598 tests passing, docs updated. LGTM — merging.

PR #208, opened 2026-08-15
feat(metrics): measure reviewer misses

YoungJinJung:
## Why

Issue #38 requires emitted and non-emitted judgments to be measured separately. Reviewer `CONTINUE` decisions are explicit negative outputs, but Spotter previously rejected labels for them and could only report intervention precision.

## What changed

- allow `miss|tn|unclear` labels on explicit reviewer `CONTINUE` decisions
- report their miss rate and label coverage separately from VERIFY/NUDGE precision
- state that trajectories with no reviewer decision remain outside this sampling frame
- advance the label schema to v5 while retaining v0-v4 reads
- update current-state and roadmap documentation

Candidate-silence sampling remains follow-up work in #38; this PR does not claim full detector recall.

## Validation

- `ruff format --check .`
- `ruff check .`
- `mypy src tests`
- `pytest` (599 passed)


YoungJinJung:
## Review

Reviewed the full diff (labels.py, metrics.py, cli.py, docs, tests).

- Flipping CONTINUE from "rejected at label time" to "labelable with the negative vocabulary" is clean: the elif chain gives each target kind exactly one vocabulary, and the updated test confirms `tp` on a CONTINUE is now a vocabulary error rather than a target error.
- No double counting: `tally_session`'s reviewer precision still skips CONTINUE rows entirely (pinned by `reviewer_precision.total == 1` against 7 continues + 1 nudge), so a `miss` on a CONTINUE can never leak into the P4 precision rate — the two rates partition reviewer decisions by decision type.
- Coverage semantics match the other negative frames: every CONTINUE counts in `total`, labels fill `labeled`, and `miss`→positive / `tn`→negative reuses the established `Tally` mapping.
- The printed sampling-boundary caveat (trajectories with no reviewer decision are outside the rate) is a good honesty guard — this measures abstention quality, not detector recall, and the PR description says the same.
- No legacy-label migration risk: labels on CONTINUE steps were previously impossible, so v5 introduces no reinterpretation of existing rows. Schema bump follows the established backward-read pattern.
- Since labels on reviewer decisions already map to the next proposal for replay branch coverage, a labeled CONTINUE miss naturally becomes a forkable intervention opportunity without further replay changes.

No blocking or non-blocking findings beyond that — 599 tests passing, docs updated. Will merge once the remaining CI jobs finish.

PR #209, opened 2026-08-15
feat(metrics): sample signal misses

YoungJinJung:
## Why

Issue #38 requires detector negatives to be sampled and reported with their denominator and selection bias. Attaching ad hoc miss labels directly to arbitrary journal steps would lose the sampling frame and could make a sparse detector look better than it is.

## What changed

- add `spotter sample-signals` to persist deterministic Bernoulli samples over explicitly declared event-kind strata
- store disjoint journal-suffix frames with inclusion probability and emitted, suppressed, and unobservable exclusions
- add schema-v6 scoped labels so signal miss judgments can coexist with gate judgments on the same step
- require `miss|tn|unclear` plus written criteria for sampled detector-negative labels
- report miss rate and coverage per detector/event-kind/probability stratum, including blind-spot and bias boundaries
- mark late signal emissions and changed targets stale instead of counting them as misses
- include scoped misses in rater agreement and replay branch-opportunity coverage
- document the new command and persisted sampling schema

Mechanically scored failure seeding and collection of representative double-labeled evidence remain experiment work in #38; this PR adds the honest instrument and does not claim recall results.

## Validation

- `ruff format --check .`
- `ruff check .`
- `mypy src tests`
- `pytest` (609 passed)


YoungJinJung:
## Review

Reviewed the full diff (new sampling.py, labels.py, metrics.py, cli.py, replay.py, docs, and both test files).

- The deterministic Bernoulli draw hashes (session, type, kinds, rate) with the event id and is step-independent, so re-sampling after journal growth selects consistently, and the suffix-only frame (`step > previous_end`) keeps repeated invocations disjoint — the idempotency and suffix tests pin both properties.
- Exclusion sets (emitted/suppressed sources) are computed over the whole journal while the frame is suffix-only, and late emissions are handled at the right layer: `signal_source_is_silent` re-checks at label and tally time, so a source that fires after being sampled goes stale instead of polluting the miss rate. The late-emission test verifies total == stale == 1 with labeled == 0.
- Scoped labels are safely partitioned: `load_labels` defaults to scope "" so every pre-existing call site keeps its exact behavior, latest-wins is per (step, scope), and the v6 loader forces scope to "" for older rows — a forged scope can't ride in on a legacy version number. The schema test covers the v1-row read.
- Agreement keying on (step, scope, rater) lets a gate `tn` and a signal `miss` coexist on the same step as independent measurement targets, which the coexistence test demonstrates end to end.
- Frame-file validation on load is thorough — duplicate batch ids, orphan samples, batch/sample disagreement, selected-count mismatch, and newer-schema refusal all raise instead of degrading silently, consistent with the durable-state rules.
- The rendered report keeps the honesty guarantees: per-stratum denominators include unlabeled samples, unobservable exclusions print as a blind-spot fraction, and the explicit bias line stops the rate from being read as trajectory-wide recall.

**Minor (non-blocking):**
- Overlap conflict detection compares `inclusion_probability` with float equality; fine since values round-trip through JSON doubles, just worth remembering if rates ever come from arithmetic instead of CLI input.
- `_metrics_main` is accumulating a lot of parallel per-session aggregates; a small session-summary dataclass would tame the signature creep next time one is added.

Checks green (CI + CodeQL), 609 tests passing, README/user-guide/lifecycle/architecture docs all updated. LGTM — merging.

PR #210, opened 2026-08-15
feat(detection): attribute reviewer launch triggers

### LINKED ISSUE #28: Replace periodic review with event-driven signals

YoungJinJung:
## Why

Issue #28 needs periodic-only, event-driven, and mixed cohorts to remain distinguishable when #38/#24 compare reviewer cost, precision, misses, and delay. Reviewer decisions previously lost their launch source, so those cohorts could be silently pooled.

## What changed

- persist `signal` or `periodic` trigger provenance on queued jobs and propagate it through reviewer terminal records; manual reviews are classified explicitly
- infer legacy signal/periodic provenance only from durable queue identity and otherwise report `unknown`
- stratify reviewer-call counts, intervention precision, and CONTINUE miss labels by launch trigger in `spotter metrics`
- update architecture, status, example configuration, and user guides

## Validation

- `ruff format --check .`
- `ruff check .`
- `mypy src tests`
- `pytest` (610 passed)

Closes #28


YoungJinJung:
## Review

Reviewed the full diff (cli.py, hook.py, metrics.py, runtime_metrics.py, review_executor.py, review_scheduler.py, docs, config example, tests).

- Every write path stamps its provenance at the source: the hook's periodic queue writes `periodic`, the signal scheduler's queued event and the executor's terminal records write `signal`, and the CLI review path resolves its trigger from the durable queued event by job id before falling back. Decision, error, and inference-started events all carry it, so the trigger survives any terminal outcome.
- Legacy classification is conservative in the right order: explicit `review_trigger` > durable `signal_id` on the queued event > `proposal:` job-id prefix > `unknown`, and no-job-id decisions classify as `manual`. Old journals never silently pool into a cohort they can't prove.
- The stratified tallies reuse the existing intervention/CONTINUE partition (verify/nudge vs continue), so per-trigger rows are strict refinements of the P4 and negative-decision totals — no decision can be counted under two triggers or two partitions.
- The runtime-cost report's `triggers=` counts use the same resolution logic against the same queued-event join, and the updated test pins both the count and the rendered line (plus the now-correct `decided=1` since the decision gained its job-id join).
- Docs in all three languages, the example config, and status/architecture were updated together with the contract — consistent with the repo's change discipline.

**Minor (non-blocking):**
- `_review_job_triggers`/`_review_trigger` are duplicated between `metrics.py` and `runtime_metrics.py` with near-identical logic (differing only in payload-id extraction). If the classification rule ever changes, both copies must move together — worth extracting to a shared helper next time either is touched.

Checks green (CI + CodeQL), 610 tests passing. LGTM — merging.

Issue #24, opened 2026-08-11
Measure detection delay and wasted actions before intervention


YoungJinJung:
### Question / hypothesis

How late does Spotter notice a trajectory problem after the earliest point where intervention was both warranted and observable? A detector that eventually fires may still be too late to save meaningful work.

### Experiment design

On labeled trajectories, annotate a defensible **interval** where intervention became warranted rather than forcing a single exact first-error step. Compare that interval with the detector/reviewer decision and, once live intervention exists, actual delivery/adoption.

Record:

- step/turn delay
- wall-clock delay where timestamps are available
- tool/actions taken after the intervention window opened
- cases where the detector never fired

Use the App Server timeline after the Observe stage; older Hook-era records without trustworthy time remain step-only observations.

### Implementation references — study, do not depend on

OpenClaw has exposed several concrete `turn/steer` timing failures that show why Spotter must measure more than reviewer inference latency:

- steering serialized behind a session queue arrived tens of seconds late: https://github.com/openclaw/openclaw/issues/48003
- a steer can be accepted/persisted but never consumed if the target turn ends in the enqueue/consume window: https://github.com/openclaw/openclaw/issues/92433
- a late steer near terminal output can perturb final answer selection rather than help the trajectory: https://github.com/openclaw/openclaw/issues/112468
- RPC acceptance is not equivalent to durable/idempotent adoption: https://github.com/openclaw/openclaw/issues/103920

These are reference failure cases only. Spotter keeps its own Python control path.

### Timing model

Separate at least these points:

```text
T0  earliest warranted interval opens
T1  required evidence first observable in raw source
T2  evidence available in Trace IR / ThreadState
T3  candidate signal emitted
T4  reviewer job queued
T5  reviewer inference starts
T6  reviewer decision ready
T7  control dispatch begins
T8  turn/steer RPC accepted, if applicable
T9  intervention input observed in target turn, if observable
T10 next relevant Main reasoning/tool/action boundary
T11 turn completed / terminal answer
```

Do not require every timestamp for every run; record coverage explicitly.

Derived delays:

```text
observation_delay      = T2 - T1
signal_delay           = T3 - T2
queue_delay            = T5 - T4
inference_latency      = T6 - T5
control_dispatch       = T8 - T7
delivery/adoption_lag  = T9 - T8 where observable
end_to_end_delay       = T9 - T0 or best available equivalent
```

And, most importantly for speculative supervision:

```text
[...]

PR #211, opened 2026-08-15
feat(metrics): record intervention opportunity windows

YoungJinJung:
## Why

Issue #24 cannot calculate honest detection delay from a single guessed failure timestamp. It first needs independently attributable semantic and observable warranted intervals pinned to durable Trace IR identity.

## What changed

- add append-only opportunity schema v1 with semantic and observable earliest/latest bounds, required evidence, rater identity, rationale, and schema version
- pin every anchor to stable event ID plus content fingerprint; changed or legacy events without stable identity cannot silently become evidence
- keep independent raters and append-only corrections, with latest-per-opportunity/rater loading
- add `spotter label-opportunity` and document the workflow and durable storage lifecycle

This is the annotation slice of #24. A follow-up will consume these windows to report bounded delay, never-fired opportunities, and post-window work.

## Validation

- `ruff format --check .`
- `ruff check .`
- `mypy src tests`
- `pytest` (615 passed)

Refs #24


YoungJinJung:
## Review

Reviewed the full diff (new opportunities.py, cli.py, docs in three languages, tests).

- Anchors are pinned the right way: every window bound and evidence step stores (step, event id, content fingerprint), `_anchor` refuses steps without a stable event id at write time, and `matches` requires every anchor to still agree — so a drifted or legacy event makes the annotation stale instead of silently relocating the warranted window. The drift test mutates a payload under the same event id and confirms staleness.
- Validation is complete on both sides: write-time checks (id/rater length, non-empty rationale, window ordering, non-empty evidence, in-range steps) and load-time re-checks (strict field typing, session mismatch, window-order reversal, future/pre-v1 schema refusal, unreadable-line raise). Since `OpportunityError` is a `ValueError` subclass not caught by the loader's `(JSONDecodeError, KeyError, TypeError)` tuple, schema-gate errors propagate with their precise message rather than being rewrapped — correct.
- Latest-per-(opportunity, rater) loading mirrors the label store's correction semantics, keeping independent raters separate for future agreement analysis; the three-append test covers correction plus independence.
- CLI validation uses `is None` checks so step 0 passes correctly, and evidence steps deduplicate while preserving order.
- Scope discipline is good: this is purely the annotation instrument — nothing consumes the windows yet, no delay distribution is claimed, and the docs say exactly that.

**Minor (non-blocking):**
- The user guide describes the observable window as "narrower" than the semantic one, but nothing enforces containment (e.g. `observable_earliest >= semantic_earliest`). If the #24 consumer will assume containment, validating it at write time would catch annotation mistakes earlier; if the windows are genuinely independent, a sentence in the doc saying so would prevent the assumption.

Checks green (CI + CodeQL), 615 tests passing, README/architecture/lifecycle/user guides updated. LGTM — merging.

PR #212, opened 2026-08-15
feat(metrics): measure signal opportunity delay

YoungJinJung:
## Why

The opportunity windows added for #24 need a conservative consumer before Spotter can claim detection timing or wasted-work measurements. Any unrelated candidate, observation gap, or still-open trajectory must not turn into a fabricated detection or miss.

## What changed

- link an active signal only when it cites every required evidence event in the opportunity annotation
- classify evidence-linked detection as EARLY, WITHIN_WINDOW, LATE, NEVER, or UNJUDGEABLE
- report step-bound delay and source-wall delay only within one connection epoch
- count identity-deduplicated actions, failed outcomes, and files after the observable window opens, with unattributed observations disclosed
- keep stale annotations, observation gaps, and open windows outside claimed delay/miss distributions
- aggregate and render the report from `spotter metrics`, with updated architecture, research, status, and multilingual user guidance

This is the signal-stage consumer for #24. Reviewer, control-dispatch, and adoption linkage remain follow-up work.

## Validation

- `ruff format --check .`
- `ruff check .`
- `mypy src tests`
- `pytest` (620 passed)

Refs #24


YoungJinJung:
## Review

Reviewed the full diff (new opportunity_metrics.py, cli.py, docs including research.md and the three user guides, tests).

- The link rule is appropriately strict: only an active candidate citing *every* required evidence event id stops the clock, so the unrelated-candidate case in the fixture correctly classifies LATE rather than WITHIN_WINDOW. First-match-in-journal-order gives the earliest qualifying detection.
- The honesty guards all hold: stale annotations leave the distribution entirely (neither NEVER nor UNJUDGEABLE), an observation gap crossing the measured interval yields UNJUDGEABLE with no post-window work recorded, and source-wall delay refuses cross-epoch, non-finite, or negative intervals. Each has a dedicated test.
- Post-window work is bounded to `records[earliest+1 : candidate.step]` with identity-deduplicated actions/failures and unattributed observations disclosed separately — consistent with the repo's coverage-aware style.
- The docs also resolve the nesting question from #211's review: semantic and observable windows are now explicitly documented as not assumed nested, in all three languages plus architecture.md.

**One semantic edge worth confirming (non-blocking):**
- `_window_closed` treats *any* event after `observable_latest` as closing the window, so a mid-journal window in a still-live session with no qualifying candidate counts NEVER even though a late candidate could still fire and flip it to LATE on a later run. The open-window test only covers the window-at-journal-end case. Since metrics are re-derived from the journal each run this self-corrects rather than corrupting anything, but if NEVER is meant to be terminal evidence for #24, tightening closure to a terminal event (or documenting that NEVER is re-derivable) would make the claim airtight.

**Minor (non-blocking):**
- `_sample` and `_operation_identity` overlap with similar helpers in runtime_metrics.py; fine while the formats differ, worth consolidating if they converge.

620 tests passing, docs updated. CI still running — will merge once checks are green.

PR #213, opened 2026-08-15
fix(metrics): require terminal evidence for never

YoungJinJung:
## Why

The review of #212 identified that any later journal event closed an opportunity window. In a still-live turn, ordinary later activity does not prove a signal will never fire; reporting NEVER there would temporarily overstate detector misses.

## What changed

- classify an undetected opportunity as NEVER only when a terminal turn/thread/session event exists at or after the annotated window
- keep still-live windows UNJUDGEABLE even when unrelated actions appear later
- document that NEVER is terminal evidence and live opportunities remain re-derivable

## Validation

- `ruff format --check .`
- `ruff check .`
- `mypy src tests`
- `pytest` (620 passed)

Refs #24

### DIFF
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -742,6 +742,8 @@ post-window actions, failed outcomes, and files. Unrelated candidates never stop
 clock, stale annotations remain outside the distribution, and an open journal cannot become a
 fabricated `NEVER`. An observation gap crossing the measured interval also makes delay and
 post-window work `UNJUDGEABLE` rather than pretending the missing surface was quiet.
+`NEVER` requires a terminal turn/thread/session event at or after the annotated window; unrelated
+later activity in a still-live trajectory keeps the opportunity `UNJUDGEABLE` and re-derivable.
 
 ## 8.3 Snapshot state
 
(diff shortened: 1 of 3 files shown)


YoungJinJung:
## Review

Reviewed the full diff (opportunity_metrics.py, architecture.md, tests).

This resolves the semantic edge flagged in the #212 review exactly as suggested: `_window_closed` now demands a terminal turn/thread/session event at or after the annotated window, so ordinary later activity in a live trajectory keeps an undetected opportunity UNJUDGEABLE instead of prematurely reporting NEVER.

**Verified:**
- The strengthened open-window test is the right regression: it appends a non-terminal `tool_proposal` after the window and asserts the classification stays UNJUDGEABLE — precisely the case the old `latest + 1 < len(records)` shortcut got wrong.
- The existing NEVER path still holds because the shared fixture ends in `turn_completed`, which is terminal, and `records[latest:]` including the window's own latest step matches the prior boundary semantics.
- Treating `turn_completed` as terminal is consistent with interventions being turn-scoped, and the architecture doc now states both the terminal-evidence requirement and that live opportunities remain re-derivable.

Minimal, well-scoped, checks green (CI + CodeQL), 620 tests passing. LGTM — merging.

PR #214, opened 2026-08-15
feat(metrics): measure opportunity reviewer delay

YoungJinJung:
## Why

Issue #24 needs opportunity-relative timing beyond signal emission so we can tell whether the semantic reviewer finished inside the useful intervention window.

## What changed

- link an evidence-qualified signal candidate to its signal-driven reviewer job through durable candidate event IDs
- follow the reviewer job through inference start and decision by review job ID
- report candidate-to-queue, queue-to-inference, inference-to-decision, queue-to-decision, and decision-vs-window step timing
- keep stale decisions, terminal trajectories without a decision, and observation gaps explicit
- document the completed reviewer-stage slice and remaining control/adoption work

## Validation

- `ruff format --check .`
- `ruff check .`
- `mypy src tests`
- `pytest` (621 passed)

Refs #24


YoungJinJung:
## Review

- The reviewer join is a proper durable-identity chain: candidate event id → `review_job_queued` (via `candidate_event_id` or membership in `candidate_event_ids`) → inference/decision by `review_job_id`, all scanned strictly after the previous stage's step. No timestamp or adjacency guessing anywhere.
- Every linked candidate lands in exactly one reviewer bucket — the queued-missing, gap-crossed, decision-missing, stale, and decided paths each `continue` before the next check, so EARLY/WITHIN/LATE/TERMINAL_NO_DECISION/UNJUDGEABLE/STALE partition cleanly.
- Coverage honesty carries through: a gap-crossed job still counts in `queued`/`started` but contributes no step samples, so the rendered `(n/denominator)` fractions expose the shortfall rather than hiding it; stale decisions are excluded from the delay distribution rather than counted as ordinary latency.
- Denominators are chosen per stage (linked for signal-to-queue, queued for queue-to-*, started for inference-to-decision), and the `inference.step <= decision.step` guard keeps a reordered journal from producing a negative interval.
- The decision-missing case reuses the #213-hardened `_window_closed` (terminal evidence at/after the queue step), so TERMINAL_NO_DECISION inherits the same no-fabrication guarantee.

**Minor (non-blocking):**
- The no-queued-job gap check scans the full tail (`records[candidate.step:]`) while the queued path bounds its gap check to the measured interval — slightly asymmetric, though the asymmetry errs conservative (more UNJUDGEABLE, never fabricated latency), so it's fine as-is.

621 tests passing, docs updated across architecture/research/status and all three user guides. CI still running — will merge once checks are green.

PR #215, opened 2026-08-15
feat(metrics): measure opportunity control delay

YoungJinJung:
## Why

Issue #24 needs to separate reviewer lateness from control-plane lateness. Opportunity timing now reaches reviewer decisions, but it cannot yet show whether a warranted intervention was dispatched and resolved before the useful window closed.

## What changed

- follow non-stale `VERIFY`/`NUDGE` decisions into their first control attempt by durable review job ID
- follow dispatch to RPC acceptance or terminal outcome by durable control ID
- classify dispatch against the observable intervention window
- report decision-to-dispatch and dispatch-to-resolution step coverage
- keep stale, failed, unknown, terminal-without-dispatch/resolution, and observation-gap cases explicit
- document observed adoption as the remaining #24 linkage slice

## Validation

- `ruff format --check .`
- `ruff check .`
- `mypy src tests`
- `pytest` (623 passed)

Refs #24


YoungJinJung:
## Review

- The control join continues the same durable-identity discipline: `review_job_id` from a non-stale VERIFY/NUDGE decision to the first control attempt, then `control_id` to RPC acceptance or a terminal outcome, always scanning strictly after the previous stage. No adjacency or timestamp guessing.
- Treating a `control_terminal` found *before* any dispatch as a pre-dispatch outcome (stale/failed/unknown) rather than a missing dispatch is the right call — the stale-before-dispatch test pins that it lands in STALE_BEFORE_DISPATCH, not TERMINAL_NO_DISPATCH.
- The no-fabrication guards carry through the new stage: gap-crossed intervals go UNJUDGEABLE without contributing step samples, missing dispatch/resolution split by the #213-hardened terminal-evidence rule, and a dispatched-then-gapped control still counts in `dispatched` so the `(n/denominator)` fractions expose the shortfall.
- Denominators are per stage again (eligible for decision-to-dispatch, dispatched for the dispatch-relative samples), matching the reviewer-stage conventions from #214.

**Minor (non-blocking):**
- Unlike the reviewer line, the rendered control line mixes two axes: EARLY/WITHIN/LATE classify *dispatch timing* while ACCEPTED/FAILED/STALE_AFTER/TERMINAL_NO_RESOLUTION classify *resolution outcome*, so a single dispatched control appears in one bucket from each axis and the counters don't sum to `eligible`. The numbers are all correct — but a reader pattern-matching from the reviewer line (which is a strict partition) may misread it. A label tweak (e.g. `dispatch_timing: ...; resolution: ...`) would remove the ambiguity.

Checks green (CI + CodeQL), 623 tests passing, docs updated across architecture/research/status and all three user guides. LGTM — merging.

PR #216, opened 2026-08-15
feat(metrics): measure opportunity adoption delay

YoungJinJung:
## Why

Issue #24 needs to distinguish RPC acceptance from guidance actually appearing in the target turn. Acceptance alone is not evidence that Main consumed the intervention.

## What changed

- link accepted steers to observed `user_prompt` adoption by client message ID and exact thread/turn/connection identity
- classify observed adoption against the opportunity window
- report acceptance-to-adoption and decision-to-adoption step coverage
- keep `RPC_ACCEPTED_ONLY`, stale-after-accept, observation-gap, and missing-identity cases explicit
- separate dispatch-timing and resolution-outcome labels in the control report, addressing the non-blocking #215 review note
- update architecture, research status, and user guides

## Validation

- `ruff format --check .`
- `ruff check .`
- `mypy src tests`
- `pytest` (625 passed)

Refs #24


YoungJinJung:
## Review

- Adoption is held to the strictest identity bar in the chain so far: an accepted steer is adoption-eligible only with a `client_user_message_id` and full target identity, and the observed `user_prompt` must match both the message id and the exact (thread, turn, epoch) target key. The wrong-turn test pins the key failure mode — the same message id appearing on another turn classifies RPC_ACCEPTED_ONLY, not adopted.
- The missing-adoption decision order is right: observation gap first (ADOPTION_UNKNOWN), then stale-after-accept via a later `control_terminal` with the same `control_id`, then target-turn completion (RPC_ACCEPTED_ONLY), else UNKNOWN while the turn is still live — acceptance alone can never masquerade as consumption, per the stated goal.
- Scanning for adoption from `dispatch.step + 1` (not acceptance) is the right anchor since dispatch is when the steer was actually sent, and the `adoption.step >= resolution.step` guard keeps the acceptance-to-adoption sample non-negative when the prompt lands between dispatch and the acceptance record.
- The render now splits `Control dispatch timing` / `Control resolution` / `Control adoption` into separate labeled lines — this resolves the axis-mixing readability note from the #215 review exactly, and the updated assertions pin the new format.
- `_target_window_closed` scopes closure to the *target turn's* completion rather than any terminal event, which is the correct boundary for RPC_ACCEPTED_ONLY.

**Minor (non-blocking):**
- `_control_target_key` is a third sibling of the target-key builders in runtime_metrics.py; same consolidation note as before if the identity resolution rule ever changes.

625 tests passing, docs updated across architecture/research/status and all three user guides. CI still running — will merge once checks are green.

PR #217, opened 2026-08-15
feat(intervention): deliver opt-in live advisories

YoungJinJung:
## Why

The runtime already has exact-target `turn/steer`, durable reviewer jobs, freshness fencing, and control/adoption telemetry, but reviewer decisions still stop at the journal. This adds the first conservative live-delivery slice for #22 without changing the safe default.

## What changed

- add a separate `reviewer.deliver_on_signals` opt-in that requires signal reviews and active mode
- deliver only fresh signal-driven `VERIFY`/`NUDGE` decisions; `CONTINUE` and stale results never send
- bind delivery to the immutable review job attachment/turn/connection epoch
- use one stable intervention/control ID and never retry ambiguous acceptance
- send a bounded current-turn advisory that explicitly preserves the original user task
- persist current-turn scope, expiry, non-user-goal intent, and intervention identity in control telemetry
- keep live delivery failures visible and retain shadow mode by default
- update current-boundary docs and configuration examples

## Validation

- `ruff format --check .`
- `ruff check .`
- `mypy src tests`
- `pytest` (628 passed)

Refs #22


YoungJinJung:
## Review

Reviewed the full diff (config.py, daemon.py, review_executor.py, runtime_connection.py, docs, example config, tests). Since this is the first slice that lets a reviewer decision touch a live turn, I read it against the repo's runtime rules specifically.

**Correctness / safety — holds up well:**
- The gate is genuinely double opt-in: `deliver_on_signals` hard-fails config load without `on_signals` AND `observation_only = false`, defaults stay off, and both rejection paths are pinned by tests. Existing configs are untouched.
- The delivery predicate is conservative in the right order — the decision is journaled first, then delivery is skipped for stale results, non-intervention decisions, and disabled mode; the stale-race test proves a decision that went stale mid-inference records `stale: true` and never sends.
- At-most-once semantics come from the stable `spotter:intervention:{job_id}` control id riding the existing durable-uniqueness check, and delivery binds to the immutable job snapshot identity + epoch with no retargeting — a `StaleControlTarget` (missing epoch) surfaces as a journaled `intervention_delivery_error` rather than a silent drop, consistent with fail-visibly.
- Nice coherence with the #216 adoption instrument: the steer's `client_user_message_id` is the control id, and the new `intervention_id`/`supervision_scope`/`must_not_become_user_goal`/`expires_on` fields ride only review-linked steers, so manual steers are unchanged and delivered advisories are immediately measurable by the existing adoption metrics.
- Docs are honest about the boundary — README/status/research all say benefit and task-ownership safety are *not* established, which matches the repo's implementation-vs-evidence discipline.

**Minor (non-blocking):**
- `shadow: not deliver_on_signals` labels the *mode*, not the delivery outcome — under live mode, a stale or CONTINUE decision also records `shadow: false` despite never sending. Anyone filtering `shadow is false` to find delivered interventions would over-count; the durable control telemetry is the correct source. Worth either a doc note or an explicit `delivered` field on the decision event in a follow-up.

Checks green (CI + CodeQL), 628 tests passing. LGTM — merging.

PR #218, opened 2026-08-15
feat(intervention): correlate observed advisories

YoungJinJung:
## Why

Live `turn/steer` inputs are serialized by Codex as user messages. Without Spotter-owned correlation, the live-state reducer can mistake its own advisory for a new authoritative user goal, and RPC acceptance remains indistinguishable from an input actually observed on the target turn.

## What changed

- correlate known intervention `clientUserMessageId` values to their durable control request and exact attachment/turn/epoch
- journal `control_observed_in_turn` when the advisory input appears on its intended target
- journal `control_observed_outside_target` / `expired_advisory_visible` when the same intervention appears outside its scope
- tag correlated inputs as Spotter supervision so ThreadState preserves the original user goal
- cover exact-target adoption, later-target leakage, and task-ownership reduction behavior
- update architecture, lifecycle, status, and user guidance for the live contract

## Validation

- `ruff format --check .`
- `ruff check .`
- `mypy src tests`
- `pytest` (630 passed)

Refs #22

### DIFF
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -1,7 +1,7 @@
 # Architecture
 
 > **Status:** this document describes both the current hook-based prototype and the target architecture. Active implementation is tracked by the [Roadmap](roadmap.md) and native GitHub Milestones.
-> The `spotterd` process/control foundation, bounded Hook gate IPC, shared-server PoC, production App Server transport, durable Trace IR ingestion, daemon reconnect/reconciliation, and stable packaged runtime layout are implemented. Live supervision delivery remains **target behavior**.
+> The `spotterd` process/control foundation, bounded Hook gate IPC, shared-server PoC, production App Server transport, durable Trace IR ingestion, daemon reconnect/reconciliation, stable packaged runtime layout, and off-by-default live advisory delivery are implemented. Broader live enablement remains evidence-gated.
 
 ---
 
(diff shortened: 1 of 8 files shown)


YoungJinJung:
## Review

Reviewed the full diff (runtime_connection.py, thread_state.py, docs, tests).

- The task-ownership guard is enforced at the right layer: the supervision tag is stamped on the event payload *before* live-state reduction, and the reducer routes tagged prompts into `supervision.interventions` while leaving `task.goal` and evidence untouched — both the reducer unit test and the end-to-end WebSocket test pin that the original goal survives an adopted advisory.
- Because the annotation is persisted on the recorded event, restart hydration replays already-tagged prompts, so the reducer's behavior is stable across daemon restarts without re-running correlation logic. The `_control_requests` registry is likewise re-hydrated from durable dispatch records, so a restart between dispatch and observation doesn't lose the correlation.
- Target matching reuses the exact (turn, epoch, attachment-when-present) identity discipline, and the outside-target path is properly diagnostic: relation `outside_target`, a distinct `control_observed_outside_target` event with `expired_advisory_visible`, per-observation event ids (suffixed with the input's event id, vs. the single deterministic id for in-turn), and no state mutation — the leakage test confirms the goal stays intact.
- Registry entries only ever contain Spotter-generated `spotter:intervention:{job_id}` ids from Spotter's own dispatches, so a genuine user message can't be mis-tagged unless the App Server echoes a forged client id — inside the existing transport trust boundary, consistent with how adoption matching already works.
- Docs across architecture/lifecycle/status/user-guide were updated to describe the live contract, including the honest reframing of the status board ("broadly enabled" remains ❌).

**Minor (non-blocking):**
- Registry hydration keeps the first non-stale record per control id but never evicts on a later stale terminal — which is actually the safer behavior (a late-arriving advisory input still gets tagged instead of contaminating the goal), just worth a comment since the `outcome != \"stale\"` filter reads as if stale controls are dropped entirely.

Checks green (CI + CodeQL), 630 tests passing. LGTM — merging.

PR #219, opened 2026-08-15
fix(intervention): fence steers after final answers

YoungJinJung:
## Why

A turn can still appear active briefly after Codex has emitted its terminal assistant answer. Sending a soft intervention in that window is no longer causally useful and can perturb final-output selection.

## What changed

- retain completed `agentMessage.phase=final_answer` as a turn-identity-bound terminal hint in ThreadState
- stale queued/running reviewer jobs as soon as their target final answer settles
- reject later `turn/steer` dispatches locally with `terminal_answer_settled` while keeping interrupt as a separate recovery control
- ignore started, commentary, missing-phase, and other-turn answers for the target fence
- cover protocol normalization, lifecycle reduction, in-flight review races, and steer-vs-interrupt behavior
- document the conservative final-answer boundary

## Validation

- `ruff format --check .`
- `ruff check .`
- `mypy src tests`
- `pytest` (634 passed)

Refs #22


YoungJinJung:
## Review

Reviewed the full diff (thread_state.py, review_scheduler.py, runtime_connection.py, docs, tests).

- The fence triggers only on the strictest available evidence — `phase == final_answer` AND `lifecycle == completed` with non-empty text — and the started/commentary/absent-phase cases are explicitly non-terminal, which is the right posture given inconsistent provider population of the field. The normalizer test pins the phase passthrough.
- The terminal hint is turn-identity-bound end to end: the scheduler stales only jobs whose target turn matches the answer's provenance turn (edge-triggered against `state_before`, and defensively idempotent since `_discard` removes the job), `review_job_is_fresh` adds the same check so an in-flight decision records stale, and `_validate_target` rejects only steers whose target turn owns the settled answer.
- Steer vs. interrupt separation is exactly right — `reject_terminal_answer` is scoped to `control_kind == \"steer\"`, so the recovery control path stays available, and the WebSocket test proves both the local `terminal_answer_settled` rejection (journaled as a stale control_terminal with the precise reason code) and the still-working interrupt.
- `turn_started` clears a previous turn's answer, and the parametrized executor test shows both settle paths (turn_completed → `target_changed`, final answer → `terminal_answer_settled`) produce a stale decision with the correct reason.

**Minor (non-blocking):**
- The semantic reducer sets `terminal_answer` regardless of whether the answer's turn matches the active turn — the thread-state test deliberately keeps a late-arriving turn-1 answer while turn-2 is active. Since the fence compares provenance turn ids this can never fence the wrong turn, but in the contrived ordering where turn-2 had already settled its own answer, a late turn-1 answer would overwrite it and briefly un-fence turn-2. Guarding the setter with `provenance.turn_id == active_turn_id` (or keeping per-turn answers) would close that window if it ever matters in practice.

634 tests passing, docs updated across architecture/lifecycle/status/user-guide. Will merge once the remaining CI jobs are green.

PR #220, opened 2026-08-15
fix(intervention): classify steer rejections

YoungJinJung:
## Why

Codex currently exposes no-active-turn and expected-turn-mismatch steer failures only as human-readable RPC messages, while non-steerable active turns have structured data. Treating all three as generic failures obscures stale targets and tempts supervision policy to parse adapter-specific English.

## What changed

- introduce semantic App Server control errors for no active turn, turn mismatch, and active-turn-not-steerable
- isolate current Codex message compatibility handling inside the App Server adapter
- map no-active/mismatch rejections to durable stale outcomes and non-steerable to failed
- retain unrecognized rejections as generic `rpc_rejected` and transport ambiguity as unknown
- cover adapter mapping and runtime telemetry behavior
- document the compatibility boundary and stable reason codes

## Validation

- `ruff format --check .`
- `ruff check .`
- `mypy src tests`
- `pytest` (635 passed)

Refs #22
Refs openai/codex#31438


YoungJinJung:
## Review

Reviewed the substantive diff (app_server.py, runtime_connection.py, docs, tests — the branch also carries #219's already-merged commit, see the merge note below).

- The adapter-boundary discipline is exactly right: English-message pattern matching (casefold-normalized, tightly anchored prefixes/suffixes) lives only in `app_server.py`, structured `codexErrorInfo.activeTurnNotSteerable` is preferred when present, and the supervision core only ever branches on the semantic `ControlFailureReason` — consistent with the repo rule that policy must not depend on Codex message shapes, with the upstream structured-error ask tracked via openai/codex#31438.
- The outcome mapping is conservative and telemetry-coherent: no-active-turn and turn-mismatch become `stale` (they are staleness evidence, matching the local `terminal_answer_settled`/`stale_target` family), non-steerable stays `failed`, unrecognized rejection text stays generic `rpc_rejected`, and transport loss after dispatch remains `acceptance_unknown` — no rejection is ever upgraded by guesswork.
- `AppServerControlError` subclassing `AppServerRpcError` keeps every existing `except AppServerRpcError` path working when the new handler doesn't match, and the ordering of the except clauses in `_dispatch_control` is correct (subclass first).
- Tests cover all four adapter mappings including the negative case (generic rejection is *not* a control error), plus the end-to-end telemetry triple of stale/failed/unknown with reason codes.

**Merge blocker (process, not code):**
- The PR is currently **CONFLICTING** — it was branched on top of the #219 branch, whose commit was squash-merged to main, so the duplicated first commit now conflicts. Please rebase onto main (dropping the already-merged `163f61b` commit); the actual change (`0e68b9f`) should apply cleanly. I'll merge once the rebase lands and checks pass.

635 tests passing per the description, docs updated across architecture/lifecycle/status/user-guide.

YoungJinJung:
Rebase confirmed — the diff body is byte-identical to the reviewed change, now as a single clean commit on main. All checks green. Merging as promised.

PR #221, opened 2026-08-15
feat(intervention): reconcile accepted-only steers

YoungJinJung:
## Why

A successful `turn/steer` RPC only proves server acceptance. If the target settles before the correlated input appears, Spotter must durably distinguish that from observed-in-turn delivery instead of leaving acceptance unresolved or counting a late persisted input as timely adoption.

## What changed

- retain accepted intervention controls and exact observed-input identities across runtime hydration
- close unobserved accepted steers as `rpc_accepted_only` on completed final-answer or turn boundaries
- reconcile both response-before-boundary and boundary-before-response ordering
- require the target to remain active and pre-terminal when classifying an input as observed in turn
- classify same-turn inputs arriving after completion as expired/outside-target evidence
- cover final-answer and turn-completion terminal paths plus late-input behavior
- document accepted-only semantics and non-adoption claims

## Validation

- `ruff format --check .`
- `ruff check .`
- `mypy src tests`
- `pytest` (637 passed)

Refs #22


YoungJinJung:
## Review

Reviewed the full diff (runtime_connection.py, docs, tests), plus the surrounding control-telemetry machinery on main to verify the reconciliation interactions.

- Both orderings are genuinely covered: boundary-first closes pending acceptances from `_record` (turn_completed / completed final_answer each with their own reason code), and acceptance-after-boundary closes immediately via `_reconcile_settled_acceptance` against `completed_turns`/`terminal_answer` state. The parametrized WebSocket test pins the dispatch → accepted → `rpc_accepted_only` terminal sequence for both boundary kinds with no observed-input event.
- I specifically checked the double-boundary case (final_answer settles, then turn/completed arrives for the same turn): `_reconcile_unobserved_acceptances` would fire twice, but `_record_control_event` derives the event id as `spotter:control:{control_id}:rpc_accepted_only` and `_enqueue_control_event` dedups on that id — so exactly one terminal survives, carrying the first (final-answer) reason code, which is the correct precedence. No duplicate terminal outcome is possible.
- The `live_target` tightening in `_annotate_intervention_input` closes the retroactive-adoption hole: a same-turn input persisted after the boundary now classifies `outside_target` instead of proving timely in-turn delivery, and the updated leakage test exercises exactly that same-identity-after-completion path.
- Restart safety holds: `_accepted_controls` and `_observed_control_ids` hydrate from durable accepted/observed rows, so a daemon restart between acceptance and boundary still reconciles, and the pre-existing event-id hydration prevents re-emitting terminals already journaled.
- The docs are explicit that `rpc_accepted_only` is non-observation coverage, not evidence of consumption — consistent with the #216 adoption vocabulary.

**Minor (non-blocking):**
- Entries are never removed from `_accepted_controls` after closure, so every later boundary for the same turn re-attempts reconciliation and gets absorbed by the event-id dedup. Correct but slightly wasteful; moving closed ids into `_observed_control_ids`-style bookkeeping (or a `_closed_control_ids` set) would skip the re-walks.

637 tests passing, docs updated across architecture/lifecycle/status/user-guide. Will merge once checks are green.

Issue #45, opened 2026-08-11
Surface supervision provenance and human feedback


YoungJinJung:
### Problem / outcome

A developer currently has little direct visibility into what Spotter observed, blocked, or tried to tell Main. Live supervision needs clear provenance so users can distinguish Spotter actions from agent/user intent and can disagree without digging through JSONL.

This becomes more important under the current Codex delivery primitive: `turn/steer` is persisted/model-visible as user input rather than as a distinct native supervisor role. Spotter therefore cannot rely on Codex's transcript rendering alone to preserve "this came from Spotter, not the user" semantics.

### Proposed direction

Provide a compact human-facing path to:

- see flags, deterministic blocks, reviewer decisions, delivered interventions, and relevant cost/timing
- explain a block with its rule, reason, and safe remedy/config path
- identify live `VERIFY/NUDGE` as Spotter supervision, not a new user instruction
- show whether an intervention was merely decided, RPC-accepted, observed in the turn, stale, unknown, or later judged useful/harmful
- record user disagreement/dismissal as evaluation feedback without automatically treating every dismissal as a false positive
- expose thread/session summaries through CLI/status/analyze before considering more intrusive UI

### Implementation / failure references — study, do not depend on

Current Codex steering semantics make provenance a real product requirement rather than cosmetic labeling:

- a steer can look like the latest standalone user request and displace the unfinished task: https://github.com/openai/codex/issues/29582
- a completed steer can survive compaction as user-role history and later be re-promoted: https://github.com/openai/codex/issues/30859
- steer input has `clientUserMessageId` correlation, but current upstream does not provide idempotent retry semantics: https://github.com/openai/codex/issues/32254
- originating steer/input identity is not currently carried immutably into every descendant native tool call: https://github.com/openai/codex/issues/36994

OpenClaw also documents steering as a separate privileged injection path whose normal inbound security/provenance hooks may not automatically apply:

- https://github.com/openclaw/openclaw/issues/78830

These are references only. Spotter should keep its own provenance/feedback model and should not depend on OpenClaw.

### Provenance model

Every user-visible supervision object should resolve to one durable record with a stable ID and explicit source/state.

Suggested common fields:

```text
supervision_event_id
kind = SIGNAL | BLOCK | VERIFY | NUDGE | INTERRUPT | RESTART
thread_id
turn_id?
connection_epoch?
source = spotter_gate | spotter_reviewer | spotter_navigator
trigger/evidence ids
policy/reviewer/model/config versions
[...]

PR #222, opened 2026-08-15
feat(intervention): surface advisory provenance

YoungJinJung:
## Why

Codex persists live steering as user-role input, so the advisory text itself needs an explicit, correlatable Spotter identity instead of relying on transcript role semantics.

## What changed

- derive a stable compact `spt-...` intervention ID from the durable reviewer job
- use that ID as the control/client-message correlation key
- render `[Spotter / VERIFY|NUDGE / spt-...]` in every live advisory
- retain current-turn scope and original-task ownership language
- update status documentation and regression coverage

## Validation

- `ruff format --check .`
- `ruff check .`
- `mypy src tests`
- `pytest` (637 passed)

Refs #45


YoungJinJung:
## Review

Reviewed the full diff (runtime_connection.py, status.md, test).

- The correlation chain is format-agnostic end to end, so swapping `spotter:intervention:{job_id}` for the compact `spt-<sha256[:12]>` id is safe: `_control_payload` still sets `intervention_id = control_id` whenever a `review_job_id` rides along, the #218 registry keys on whatever id was durably stored, and adoption/observation matching compares stored ids rather than parsing the prefix. Old-format ids in existing journals keep hydrating and correlating unchanged.
- The id is deterministic per review job (stable across retries of the same job, though delivery is at-most-once anyway) and 48 bits of hash is far beyond any realistic collision horizon for per-job interventions; the durable-uniqueness check would still catch one loudly.
- Rendering the id inside the visible `[Spotter / VERIFY|NUDGE / spt-…]` marker gives the transcript-visible provenance #45 asks for while keeping the ownership language intact — the updated test pins the marker, the 16-char id shape, and the preserved task-ownership phrasing.
- `hashlib` was already imported in this module; no new dependency surface.

**Process note (non-blocking):**
- The branch name `codex/feat-intervention-provenance` uses an agent-source prefix; repo convention prefers purpose-driven prefixes (`feature/`, `fix/`) — worth renaming the pattern for future branches since squash-merge discards this one anyway.

Checks green (CI + CodeQL), 637 tests passing. LGTM — merging.

PR #223, opened 2026-08-15
feat(cli): explain intervention provenance

YoungJinJung:
## Why

The visible `spt-...` marker added in #222 needs an on-demand path back to durable supervision facts. Users should be able to distinguish reviewer judgment from observed delivery state without reading JSONL.

## What changed

- add `spotter interventions` to list recent live interventions and lifecycle states
- add `spotter explain --intervention-id ...` with separate target, model-judgment, evidence-reference, and delivery sections
- join reviewer jobs/decisions to control telemetry through durable IDs
- keep unavailable fields explicit instead of inventing evidence
- document the implemented CLI surface and cover it with journal round-trip tests

## Validation

- `ruff format --check .`
- `ruff check .`
- `mypy src tests`
- `pytest` (639 passed)

Refs #45


YoungJinJung:
## Review

Reviewed the full diff (new provenance.py, cli.py, status.md, tests), and cross-checked the durable event shapes on main.

- The join runs purely on durable IDs (`review_job_id` → decision/job payloads, `intervention_id` on control events) with no adjacency or text parsing, and I verified against `review_scheduler.queued_event` that `signal_ids` and `candidate_event_ids` are real fields on queued events — so the evidence-reference section will populate in production, with the 'unavailable' fallback reserved for genuinely missing data.
- Coverage honesty carries into the human-facing surface: the explain output explicitly labels the judgment section "model judgment, not ground truth", every absent field renders as unknown/unavailable rather than being invented, and the status mapping falls back to the raw outcome/kind rather than guessing a friendly name for an unrecognized state.
- Progressive summary construction is sound: control events are folded in wall-clock order with `or summary.<field>` retention, so the last lifecycle event determines status while earlier-known fields survive — the round-trip test pins the full chain from queued job through `RPC_ACCEPTED_ONLY` with its reason code.
- Read path is appropriately defensive for a diagnostic command: `repair_tail=True` on load, and both commands fail with a clear stderr message and non-zero exit instead of a stack trace when journals are unreadable or empty.

**Minor (non-blocking):**
- Cross-journal ordering relies on wall-clock `record.at` with None-first sorting — fine for a display CLI, but if two journals ever carry the same intervention id across a clock adjustment, the "last event wins" status could pick the wrong side. The monotonic fields exist on control telemetry if this ever needs tightening.
- `_intervention_history` re-reads and re-summarizes all journals twice when `explain` follows `interventions`; irrelevant at current scale.

Checks green (CI + CodeQL), 639 tests passing, status doc updated. LGTM — merging.

PR #224, opened 2026-08-15
feat(cli): record intervention feedback

YoungJinJung:
## Why

Users need a durable way to disagree with or qualify Spotter guidance without turning a dismissal into detector ground truth or rewriting the original intervention history.

## What changed

- add `spotter feedback --intervention-id ... --category ...` with the structured #45 category vocabulary
- persist versioned feedback as append-only, locked, fsynced JSONL with stable feedback/rater identity
- redact and bound optional notes before the durable write
- reject feedback for unknown intervention IDs
- show feedback in `spotter explain` under an explicitly non-ground-truth section
- cover multiple/revised feedback, redaction, validation, unknown IDs, and CLI round trips

## Validation

- `ruff format --check .`
- `ruff check .`
- `mypy src tests`
- `pytest` with an isolated `SPOTTER_HOME` (642 passed)

A first run against the developer's persistent Spotter home hit the pre-existing daily reviewer budget in `test_live_opt_in_delivers_a_fresh_intervention_decision`; rerunning the required suite against an isolated test home passed in full.

Refs #45


YoungJinJung:
## Review

Reviewed the full diff (new feedback.py, cli.py, status.md, tests).

- The durable store follows the established label/opportunity patterns and the repo's durable-state rules: append-only JSONL under `secure_dir`, exclusive `flock` around the append, explicit fsync, 0o600 on first create, versioned rows with a future-schema refusal, and a strict per-line loader that raises instead of silently skipping.
- Redaction happens *before* the durable write and the note is bounded to 2000 chars, so secrets never touch disk — the test pins the `ghp_…` token becoming `[REDACTED]`. Category input is case-normalized against a closed #45 vocabulary with a helpful rejection listing the allowed values.
- The separation-of-evidence framing is enforced, not just documented: feedback lives in its own store, never feeds detector metrics or labels, and `explain` renders it under an explicit "evaluation evidence, not ground truth" heading. Appending a second judgment preserves the first — history is never rewritten, matching the PR's stated goal.
- The CLI refuses feedback for unknown intervention ids by checking durable history first, and distinguishes that from history-unavailable errors — both with stderr + non-zero exit.
- The PR description's disclosure about the first test run tripping the developer-home budget ledger (and the isolated-home rerun) is exactly the kind of honest validation reporting the contribution rules ask for.

**Minor (non-blocking):**
- The known-intervention check lives only in the CLI layer; a direct `add_feedback` caller could append feedback for a nonexistent id. Fine while the CLI is the only entry point — worth folding the existence check into the module if programmatic callers ever appear.
- The `.lock` sidecar file is created and never removed; harmless, just a stray file.

642 tests passing, status doc updated. Will merge once the remaining CI jobs are green.

PR #225, opened 2026-08-15
feat(cli): explain deterministic blocks

YoungJinJung:
## Why

#45 requires deterministic blocks to be as inspectable as live advisories: stable provenance, the exact policy fact that stopped work, normalized resource/effect context, and a safe next action.

## What changed

- assign every enforced or shadow block a durable `spt-block-...` supervision ID and event ID
- persist gate rule version plus normalized reversibility/effect/resource context on block events
- include BLOCK records in `spotter interventions`
- allow `spotter explain --supervision-id ...` to render policy facts separately from model-backed advisory judgments
- provide rule-specific safe remedies and explicit unknown fallbacks
- allow append-only structured feedback on block IDs as well as live intervention IDs
- update status documentation and regression coverage

## Validation

- `ruff format --check .`
- `ruff check .`
- `mypy src tests`
- `pytest` with isolated `SPOTTER_HOME` (643 passed)

Refs #45


YoungJinJung:
## Review

Reviewed the full diff (core.py, gates.py, provenance.py, cli.py, status.md, tests).

- Block provenance is stamped at the source: every enforced/shadow block now carries a durable `spt-block-…` supervision id, `rule_version`, and the normalized reversibility/effect/resource context copied from the proposal payload, with a deterministic `spotter:supervision:{id}` event id. The addition to the PreToolUse path is a uuid + dict merge — bounded and effectively free, so the synchronous-gate latency rule holds.
- Backward compatibility is coverage-honest: `summarize_blocks` simply skips legacy block events without a supervision id rather than fabricating identities for them, so old journals list nothing false and new ones are fully inspectable.
- The explain output keeps the epistemic separation #45 asks for — blocks render "Observed policy facts" (rule/version/reason/resource/effect) while advisories keep the "model judgment, not ground truth" section, and both share the same append-only feedback surface via the id-agnostic `_print_feedback`/known-id check. The `--supervision-id` alias mapping onto the same dest keeps the old flag working.
- Remedies are a closed static map keyed by rule name with an explicit safe fallback for unknown rules — actionable without ever guessing, and the dependency_change remedy correctly points at the configuration knob rather than telling the user to bypass the gate.
- The hook test extension proves the enriched payload flows through the real local-gate fallback path, not just a synthetic journal.

**Minor (non-blocking):**
- `_block_history` globs `*.jsonl` (including app-server journals that won't contain gate blocks) while `_intervention_history` scans only `app-server-*.jsonl` — harmless double-read today, but unifying the two scans would halve the I/O for `interventions`/`explain`.
- `GATE_RULE_VERSION` is a single global constant; if individual rules ever evolve independently, per-rule versions would date blocks more precisely.

Checks green (CI + CodeQL), 643 tests passing, status doc updated. LGTM — merging.

PR #226, opened 2026-08-15
fix(observability): classify advisory state coverage

YoungJinJung:
## Why

After Spotter advisory inputs stopped replacing `ThreadState.task.goal`, the observability audit still checked them against that goal slot. Correctly preserved advisories were therefore reported as dropped user-goal evidence, corrupting App Server parity measurements.

## What changed

- classify Spotter-authored `user_prompt` events as intervention-delivery evidence in source and Trace IR coverage
- verify advisory preservation against `ThreadState.supervision.interventions` instead of `task.goal`
- retain ordinary user-prompt coverage against the real goal slot
- reuse the collection-preservation check for evidence items
- add a regression proving the user goal stays untouched while intervention coverage is exact
- update current-status documentation

## Validation

- `ruff format --check .`
- `ruff check .`
- `mypy src tests`
- `pytest` with isolated `SPOTTER_HOME` (644 passed)

Refs #22
Refs #45
Refs #86


YoungJinJung:
## Review

Reviewed the full diff (observability.py, status.md, tests).

- This is the right root-cause fix for the #218 interaction: since supervision-tagged prompts deliberately stopped touching `task.goal`, checking them against the goal slot made every correctly-preserved advisory read as `ADAPTER_DROPPED`, silently deflating App Server parity. The fix reclassifies at the shared classification points (`source_audit_sample`, `_trace_families`, `state_coverage_status`), so every audit path agrees rather than one being patched.
- The family override and the state check key on the same durable marker (`input_origin == \"spotter_supervision\"`) that the live reconciler stamps and persists, so hydrated journals classify identically to live ingestion.
- Ordinary user prompts keep their exact previous semantics (goal-slot check), and extracting `_state_collection_status` reuses the existing evidence-collection logic instead of duplicating the event-id scan — the preservation check is by provenance event id in either collection.
- The regression test is end-to-end through the real normalizer: goal stays `None`, the intervention item carries the input's event id, the family is exactly `INTERVENTION_DELIVERY`, and state coverage is `OBSERVED_EXACT`.

**Minor (non-blocking):**
- `supervision.interventions` is a bounded tuple, so an advisory evicted by the bound would report `ADAPTER_DROPPED` in a very long session — arguably honest (it *was* dropped from live state), just worth knowing when reading parity numbers for advisory-heavy sessions.

Checks green (CI + CodeQL), 644 tests passing, status doc updated. LGTM — merging.

Issue #98, opened 2026-08-13
Generalize reversibility classification and external-effect lifecycle


Bogyie:
### Problem / desired outcome

#30, implemented by #97, establishes the minimum recovery contract: actions receive an A/B/C reversibility class, Class B mutations carry checkpoint lineage, and Class C outcomes are persisted as `external_effect` records.

That is enough to make restart/recovery honest, but the current classifier is intentionally narrow. `effects.py` relies on a small hard-coded command vocabulary plus MCP/tool-name heuristics. As Spotter supports more tools and transports, this will produce both conservative false positives and unsupported cases. For example, remote CLIs such as `gh`, `kubectl`, and `terraform` are currently treated as external writes as a family even when the concrete operation is observational, while wrappers, shell composition, scripts, network clients, cloud CLIs, and richer MCP semantics are not modeled deeply.

The effect ledger is similarly minimal: it records that an external effect occurred, but has limited outcome semantics, stable effect identity, reconciliation, or representation of a later explicit reversal/compensation.

Evolve this into an extensible, transport-independent effect model that remains bounded and auditable on the synchronous path while providing enough fidelity for recovery, intervention policy, and future adapters.

### Proposed direction

- Define a normalized classification input / effect descriptor around semantic operation, resource, tool/provider provenance, outcome, and reversibility instead of coupling policy to raw Hook payloads.
- Replace the single closed vocabulary with composable classifiers or registries for major families: shell/Git, remote CLIs, MCP tools, and adapter-provided operations.
- Make command-family classification verb-aware where semantics are known, e.g. distinguish `gh ... view/list` from writes, `kubectl get` from `apply/delete`, and `terraform plan` from `apply`.
- Handle common bounded shell structure such as `env`/`sudo` wrappers, command chains, pipelines, redirection, and interpreter/script entry points without attempting arbitrary execution or model-based interpretation.
- Allow explicit per-tool/per-server configuration for MCP or unknown tools, with conservative Class C fallback when Spotter cannot prove a safer class.
- Record why a classification was chosen (classifier/rule provenance) so decisions remain inspectable and testable.
- Give external effects stable correlation/identity and richer outcome state such as attempted / succeeded / failed / unknown / partial, with duplicate suppression when the same semantic action is observed through overlapping sources.
- Represent later explicit reversal, compensation, or reconciliation of an external effect in the ledger. Automatic compensating rollback is not required by this issue.
[...]

PR #227, opened 2026-08-15
feat(effects): classify remote command semantics

YoungJinJung:
## Why

Issue #98 identifies broad family-level Class C treatment as both noisy and unauditable. The synchronous path needs bounded verb-aware semantics without becoming an interpreter.

## What changed

- classify known GitHub CLI, kubectl, and Terraform read/write operations by verb
- unwrap bounded env, sudo, and shell command layers; combine chains, pipelines, and redirection conservatively
- preserve explicit unknown states for unsupported family commands, scripts, malformed shell, and excessive nesting
- persist classifier, reason, confidence, semantic operation, and normalized resource provenance on Hook and external-effect records
- document the current boundary and remaining #98 work

## Validation

- ruff format --check .
- ruff check .
- mypy src tests
- SPOTTER_HOME=/tmp/spotter-test-home.wvWzFP pytest (650 passed)

Refs #98


YoungJinJung:
## Review

Reviewed the full diff (effects.py, hook.py, core.py, docs, tests) adversarially, since this classifier's output can downgrade previously-conservative Class C treatment to A/B on the synchronous path.

**What holds up well:**
- The conservative spine is intact almost everywhere: any unknown component poisons a composition to unknown-C (`_strongest` prefers the unknown over the strongest known), scripts/malformed shell/over-deep wrappers/oversized commands all stay explicit unknown-C, and output redirection upgrades an A read to a B local write instead of staying A — several of these are actually *more* conservative than the old `_words`-based fallback.
- Verb allowlists are closed sets per family; unsupported subcommands (`kubectl exec`, `terraform workspace`) correctly land in family-specific unknown reasons rather than being guessed. Provenance (classifier/reason/confidence/semantic op) flows through hook records, block context, and the Class C effect ledger with tests at each layer.
- Wrapper unwrapping is bounded (`depth <= 2`), `env`/`sudo` option tables are value-aware, and `bash -c` extraction refuses script files. Confidence properly degrades to `bounded` through any wrapper.

**One real gap — please fix before merge:**
- `gh api -XPOST repos/org/repo/issues` (shorthand with *attached* value, which gh's pflag parsing accepts, same as curl habit) is not seen by `_option_value` — it matches neither `-X` exactly nor `-X=`. With no write fields present, the method defaults to GET and the command classifies as `gh.api.get`, **Class A external_read** — a real remote write labeled as a read. That inverts the module's stated invariant (unknown shapes must never be mislabeled as safe) in the one place an agent is most likely to produce the pattern. Suggested minimal fix: in `_classify_gh`'s api branch, treat any token matching `-X<value>` / `--method=<value>` via a prefix scan (or route any unrecognized `-X`-prefixed token to unknown) so attached-value shorthand either resolves to its method or degrades to unknown-C.
- Same prefix-scan consideration applies to `_option_value` generally if other families ever grow attached-value shorthands, but `gh api` is the only currently reachable case I found.

Everything else looks merge-ready (650 tests passing, CI green, docs updated with the honest #98 boundary). I'll merge once the `-XPOST` handling lands.

YoungJinJung:
Fixed the attached short-option case in 9f8a9a5: `-XPOST` now resolves through the shared option parser (and short attached `-R`/`-n` values are handled consistently). Added a regression proving it remains Class C / `gh.api.post`. Full suite remains green: 650 passed.

YoungJinJung:
The follow-up commit resolves the review finding exactly as suggested: `_option_value` now recognizes attached two-char shorthand values, so `gh api -XPOST …` resolves to its real method and classifies `gh.api.post` / Class C — with a regression case added to the semantics table. Side effects are all improvement-direction (`-nprod` for kubectl namespace, `-Rorg/repo` for gh resource now parse; `-out` is unaffected at 4 chars). Checks green — merging.

PR #228, opened 2026-08-15
feat(effects): classify HTTP and database commands

YoungJinJung:
## Why

Issue #98 calls for operation-aware external command classification. After the remote CLI slice, curl and database clients were still blanket unknown/Class C even for bounded reads, while their resource strings could retain URL secrets.

## What changed

- classify curl GET/HEAD/OPTIONS, remote writes, uploads, request bodies, and local output files conservatively
- reject config-driven, multi-request, unsupported-method, and ambiguous read-with-body shapes as explicit unknown Class C
- strip URL credentials, query strings, and fragments before resource persistence
- classify bounded PostgreSQL/MySQL metadata reads and known write verbs while keeping SELECT, scripts, interactive use, multiple commands, and mixed statements unknown
- document the extended #98 boundary

## Validation

- ruff format --check .
- ruff check .
- mypy src tests
- SPOTTER_HOME=/tmp/spotter-test-home.wvWzFP pytest (654 passed)

Refs #98


YoungJinJung:
## Review

Reviewed the full diff (effects.py, docs, tests) adversarially, same lens as #227 — hunting for any shape where a remote write could inherit read semantics.

**No safety inversions found:**
- The read path is guarded on both sides: read methods with a request body (`-XGET -d …`) go unknown-C rather than A, while `-G` (which genuinely converts data to query params) legitimately reads. Config-driven (`-K`), multi-request (`--next`), unsupported methods (`-XPROPFIND`), and schemeless URLs all refuse classification instead of defaulting to GET — each pinned by the never-inherit-GET test.
- The #227 attached-shorthand fix pays off here: `-XDELETE` and `-eUPDATE…` both resolve correctly, and I checked the inverse direction — a clustered `-sG` that escapes `_has_option` degrades to POST/C (over-conservative), never to a false read.
- SQL classification is appropriately narrow: single-statement only, metadata reads (`SHOW`/`DESCRIBE`) as the only A verbs, and — the right call — `SELECT` stays unknown because function calls can mutate. Scripts, init-commands, interactive sessions, multiple `-c` options, and semicolon-split statements (including literals containing `;`, which err conservative) all refuse. psql backslash metacommands land in unsupported_sql_shape.
- Secret hygiene is real: `_sanitize_remote_resource` rebuilds the netloc from hostname+port only (dropping userinfo), strips query/fragment, and re-brackets IPv6; the test persists a URL with embedded credentials and token query and asserts the clean resource.

**Minor (non-blocking):**
- Clustered short options hide the output flag: `curl -so out.json URL` won't match `-o`'s attached rule (token starts with `-s`), so a download classifies A instead of B. Read-vs-local-write fidelity only — no external-write risk — but common enough idiom to note for a follow-up.
- The `ValueError` fallback in `_sanitize_remote_resource` (e.g. an out-of-range port) strips query/fragment textually but keeps any `user:secret@` userinfo. Rare shape; a `partition('@')` in the fallback would close it.

Checks green (CI + CodeQL), 654 tests passing, architecture/status docs updated with the extended boundary. LGTM — merging.

PR #229, opened 2026-08-15
fix(effects): parse clustered curl options

YoungJinJung:
## Why

The #228 review identified two fidelity gaps after merge: common clustered curl output flags such as `-so` could stay Class A instead of local-write Class B, and malformed-port URL fallback could retain userinfo.

## What changed

- parse value-taking short options inside bounded clusters (`-so`, `-sXPOST`, `-sd`, `-sKfile`)
- reuse clustered recognition for boolean/presence checks so config-driven requests still refuse classification
- strip userinfo in the invalid-URL fallback before persisting the resource
- add regressions for output, POST/data, config, and malformed-port credential cases

## Validation

- ruff format --check .
- ruff check .
- mypy src tests
- SPOTTER_HOME=/tmp/spotter-test-home.wvWzFP pytest (656 passed)

Refs #98

### DIFF
--- a/src/spotter/effects.py
+++ b/src/spotter/effects.py
@@ -112,6 +112,37 @@ _SCRIPT_RUNNERS = frozenset(
 )
 _HTTP_READ_METHODS = frozenset({"GET", "HEAD", "OPTIONS"})
 _HTTP_WRITE_METHODS = frozenset({"DELETE", "PATCH", "POST", "PUT"})
+_CURL_BOOLEAN_SHORT_OPTIONS = frozenset("#0123456789:BaGfghiIJjklLMnNOqRSsvVZ")
+_CURL_VALUE_SHORT_OPTIONS = frozenset(
+    {
+        "A",
+        "b",
+        "C",
+        "c",
+        "D",
+        "d",
+        "E",
+        "e",
+        "F",
+        "H",
+        "K",
+        "m",
+        "o",
+        "P",
+        "Q",
+        "r",
+        "T",
+        "t",
+        "U",
+        "u",
+        "w",
+        "X",
+        "x",
+        "Y",
+        "y",
+        "z",
+    }
+)
 _SQL_READ_VERBS = frozenset({"DESC", "DESCRIBE", "SHOW"})
 _SQL_WRITE_VERBS = frozenset(
     {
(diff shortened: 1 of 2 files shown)


YoungJinJung:
## Review

Reviewed the full diff. The two #228 findings are addressed as described — clustered value-taking options resolve (`-so`, `-sXPOST`, `-sd`, `-sK…` all pinned) and the invalid-port fallback now strips userinfo. However, the cluster matcher introduces a **new safety inversion**, so please hold the merge:

**Blocking finding — false-positive flag detection inside attached values:**
- `_short_cluster_contains(token, flag)` matches the flag character *anywhere* in `token[1:]`, including inside another option's attached value. Concrete counterexample:

  `curl -dsize=BIG https://api.example.com/x`

  Real curl: `-d size=BIG` → **POST** with a body. Classifier: `has_data` is true, but `_has_option(words, \"-G\", \"--get\")` also fires because the token `-dsize=BIG` "contains G" — so the method resolves to GET via the `-G` branch, and the read-with-body guard is skipped *because* `-G` appears present. Result: a real remote POST classifies **Class A external_read**. Any `-d`/`-F` attached value containing an uppercase `G` (or `I`) reproduces it.
- The same mechanism can misread other flags out of values (e.g. `-dXfoo` yields method `FOO`), though those all degrade to unknown-C — conservative. The `-G`/`-I` presence checks are the ones that flip toward *permissive*.

**Suggested fix:** scan clusters left-to-right instead of substring-matching: iterate characters after `-`; a known boolean flag consumes one char; the first known *value-taking* option char (`X d F T o c D w K u …`) consumes the remainder (or the next token) as its value and terminates the scan; an unrecognized char refuses the cluster as `unsupported_curl_shape`. That preserves all four new regression cases while making `-dsize=BIG` invisible to the `-G` check (its chars are inside `-d`'s value). A regression asserting `curl -dsize=BIG URL` classifies `http.post`/C would pin it.

Everything else (fallback credential strip, the four cluster regressions) is good — I'll merge once the cluster scan is ordered rather than substring-based.

YoungJinJung:
Fixed the blocking cluster safety inversion in 33295b3 by replacing substring flag detection with a bounded left-to-right curl option scanner: boolean flags consume one character, value-taking flags consume the remaining token/next argument, and unknown flags refuse classification. Added the exact `-dsize=BIG` regression (Class C / `http.post`) plus an unsupported-cluster case. Full suite remains green: 656 passed.

YoungJinJung:
The follow-up commit resolves the blocking finding exactly along the suggested lines: `_normalize_curl_args` scans clusters left-to-right against explicit boolean/value-taking option tables, the first value-taking char consumes the remainder (or next token) as its value, and any unrecognized char refuses the whole command as `unsupported_curl_cluster`. The substring matcher is fully removed — including its reach into the generic `_option_values`/`_has_option`/`_has_option_prefix` helpers, so the other families return to their pre-#229 attached-prefix behavior. The `-dsize=BIG` counterexample now pins `http.post`/C, and `-s@` pins the cluster refusal. Missing trailing values also refuse rather than guess. Checks green — merging.

PR #230, opened 2026-08-15
feat(effects): configure exact MCP tool semantics

YoungJinJung:
## Why

MCP tool-name verbs are only a fallback heuristic: the same tool name can have different effects on different servers. Issue #98 calls for bounded, explicit server/tool metadata before the bundled heuristics.

## What changed

- parse exact `[mcp_semantics."server"."tool"]` rules with operation, A/B/C reversibility, and declared resource fields
- reject malformed or unsafe combinations and secret-bearing resource field names
- apply configured rules before MCP name heuristics while retaining conservative Class C fallback
- persist exact classifier, semantic operation, resource, and confidence provenance through Hook events
- document the configuration and update current capability status

Part of #98.

## Validation

- `ruff format --check .`
- `ruff check .`
- `mypy src tests`
- `pytest` (665 passed)


YoungJinJung:
## Review

Reviewed the full diff (config.py, effects.py, hook.py, docs, example config, tests).

- The config schema enforces the safety lattice at load time, which is the right trust boundary: `read` must be A, `write`/`delete` can never be A, `unknown` must be C, unknown keys/operations reject, duplicates (casefolded) reject, and secret-bearing resource field *names* are refused — every rule pinned by the parametrized rejection test. A user can still deliberately declare trust for a specific server/tool pair, which is exactly what #98 asks for, and the docs are explicit that tool descriptions never authorize a safer class.
- Matching is exact and conservative: identity requires the full `mcp__server__tool` shape (single-underscore and malformed names fall through), configured rules take precedence over the name heuristics, and anything unmatched keeps the existing verb-heuristic → unknown-C ladder — the unknown-server test proves an identically named tool on another server stays C.
- Resource persistence is opt-in and bounded: only declared scalar fields render, URLs route through the existing `_sanitize_remote_resource` (the test embeds credentials + token query and asserts the clean value), and the fallback resource is the server/tool identity rather than arbitrary input.
- `reversible = cls != \"C\"` keeps configured Class B checkpoint-eligible (pinned by its own test), and provenance (`mcp_config` / `configured_semantics` / exact confidence / `mcp.<server>.<tool>.<op>`) flows through hook events end to end.
- No import cycle: effects → config is one-way, and the defaulted parameter keeps every existing `classify` call site working.

**Minor (non-blocking):**
- The linear scan over `semantics[:256]` per MCP call is fine at the cap, but a dict keyed by (server, tool) built once in `from_mapping` would make the lookup O(1) and drop the defensive slice.
- Secret screening is name-based only; a declared innocuous field could still carry a secret value. That's inherently the config author's responsibility and the example config warns about it — just noting the boundary.

Checks green (CI + CodeQL), 665 tests passing, docs and example config updated. LGTM — merging.

PR #231, opened 2026-08-15
feat(effects): record evidence-backed external outcomes

YoungJinJung:
## Why

A tool process exiting zero does not prove that its intended external mutation committed. Recovery currently records that case as succeeded, which overstates the evidence available from a generic Hook result.

## What changed

- classify external-effect results as `succeeded`, `failed`, `partial`, or `unknown`
- require explicit protocol success evidence before recording `succeeded`
- retain exit-zero-only results as `unknown` with auditable `outcome_evidence`
- expose the normalized outcome in recovery digests while preserving the legacy `result` field
- update current capability status and regression coverage

Part of #98.

## Validation

- `ruff format --check .`
- `ruff check .`
- `mypy src tests`
- `pytest` (666 passed)


YoungJinJung:
## Review

Reviewed the full diff (effects.py, digest.py, status.md, tests).

- The evidence ladder is ordered sensibly: explicit partial beats everything (a partially-committed mutation is the most important thing to surface), explicit failure next via the shared `outcome_failure` helper (consistent with how metrics already judge failures), explicit success requires protocol-level evidence (`ok`/`success` true or a success-vocabulary status), and exit-zero-only degrades to `unknown` with the auditable `exit_zero_only` evidence — exactly the epistemic honesty the recovery digest needs, since a zero exit never proved the remote mutation committed.
- The `value == 0 and not isinstance(value, bool)` guard correctly keeps `False` from counting as a zero exit code.
- Legacy compatibility is handled at both ends: new rows carry both `result` and `outcome` (same value), and the digest reads `outcome` with a `result` fallback so pre-existing journaled effects still render their historical values.
- The behavior change is deliberate and disclosed — the git-push hook test now pins `unknown`/`exit_zero_only` where it previously claimed `succeeded`, and the status doc says the same. That's the point of the PR, not a regression.
- The five-way outcome test covers every evidence branch including the no-response case.

**Minor (non-blocking):**
- Explicit-success detection trusts the tool's own status strings; a misbehaving tool could still claim success falsely. That's inherent to protocol-level evidence and strictly better than the old exit-code inference — just noting the boundary, which #98's reconciliation slice is the eventual answer to.

Checks green (CI + CodeQL), 666 tests passing, status doc updated. LGTM — merging.

PR #232, opened 2026-08-15
feat(effects): correlate external effect observations

YoungJinJung:
## Why

The same external mutation can be observed through overlapping Hook and App Server surfaces. Recovery must identify one logical effect without using heuristic fingerprints to suppress legitimate repeated actions.

## What changed

- promote Hook tool-use IDs into normalized operation/item metadata
- derive stable effect IDs from native operation IDs and mark fallback fingerprints as inferred
- retain source event/operation/item/tool identifiers and runtime provenance on effect records
- merge exact duplicate observations in recovery projections while preserving all inferred observations
- surface conflicting terminal observations as unknown with the complete observed outcome set
- update current capability status and regression coverage

Part of #98.

## Validation

- `ruff format --check .`
- `ruff check .`
- `mypy src tests`
- `pytest` (668 passed)


YoungJinJung:
## Review

Reviewed the full diff (effects.py, hook.py, status.md, tests).

- The exact/inferred split is the right safety design: only native operation-id-derived effect ids ever merge, while fingerprint-derived ids are marked `inferred` and never deduplicate — so overlapping Hook/App Server observations of one mutation collapse, but two legitimate identical pushes never disappear. Both properties are pinned by dedicated tests.
- Conflict handling refuses to pick a winner: divergent terminal observations force `unknown` with `conflicting_observations` and the complete `observed_outcomes` set, which composes correctly with #231's evidence ladder (the test merges an `ok: true` and an `exit_code: 1` observation of the same operation).
- `uuid5` over a fixed namespace keeps effect ids deterministic across restarts and processes — same operation id always yields the same effect id regardless of which surface observed it first.
- Promoting the Hook `tool_use_id` into `operation_id`/`item_id` and stamping identity/provenance onto the event before recording is what makes cross-surface correlation possible at all, and it aligns Hook events with how the runtime-metrics action key already correlates. Merged payloads remain dict supersets, so the recovery digest and other `external_effects` consumers read unchanged fields.
- Source-id unions and observation counts are bounded (16 ids / validated positive ints), keeping merged records compact.

**Minor (non-blocking):**
- On a third observation of an already-conflicted effect, the derived `unknown` outcome from the previous merge leaks into `observed_outcomes` (e.g. `[\"failed\", \"succeeded\", \"unknown\"]`) because `previous.outcome` is folded into the set alongside genuine observations. Still conservative — the outcome stays unknown — but excluding the synthetic `unknown` from the observed set would keep it purely observational.

Checks green (CI + CodeQL), 668 tests passing, status doc updated toward the remaining #98 reconciliation slice. LGTM — merging.

PR #233, opened 2026-08-15
feat(effects): record explicit effect resolutions

YoungJinJung:
## Why

Recovery must preserve an external mutation even after a later probe, reversal, or compensating action. Rewriting or deleting the original effect would erase evidence, while a compensation claim without its own journaled effect would overstate recovery.

## What changed

- add append-only reversal, compensation, reconciliation-present/absent, and unresolved records
- project resolution history onto exact effects without removing the original effect
- require compensation to reference a separately journaled exact effect before marking the target resolved
- add `spotter effects list|resolve --session ...` for explicit operator evidence
- expose resolution records in analysis digests and document the workflow
- keep conflict outcome sets purely observational on third and later correlated observations

Part of #98.

## Validation

- `ruff format --check .`
- `ruff check .`
- `mypy src tests`
- `pytest` (675 passed)


YoungJinJung:
## Review

Reviewed the full diff (effects.py, cli.py, digest.py, docs, tests).

- The core honesty property holds everywhere it matters: a compensation claim only marks its target `resolved` when the related compensating effect exists as a separately journaled *exact* effect — verified at projection time (`relation_verified`), refused at the CLI before a record is even written, and pinned by tests at both layers. `reconciled_present` correctly never resolves, `reconciled_absent` does.
- Append-only is genuine: the original `external_effect` rows are never rewritten; resolutions project onto the merged view with a bounded (16-entry) history, and legacy/unknown resolution values are ignored rather than misapplied. Resolutions targeting non-exact ids are silently inert in projection and unreachable through the CLI.
- Input validation on `effect_resolution_event` is tight — closed vocabulary, bounded ids with prefix check, 1–500 char evidence, self-resolution rejection — each with a rejection test.
- Nice follow-through on the #232 review note: `observed or (previous.get(\"outcome\"),)` keeps the synthetic `unknown` out of `observed_outcomes` on third-and-later merges, and the extended three-observation test pins the purely observational set.
- `uuid4` for resolution event ids is fine here since the id is minted once at CLI record time, not re-derived; digest and analysis surfaces render both effect observations and resolution lines.

**Minor (non-blocking):**
- `spotter effects resolve` validates the target against the *current* projection, so an effect id from another session's journal correctly refuses — but the error message says "was not found" without hinting the session mismatch case; a tiny wording tweak would save an operator a puzzled minute.

Checks green (CI + CodeQL), 675 tests passing, user guide and status docs updated. LGTM — merging.

PR #234, opened 2026-08-15
feat(effects): report classifier coverage gaps

YoungJinJung:
## Why

Unsupported effect semantics currently become conservative Class C, but without coverage metrics those gaps remain invisible and classifier work cannot be prioritized from evidence.

## What changed

- measure tool proposals once, excluding duplicate result observations
- report exact and bounded classification coverage
- split unknowns into command family, shell shape, MCP tool, and adapter operation buckets
- report concrete unknown reason counts, missing legacy provenance, and conservative-C fallbacks
- include effect coverage in `spotter metrics` and document the output

Part of #98.

## Validation

- `ruff format --check .`
- `ruff check .`
- `mypy src tests`
- `pytest` (676 passed)


YoungJinJung:
## Review

Reviewed the full diff (effects.py, cli.py, docs, tests).

- Counting only `tool_proposal` events is the right denominator — each operation classifies once, and result-side observations of the same operation can't inflate coverage. Legacy rows without `effect_confidence` land in an explicit `missing_provenance` count rather than being guessed into a bucket.
- The unknown partition (mcp reason → mcp, `shell_structure` classifier → shell, `unknown_tool_effect` → adapter, else family) lines up with the reason-code vocabulary established across #227–#230, and `conservative_c_from_unknown` gives exactly the prioritization signal the PR is after: how much Class C weight is opinion-free fallback rather than recognized semantics.
- The test is end-to-end through `event_from_hook` with real classifier outputs — one representative per bucket plus a provenance-less row — so the expected tuple (7 ops, 1 exact, 1 bounded, four unknown buckets, 4 conservative-C, 1 missing) pins the actual pipeline, not synthetic payloads.
- Rendering follows the repo's coverage-honest style with concrete reason counts and a `none` fallback; the metrics integration streams all runtime journals without new I/O.

No blocking or non-blocking findings beyond that. Checks green (CI + CodeQL), 676 tests passing, docs updated. LGTM — merging.

PR #235, opened 2026-08-15
feat(effects): enrich App Server effect lifecycle

YoungJinJung:
## Why

App Server observations did not carry the same reversibility provenance and durable external-effect lifecycle as Hook events, leaving adapter coverage incomplete for #98.

## What changed

- classify App Server command, MCP, dynamic-tool, and file-change items with bounded provenance
- load configured MCP semantics through the daemon recovery path
- derive durable effect rows from completed App Server operations with native correlation and evidence-backed outcomes
- conservatively downgrade Class B observations when App Server cannot prove a pre-mutation checkpoint
- document the current lifecycle coverage

Part of #98.

## Validation

- ruff format --check .
- ruff check .
- mypy src tests
- pytest (680 passed)


YoungJinJung:
## Review

Reviewed the full diff (ingestion.py, effects.py, daemon.py, runtime_connection.py, docs, tests).

**What holds up well:**
- Cross-surface parity is real: App Server command/MCP/dynamic-tool/file-change items now carry the same classification provenance through the shared `classify`, and the correlation test proves a Hook `tool_use_id` and an App Server item id for the same operation yield the same effect id — exactly what #232's exact-merge needs.
- The Class B downgrade is the right conservative call: without provable pre-mutation lineage, B observations become explicit `uncheckpointed_*` Class C effects with `expected_reversibility_class`/`checkpoint_required` provenance rather than silently claiming recoverability. Configured MCP semantics load through the daemon path with a fail-visible stderr fallback, and the mcpToolCall test shows the model-authored `description` field never influences classification.
- Effect derivation at ingest time is restart-safe via the existing `_seen` dedup, and A-class reads still produce no ledger rows.

**One finding — please fix before merge (consistency with #231):**
- `_effect_outcome` treats any `status == \"completed\"` as explicit success, but for App Server `commandExecution` items that `status` is the *lifecycle* field (\"the process finished\"), not a protocol assertion that the mutation committed. The new test pins `git push` + `exitCode: 0` + `status: completed` → outcome `succeeded` — which is precisely the exit-zero-only overstatement #231 removed on the Hook path (where the same command pins `unknown`/`exit_zero_only`). The same remote mutation now gets two different evidentiary claims depending on which surface observed it, and the recovery digest overstates the App Server one.
- Suggested minimal fix: when an exit code is present, don't let a lifecycle \"completed\" status satisfy `_has_explicit_success` — let the exit code path decide (0 → `exit_zero_only`/unknown, nonzero → failed). A \"completed\" status with *no* exit code can keep its current weight for tool-asserted statuses. Updating the new command-effect test to pin `unknown`/`exit_zero_only` would align both surfaces.

Everything else is merge-ready (680 tests passing, CI green, status doc updated). I'll merge once the outcome-evidence asymmetry is resolved.

YoungJinJung:
The follow-up commit resolves the finding exactly as suggested: with an exit code present, a lifecycle `completed` status no longer satisfies explicit success — while genuine success assertions (`succeeded`/`success`) keep their weight — so the same `git push` now reports `unknown`/`exit_zero_only` identically on both the Hook and App Server surfaces. Both the ingestion test and a new command_result unit case pin the aligned behavior. Checks green — merging.

PR #236, opened 2026-08-15
feat(effects): record attempted operations

YoungJinJung:
## Why

Class C proposals were distinct from completed effects, but Spotter still had no durable state for an accepted App Server operation whose terminal result had not arrived. That gap could make recovery treat a possibly-started external mutation as absent.

## What changed

- derive attempted effect observations from App Server command, MCP, dynamic-tool, and file-change starts
- enrich the same stable effect with terminal evidence when completion arrives
- preserve terminal outcomes when start/completion observations arrive out of order
- keep Hook proposals separate from accepted work and document the lifecycle contract

Part of #98.

## Validation

- ruff format --check .
- ruff check .
- mypy src tests
- pytest (682 passed)


YoungJinJung:
## Review

Reviewed the full diff (effects.py, ingestion behavior via tests, docs).

- The gap this closes is real: an accepted App Server operation whose terminal result never arrived previously produced no durable effect row, so recovery could treat a possibly-started remote mutation as absent. Now a Class C start yields an `attempted` observation with `operation_started` evidence, and it stays visible until terminal evidence arrives.
- The merge handles both arrival orders correctly: when the terminal observation lands first, the swapped merge (`{**current, **previous}`) keeps the terminal outcome from being overwritten by a late-arriving start — pinned by the out-of-order test. Attempted placeholders are also discarded from the conflict set, so an attempt can never manufacture a fake conflict against its own terminal result, while genuine terminal-vs-terminal conflicts still surface.
- The boundary contract is kept clean: Hook `tool_proposal`s remain proposal evidence and generate no effect rows — only App Server *accepted* starts do — while Hook PostToolUse results keep producing completed observations. The architecture doc states this distinction explicitly.
- The #235 evidence discipline carries through: the ingestion test shows a started `git push` that completes with exit 0 merges to `completed`/`unknown`/`exit_zero_only`, not to success. Two attempts merge to a still-`attempted` row with unknown outcome.
- Widening `external_effects` to `Iterable` matches how the CLI already streams records.

No blocking or non-blocking findings beyond that. 682 tests passing, architecture/status docs updated. Will merge once the remaining CI jobs are green.

Issue #26, opened 2026-08-11
Implement INTERRUPT and RESTART


YoungJinJung:
### Problem / outcome

Spotter's strong-control rungs do not exist in the live runtime. The Recover stage needs a way to stop a compounding turn and, only under much stronger evidence, start a fresh continuation from a sound state.

### Proposed direction

**INTERRUPT**

- use App Server `turn/interrupt`
- start with a shadow `would_interrupt` mode
- record target turn, evidence, latency, and result
- require explicit replanning after interruption

**RESTART**

- build a minimal new continuation from the user goal, explicit constraints, verified evidence, current repository/checkpoint lineage, and intentionally retained artifacts
- preserve lineage so recovery is reproducible/auditable
- surface external effects that the restart cannot undo (#30)

Both primitives remain explicitly gated/off until their evidence budgets are met.

### Implementation references — study, do not depend on

OpenClaw provides working Codex App Server interruption code and, more importantly, has already exposed several failure modes Spotter should design around:

- exact active-turn tracking and `turn/interrupt`: https://github.com/openclaw/openclaw/blob/main/extensions/codex/src/conversation-control.ts
- Codex runtime/control boundary: https://github.com/openclaw/openclaw/blob/main/docs/plugins/codex-harness-runtime.md
- **do not use interrupt as a generic finalizer**: https://github.com/openclaw/openclaw/issues/111019
- cleanup/active-run leaks around partially initialized App Server turns: https://github.com/openclaw/openclaw/issues/99271

Shepherd is the strongest reference for reversible execution / revert-oriented meta-agent experiments:

- runtime substrate: https://github.com/shepherd-agents/shepherd
- live intervention experiment: https://github.com/shepherd-agents/shepherd-experiments/tree/main/exp/live-intervention
- meta-agent implementation: https://github.com/shepherd-agents/shepherd-experiments/blob/main/exp/live-intervention/src/experiment_live_intervention/core/meta_agent_llm.py

These are implementation references only; Spotter retains its own Python runtime and recovery model.

#### INTERRUPT: treat it as a semantic state transition, not a cleanup RPC

OpenClaw found that `turn/interrupt` causes Codex to persist an abort marker equivalent to "the user interrupted the previous turn" in canonical thread context. That marker can affect later model behavior and the interrupt also stops genuinely in-flight work.

Therefore Spotter should call `turn/interrupt` only when it intentionally wants to abort the reasoning/tool trajectory. Never use it merely to:

[...]

PR #237, opened 2026-08-15
test(effects): validate adapter lifecycle boundary

### LINKED ISSUE #98: Generalize reversibility classification and external-effect lifecycle

YoungJinJung:
## Why

The remaining #98 adapter boundary needed an explicit regression proving unknown dynamic operations stay conservative even when their names and model-authored descriptions imply a read.

## What changed

- pin dynamic adapter starts/results to conservative Class C fallback with unknown provenance
- prove accepted and terminal observations correlate into one evidence-backed effect
- mark the implemented reversibility/effect lifecycle scope complete in current status

Closes #98.

## Validation

- ruff format --check .
- ruff check .
- mypy src tests
- pytest (683 passed with a fresh isolated SPOTTER_HOME)


YoungJinJung:
## Review

Reviewed the full diff (test + status flip).

- The regression pins the exact boundary property the adapter slice needed: a dynamic tool whose *name* (`read_resource`) and model-authored `description` ("safe read") both imply a read still classifies as unknown-C `fallback`/`unknown_tool_effect` — neither naming nor descriptions can promote safety, which is the #230/#235 trust rule applied to the last uncovered family.
- The lifecycle half composes everything the series built: the accepted start and terminal result correlate into one `exact` effect with `observation_count == 2`, and the `succeeded` outcome comes from the explicit `success: true` protocol assertion — not the lifecycle `completed` status — consistent with the #235 evidence fix.
- Flipping Reversibility/effects to ✅ is justified now rather than premature: every constituent claim in the row landed with its own tests across #227–#236, and the #234 coverage metrics keep future adapter gaps measurable instead of invisible — matching the repo's implementation-vs-evidence discipline. The next-step pointer correctly hands the ledger to #26 as a recovery input.

No findings. 683 tests passing, checks green. LGTM — merging, which closes #98.

PR #238, opened 2026-08-15
feat(replay): move snapshot boundaries to App Server

YoungJinJung:
Why

Removing redundant SessionStart and PostToolUse Hooks must not weaken snapshot and replay anchors. This moves their snapshot responsibility onto the authoritative App Server lifecycle before Hook removal begins.

What changed

- snapshot live thread starts as the baseline and completed Class B local operations as after-state boundaries
- recover thread-to-repository routing across daemon restarts
- honor snapshot_on_patch in the daemon-owned App Server ingestor
- document the migration boundary and cover baseline, mutation, restart, and disabled-config behavior

Validation

- ruff format --check .
- ruff check .
- mypy src tests
- pytest: 687 passed

Part of #86.


YoungJinJung:
## Review

Reviewed the full diff (ingestion.py, daemon.py, runtime_connection.py, docs, tests) against the durable-state rules, since this moves snapshot anchors.

- The anchor migration is additive, not weakening: `thread/started` (with a reported cwd) pins the baseline, completed Class-B-classified local operations pin after-state boundaries, and `PreToolUse` explicitly keeps the synchronous before-state during the Hook migration — so no existing replay anchor is lost while the redundant Hooks become removable. The lifecycle doc states exactly this boundary.
- Keying after-snapshots on the `expected_reversibility_class == \"B\"` marker composes cleanly with #235: the same downgrade that flags missing pre-mutation lineage now supplies the post-mutation anchor, and configured-MCP Class B writes participate too.
- The snapshot path honors the established discipline — `global_lock` around `snapshot_worktree` + record, `last_snapshot()` tree dedup, and a `SnapshotError` degrades to a snapshot-less record instead of dropping the event. The `snapshot_on_patch` load follows the daemon config pattern with a fail-visible fallback that defaults to snapshots *on* — the conservative direction.
- Restart safety is genuinely tested: the second ingestor instance recovers the thread→repository route from journal history before the mutation event arrives, and the test verifies actual git tree contents (`before`/`after`) rather than just snapshot presence. The disabled-config case pins the opt-out.

**Minor (non-blocking):**
- `snapshot_worktree` runs synchronously inside the async ingest path, so a large worktree's git snapshot briefly stalls event consumption. It's bounded (thread starts + B-mutation completions only) and matches the existing fsync-on-record pattern, but if #89's retention work ever profiles ingest latency, this is the first candidate for the same off-loop treatment control telemetry got.

Checks green (CI + CodeQL), 687 tests passing, architecture/lifecycle/status docs updated. LGTM — merging.

PR #239, opened 2026-08-15
feat(budget): version the review spend ledger

YoungJinJung:
Why

The mutable review-spend ledger enforces token ceilings but had no independent persisted schema. An older binary could otherwise reinterpret or overwrite a future ledger and silently reset the proof that a spending cap was reached.

What changed

- add the spotter.review_spend schema and schema version 1 to every new ledger write
- retain read-old/write-current support for legacy unversioned ledgers
- refuse unknown schema names, older unsupported versions, newer versions, and malformed versions without mutating the file
- fsync the temporary ledger and parent directory around atomic replacement
- document the durable contract and cover legacy upgrade plus non-destructive refusal

Validation

- ruff format --check .
- ruff check .
- mypy src tests
- pytest: 691 passed

Part of #47.


YoungJinJung:
## Review

Reviewed the full diff (budget.py, docs, tests).

- The versioning gate covers the whole input space conservatively: fully legacy files (both fields absent) read as v0 and upgrade lazily on the next successful mutation, while any partial or wrong combination — foreign schema name, missing name with a version, non-integer version, older or newer version — refuses via `LedgerCorrupt` before any mutation. The parametrized test verifies byte-identical file content after refusal, so an old binary can never reset a future budget, which is the exact #47 threat this addresses.
- Refusal composes correctly with the existing spend discipline: `reserve` degrades to a \"refusing to spend\" outcome rather than raising through the caller, so budget enforcement fails *closed* — no inference happens on an unreadable ledger — consistent with the check-before-spend rule already in the review path.
- The durability sequence is now complete: fsync the temp file, atomic `os.replace`, then fsync the parent directory — matching the commit/replace contract the lifecycle doc prescribes, and the doc gains the ledger's entry in the schema-version inventory.
- The legacy-upgrade test pins the full read-old/write-current round trip including the arithmetic (2+1 reviews, 50+25 tokens), not just the header fields.

No findings. Checks green (CI + CodeQL), 691 tests passing, lifecycle/status docs updated. LGTM — merging.

PR #240, opened 2026-08-15
feat(observability): version the source audit store

YoungJinJung:
## Why

The bounded App Server source audit was still an unversioned JSONL file. Older binaries could append samples to an unknown future container, making durable evidence ambiguous.

## What changed

- add explicit spotter.source_audit v1 container metadata
- atomically prefix valid legacy samples before the next append
- refuse corrupt or unknown containers without modifying their bytes
- keep audit schema failures visible while allowing primary App Server journaling to continue
- preserve metadata during bounded compaction and document the persistence contract

## Validation

- ruff format --check .
- ruff check .
- mypy src tests
- pytest (694 passed)

Part of #47.


YoungJinJung:
## Review

Reviewed the full diff (observability.py, ingestion.py, docs, tests).

- The operation ordering inside `record` is the load-bearing detail and it's right: `_refuse_unknown_container` runs *before* `_repair_tail`, so a foreign/future container's torn tail is never "repaired" (i.e. mutated by truncation) by an older binary — the future-schema test pins byte-identical content after a refused write, including a deliberately torn tail after valid future metadata.
- Legacy upgrade is validate-then-prefix: every legacy line must parse as a sample before the atomic `metadata + original bytes` replacement, and a corrupt legacy file refuses with zero modification (pinned by the not-json test). The upgrade preserves the original sample bytes rather than re-serializing them.
- Degradation scope matches the module's contract: `ObservabilityError` joins the caught set in `audit_source`, so an unusable audit container surfaces as a RuntimeWarning plus `last_source_audit_error` while primary App Server journaling and live-state reduction continue — the ingest-succeeds test proves the isolation.
- Bookkeeping stays consistent throughout: `_sample_count` excludes the metadata line everywhere (init, tail repair), compaction re-emits metadata ahead of the bounded sample window, `load` adjusts line numbers past the header so corruption reports still point at the right line, and `_replace` centralizes the fsync + atomic replace + parent-dir fsync sequence shared with #239's pattern.

**Minor (non-blocking):**
- A future container whose *very first* (metadata) line is itself torn mid-write parses as non-metadata, takes the legacy path, and gets tail-repaired away — an edge-of-an-edge that only an interrupted foreign writer could produce, and the sample data below would still refuse via `_parse_samples`. Not worth complicating the flow for.
- `record` reads the file up to three times per append (refuse/repair/ensure); bounded at 1000 samples, so irrelevant today.

Checks green (CI + CodeQL), 694 tests passing, lifecycle/status docs updated with the container contract. LGTM — merging.

PR #241, opened 2026-08-15
feat(feedback): version intervention records

YoungJinJung:
## Why

Intervention feedback was append-only and record-versioned, but it had no independent schema identity and writers did not validate existing history before appending. That allowed an older binary to extend unknown future or foreign feedback history.

## What changed

- write spotter.intervention_feedback schema identity and schema version on every new record
- keep released schema-name-less v1 feedback readable without rewriting it
- validate all existing history under the append lock before writing
- refuse future, foreign, mismatched, or corrupt records without changing the file
- document the durable feedback compatibility contract

## Validation

- ruff format --check .
- ruff check .
- mypy src tests
- pytest (698 passed)

Part of #47.


YoungJinJung:
## Review

Reviewed the full diff (feedback.py, docs, tests).

- The validation matrix is complete: both schema fields absent → released v1 legacy accepted as-is; wrong schema name, non-integer schema version, schema/record version mismatch, future version, and sub-1 version each refuse with a precise message — and the writer runs this over the *entire* existing history under the append lock before writing, so an older binary can no longer extend foreign or future feedback history. Both refusal tests verify byte-identical files.
- Legacy handling is the right shape for an already-released record format: old rows are read, never rewritten (the upgrade test pins the legacy row staying schema-less while the new append carries the current identity) — append-only history stays literally append-only, unlike the mutable-ledger lazy upgrade in #239, which is the correct distinction between the two store types.
- The `FeedbackError` re-raise guard inside the broad `except` tuple keeps precise schema messages from being rewrapped as generic \"unreadable\" — same pattern as the opportunities loader, and necessary since `FeedbackError` is a `ValueError`. The corrupt-history test confirms plain garbage still gets the line-numbered unreadable error.
- Moving the id filter out of the parse loop means the full file is validated even on filtered reads, so `explain` surfaces store corruption instead of silently skipping past it. The retained `SCHEMA_VERSION` alias keeps existing imports working.

**Minor (non-blocking):**
- Each append re-validates the whole history — O(n) per write, fine for human-generated feedback volume; worth revisiting only if this store ever grows machine-written rows.

Checks green (CI + CodeQL), 698 tests passing, lifecycle/status docs updated. LGTM — merging.

PR #242, opened 2026-08-15
feat(labels): guard versioned evidence history

YoungJinJung:
## Why

Evaluation labels already carried a record version, but had no independent schema identity and could be appended without validating existing history. An older binary could therefore extend future, foreign, or corrupt evidence.

## What changed

- add explicit spotter.label schema identity and schema version to new label records
- keep schema-name-less historical labels readable and immutable
- validate the full existing history under a per-file lock before every append
- fsync successful label appends
- refuse future, foreign, mismatched, or corrupt history without changing its bytes
- document the label compatibility contract

## Validation

- ruff format --check .
- ruff check .
- mypy src tests
- pytest (701 passed)

Part of #47.


YoungJinJung:
## Review

Reviewed the full diff (labels.py, docs, tests).

- Same validation matrix as the #241 feedback store, applied consistently: schema-name-less history (all released v0–v6 rows, including version-less v0 defaulting through `LEGACY_VERSION`) stays readable and immutable, while foreign schema names, non-integer schema versions, schema/record version mismatches, and future versions each refuse — and the append path validates the *entire* file under the new per-file `flock` before writing, so an older binary cannot extend future or corrupt evidence. Both refusal tests pin byte-identical files.
- The mixed-history test proves the right invariant for an already-released append-only format: the legacy row is never rewritten and the new append carries the current `spotter.label` identity — matching the feedback-store precedent rather than the mutable-ledger lazy upgrade, which is the correct choice per store type.
- The added `flock` + fsync on append is a strict hardening over the previous bare append: labels are the input to every published rate, so a torn or interleaved write here was the highest-value remaining gap in the #47 sweep. The lock also serializes the validate-then-append sequence so two concurrent labelers can't interleave between validation and write.
- `_load_label_history_path` keeps `load_label_history`/`load_labels` behavior identical for all readers; the newer-schema refusal test was updated to keep `schema_version` consistent with the mutated `version`, exercising the primary version gate rather than the mismatch gate.

**Minor (non-blocking):**
- As with #241, each append revalidates the whole per-session history — O(n) per label, trivially fine for human labeling volume.

Checks green (CI + CodeQL), 701 tests passing, lifecycle/status docs updated. LGTM — merging.

PR #243, opened 2026-08-15
feat(opportunities): guard versioned annotations

YoungJinJung:
## Why

Intervention opportunity annotations define the timing baseline for delay and effect metrics. They carried a record version but lacked an independent schema identity, locking, and pre-append history validation, so an older binary could extend future or corrupt evidence.

## What changed

- add explicit spotter.intervention_opportunity schema identity and schema version
- preserve schema-name-less released history without rewriting it
- validate the entire existing file under a per-session lock before append
- fsync successful annotation writes
- refuse future, foreign, mismatched, or corrupt history without changing its bytes
- document the annotation compatibility contract

## Validation

- ruff format --check .
- ruff check .
- mypy src tests
- pytest (705 passed)

Part of #47.


YoungJinJung:
## Review

Reviewed the full diff (opportunities.py, docs, tests).

- This completes the #47 sweep across the evidence stores with the exact pattern proven in #241 (feedback) and #242 (labels): per-record `spotter.intervention_opportunity` identity on new writes, schema-name-less released history readable and never rewritten, the full validation matrix (foreign name / non-integer / mismatch / future), and — the substantive hardening — a per-session `flock` that serializes validate-then-append with an fsync on success, so an older binary can't extend future or corrupt timing evidence and concurrent annotators can't interleave.
- The session-mismatch guard survives the loader refactor since `_load_opportunity_history_path` keeps the session parameter, and the existing per-line strict validation (window ordering, anchor typing, pre-v1 refusal) runs unchanged inside it — so the pre-append check inherits every integrity rule the reader already enforced, including the ones #211 established.
- The test suite mirrors the sibling PRs: current-write header pinning, legacy-read/current-write mixed history, both refusal parametrizations with byte-identical files, and corrupt-history refusal. Timing annotations are the baseline for the #24 delay metrics, so this was the right last store to guard.

No findings beyond the shared O(n)-per-append note already made on #241/#242 — trivially fine at human annotation volume.

Checks green (CI + CodeQL), 705 tests passing, lifecycle/status docs updated. LGTM — merging.

Issue #90, opened 2026-08-12
Implement runtime configuration and upgrade compatibility lifecycle


Bogyie:
### Problem / outcome

The current prototype loads a single TOML path/default config in short-lived processes. A standalone daemon introduces persistent configuration, reload/restart boundaries, CLI↔daemon↔Hook version windows, and Codex capability changes during upgrades. Persisted schema migration is already tracked in #47, but runtime compatibility and configuration lifecycle are not.

### Proposed direction

Define and implement the runtime-facing configuration/update contract:

- configuration layers: global, repository, and runtime override with explicit precedence
- classify settings as hot-reloadable, daemon-restart-required, or setup/reconfigure-required
- propagate/reload safe reviewer/signal/gate settings without producing mixed ambiguous state inside a turn
- add version/protocol handshakes across CLI, `spotterd`, and the PreToolUse bridge
- detect a daemon still running an older process image after package upgrade and provide a graceful restart/reconnect path
- combine runtime capability negotiation with Codex version changes instead of hard-coding a supported-version table where capabilities are observable
- make `spotter update` advisory when installed by a package manager rather than overwriting package-managed files
- integrate persisted config/manifest migrations with #47 without duplicating schema logic

### Implementation references — study, do not depend on

NVIDIA NeMo Relay has several useful production patterns for version/config/integration compatibility:

- persistent lifecycle and compatibility fingerprinting: https://github.com/NVIDIA/NeMo-Relay/blob/main/docs/nemo-relay-cli/plugin-installation.mdx
- Codex version checks, sidecar compatibility, generation fencing: https://github.com/NVIDIA/NeMo-Relay/blob/main/docs/nemo-relay-cli/codex.mdx

Relevant lessons:

- validate the selected host version before mutating integration state;
- do not silently reuse a long-lived process whose version/effective config differs from the current client;
- use a non-secret compatibility fingerprint rather than returning config/credential values from health endpoints;
- rotate an immutable installation generation on force-reconfigure;
- reject malformed/prerelease/too-old host versions consistently across install/doctor/run paths;
- doctor should identify modified, missing, disabled, duplicated, or incompatible owned integration state;
- configuration compatibility and process identity are different from mere socket/process liveness.

OpenClaw is a useful reference for Codex App Server runtime identity/capability capture:

- shared App Server client/process identity: https://github.com/openclaw/openclaw/blob/main/extensions/codex/src/app-server/shared-client.ts
[...]

PR #244, opened 2026-08-15
feat(config): version configuration schema

YoungJinJung:
## Why

Configuration syntax is a durable contract and must be distinguished from package, IPC, and runtime activation versions. Without an explicit schema gate, a binary could activate settings written for an incompatible future format.

## What changed

- define independent spotter.config schema identity and version
- mark the reference TOML with the current schema
- retain schema-less released configuration as explicit legacy v0
- expose the loaded schema version on SpotterConfig
- reject partial, foreign, older-explicit, malformed, and future schema declarations before parsing settings
- document that syntax compatibility remains separate from #90 reload/restart/reconfigure boundaries

## Validation

- ruff format --check .
- ruff check .
- mypy src tests
- pytest (712 passed)

Part of #47.


YoungJinJung:
## Review

Reviewed the full diff (config.py, spotter.example.toml, docs, tests).

- The gate runs as the *first* statement of `from_mapping`, so an incompatible declaration refuses before any setting is parsed, let alone activated — matching the stated threat model. The rejection matrix is complete: future version, explicit v0 (which no binary ever wrote), foreign schema name, non-integer, boolean, and both partial-declaration shapes each land on a precise `ConfigurationError`, all parametrized in tests.
- Schema-less files map to an explicit `LEGACY_CONFIG_SCHEMA_VERSION` rather than being silently treated as current — the loaded version is exposed on `SpotterConfig` so diagnostics can distinguish legacy from declared configs, and the reference TOML now declares v1 with the example-load test pinning it.
- Downstream composition is right without extra changes: the daemon's config readers (`_configured_reviewer`/`_configured_mcp_semantics`/`_configured_snapshot_on_patch`) all route through `from_toml` and already catch `ValueError` with a visible stderr fallback to conservative defaults — so a future-schema config degrades the daemon loudly instead of activating misread settings.
- The lifecycle doc correctly scopes this as syntax compatibility, distinct from the #90 reload/restart activation boundaries, and the status row reflects it.

One inherent caveat worth stating (not a finding): binaries released *before* this PR ignore the new top-level keys entirely, so the gate protects forward from here — that's unavoidable when introducing versioning and exactly why landing it now, at v1, is the right time.

712 tests passing, docs updated. Will merge once the remaining CI jobs are green.

PR #245, opened 2026-08-15
feat(integration): version ownership manifests

YoungJinJung:
## Why

The integration manifest is the authority for setup, rollback, and teardown. Its released numeric version supported legacy migrations, but it lacked an independent schema identity and its atomic replacements did not fsync temporary content or the parent directory.

## What changed

- add spotter.integration_manifest schema identity and explicit schema version while retaining the numeric schema field
- validate foreign, malformed, and mismatched declarations before ownership data is used
- map schema 1-3 manifests to the current identity during the existing in-memory migration
- fsync temporary content and parent directories for manifest and owned host-state replacements
- document the ownership migration and crash-durability contract

## Validation

- ruff format --check .
- ruff check .
- mypy src tests
- pytest (715 passed)

Part of #47.


YoungJinJung:
## Review

Reviewed the full diff (integration.py, docs, tests).

- The dual-identity design is right for a manifest with an already-released numeric version: name/version absent → the existing numeric path runs unchanged (1–3 in-memory migration, v4 acceptance, newer-numeric refusal), while a declared identity must be exactly consistent — foreign name, non-integer version, and name/numeric mismatch each refuse without touching the file. A genuinely future manifest (schema 5 + schema_version 5) passes the consistency checks and still lands on the existing `unsupported` numeric refusal, so both gates compose without a hole.
- Stamping the current name/version into `raw` before construction gives the correct mutable-authority semantics (like the #239 ledger): legacy manifests upgrade in memory and persist the full identity on the next successful atomic write, pinned end-to-end by the schema-3 upgrade test.
- The `_atomic_write` hardening is more than fsync: switching from `write_bytes` + `chmod` to `os.open(..., 0o600)` also closes the brief window where the temp file existed with default-umask permissions — a small but real improvement for an ownership file — and the temp-fsync/replace/dir-fsync sequence matches the #239/#240 durability contract for both the manifest and owned host-state writes.
- Test coverage is thorough: persisted header pinning after setup, the three named-schema refusals with byte-identical files, and all three legacy-migration tests updated to strip the name fields so they exercise the true legacy path.

No findings. Checks green (CI + CodeQL), 715 tests passing, lifecycle/status docs updated. LGTM — merging.

PR #246, opened 2026-08-15
feat(journal): validate versioned state caches

YoungJinJung:
## Why

Trace records carried a numeric version, but the unversioned size-only state sidecar could let a writer skip full journal validation. A matching cache written alongside a future journal could therefore allow an older binary to append into unsupported history.

## What changed

- add explicit spotter.trace_event identity while retaining the released v field for mixed history
- add a versioned spotter.journal_state cache tied to the journal schema version
- treat legacy, corrupt, or unsupported sidecars as rebuildable caches and reconstruct them from the authoritative journal
- validate the journal before append whenever a cache cannot prove current compatibility
- fsync atomic sidecar replacements and reject non-object journal records explicitly
- prove a matching future sidecar cannot hide a future journal or permit mutation

## Validation

- ruff format --check .
- ruff check .
- mypy src tests
- pytest (718 passed)

Part of #47.


YoungJinJung:
## Review

Reviewed the full diff (snapshot.py, docs, tests) — this is the most safety-critical store in the #47 sweep, so I focused on the cache-bypass threat.

- The stated hole is genuinely closed: previously a size-matching sidecar skipped full journal validation on append, so a future journal could be extended by an old binary. Now `_load_state_cache` demands a fully consistent current claim (state schema + version + `journal_schema_version` + size + typed counters), anything less is treated as a rebuildable cache, and the rebuild path runs `load(repair_tail=True)` whose `_record_version` gate refuses the future journal before any byte is appended — the hide-behind-cache test pins byte-identical journal *and* sidecar after refusal.
- The sidecar keeps its correct epistemic role: it is only ever an index. Unsupported/corrupt sidecars are silently discarded and rebuilt from the authoritative journal (pinned by the rebuild test), and `last_snapshot()` degrades to `None` on an invalid cache, which merely costs snapshot dedup — never correctness.
- Trace records gain the dual identity consistently with the sibling PRs: released `v` stays authoritative for mixed history, declared `schema`/`schema_version` must be exactly consistent, and non-object records now fail with an explicit `SnapshotError` instead of an accidental attribute error. `_write_state_cache` picks up the full 0o600 + fsync + dir-fsync sequence, and the SIGKILL crash test was correctly retargeted at the new writer.

**One boundary worth stating (not a finding):**
- The cache defends against honest version skew, not forged caches: a writer that stamped `journal_schema_version: 1` while writing v2 rows would be lying, and no sidecar design can defend the journal against an adversarial co-writer — the journal's own validation remains the authority whenever the cache can't prove itself, which is the right and documented contract.

Checks pending at review time — will merge once green. 718 tests passing, lifecycle/status docs updated. LGTM.

PR #247, opened 2026-08-15
feat(sampling): version durable signal samples

YoungJinJung:
## Why

Issue #47 requires every durable schema family to have an independent identity and to refuse mutation of unknown future data. Signal-sampling records only carried a generic version and validated outside any append lock.

## What changed

- persist the independent spotter.signal_sampling schema identity and version on batch and sample records
- retain read-old/write-current support for schema-less v1 history
- validate the complete history and append under one per-session lock
- refuse future, foreign, mismatched, corrupt, or ambiguous records before append
- flush and fsync successful appends
- document the durable compatibility contract and current implementation state

## Validation

- ruff format --check .
- ruff check .
- mypy src tests
- pytest (720 passed)

Part of #47.


YoungJinJung:
## Review

Reviewed the full diff (sampling.py, docs, tests).

- The established #47 matrix lands consistently on both record types (batch and sample): schema-less released v1 rows stay readable and immutable (mixed-history test), and foreign name / non-integer / mismatch / future each refuse with byte-identical files — this store already validated on every load thanks to its strict batch/sample cross-checks, so wrapping the load in the pre-append path completes the refuse-before-mutate guarantee.
- The single per-session `flock` around the *entire* sample flow (load → overlap-conflict check → frame computation → append) is a bonus beyond schema hardening: it closes the pre-existing TOCTOU where two concurrent samplers could both pass the overlapping-strata check and then interleave conflicting frames. Determinism of the Bernoulli draw made that mostly benign, but the disjoint-suffix invariant is now actually enforced rather than assumed.
- The batched fsync after writing the batch row plus all its samples composes correctly with the existing loader integrity checks: a crash mid-append leaves either a torn line (refused as unreadable) or a selected-count/orphan mismatch (refused by the cross-validation), so partial frames can never be silently consumed.
- Docs updated across architecture/lifecycle/status, and the schema-version inventory gains the sampling entry.

No findings. Checks green, 720 tests passing. LGTM — merging.

PR #248, opened 2026-08-15
feat(experiment): version result journals

YoungJinJung:
## Why

Issue #47 requires experiment results to keep an independent durable schema identity and to remain non-destructive when a newer or foreign writer has touched the shared history. Counterfactual rows carried only result_schema_version, and writers appended without validating prior records.

## What changed

- persist spotter.experiment_result identity/version on meta, arm, and completion records
- retain bounded reads of schema-less v1-v3 history, including legacy completion rows
- validate history under a per-result-file lock before every fsynced append
- refuse future, foreign, mismatched, non-object, invalid-UTF-8, and torn history before mutation
- validate explicit schema identity in objective outcome projections
- surface incompatible histories as normal experiment CLI failures
- document the compatibility and interrupted-append contract

## Validation

- ruff format --check .
- ruff check .
- mypy src tests
- pytest (725 passed)

Part of #47.


YoungJinJung:
## Review

Reviewed the full diff (experiment.py, runtime_metrics.py, cli.py, docs, tests).

- The writer side lands the full #47 pattern on all three record types (meta, arm, completion) with a nice extra: the incompatible-history refusal fires *before* any fork is prepared — the test's sentinel fork asserts exactly that ordering, so a poisoned result file can't cost worktree setup before failing. Per-file flock serializes validate-then-append, and each append fsyncs.
- Legacy handling threads a genuinely tricky needle: released completion rows carried *no* version at all, and the `legacy_completion` carve-out grandfathers exactly that shape (all three fields absent + `complete: true`) while any other version-less row still refuses — the round-trip test strips history down to true legacy form and proves mixed files read/write correctly.
- Torn history now *blocks* further appends (unreadable-line refusal), and the architecture doc was updated to say so explicitly — the right call for experiment evidence, where extending past an interrupted append could silently attach rows to the wrong run.
- The reader side gains matching validation scoped to experiment rows (`experiment_id` present), with the `{1,2,3}` version gate relocated from `_objective_arm` into it — same acceptance surface, earlier and more precise errors, and frozen-task rows stay untouched. The foreign-schema refusal and current-schema acceptance are both pinned.
- CLI surfaces `ExperimentResultError` as a normal experiment failure alongside the existing error family.

No findings. 725 tests passing, architecture/status docs updated. Will merge once the remaining CI job reports.

PR #249, opened 2026-08-15
feat(corpus): version frozen task artifacts

YoungJinJung:
## Why

Issue #47 requires task fixtures, task sets, and their result batches to evolve as separate durable contracts. They had independent numeric fields but no schema-family identity, and batch resume could run preflight or truncate a torn tail before rejecting incompatible history.

## What changed

- define independent spotter.task, spotter.task_set, and spotter.task_batch v1 identities
- update bundled corpus manifests and their pinned hashes while retaining schema-name-less v1 reads
- stamp batch metadata, arm, and completion records with current identity/version
- validate batch history under a per-file lock before preflight and before every fsynced append
- refuse future, foreign, mismatched, corrupt, duplicate, or ambiguous batch rows without mutation
- repair only a validated torn final row via fsynced temporary content, atomic replace, and directory fsync
- validate current task-batch identity in objective outcome projections
- document the three compatibility boundaries

## Validation

- validated dev-v1, dev-v2, validation-v1, and validation-v2 corpus manifests
- ruff format --check .
- ruff check .
- mypy src tests
- pytest (732 passed)

Part of #47.


YoungJinJung:
## Review

Reviewed the full diff (task_corpus.py, runtime_metrics.py, corpus manifests, docs, tests).

- The ordering bug called out in the description is genuinely fixed, and carefully: resume now validates the batch history *before* preflight runs, and inside `_read_task_batch_locked` the torn-tail repair is deferred until after every row has passed schema validation — so an incompatible or foreign batch can no longer trigger worktree preflight or have its tail truncated before being refused. The append path goes further and refuses a torn tail outright (`repair_tail=False`), which is right since mid-run appends should never encounter one.
- `_append_task_batch` adds integrity guards beyond the schema family: duplicate (task_id, arm) results and double completion rows now refuse under the same lock that serializes validate-then-append — both were previously possible via a racing resume.
- `_replace_task_batch` is the most careful replacement writer in the whole #47 sweep: unique temp name, exclusive-create open, fsync + atomic replace + directory fsync, and temp cleanup on failure.
- The three families get appropriately different treatment: TOML manifests (task/task_set) accept a missing schema name for released files but require exactness when present, with the bundled corpus manifests updated and their pinned hashes regenerated (all four sets validated per the description); JSONL batch rows follow the strict per-record pattern with the same legacy-completion carve-out as #248, and the objective-outcome reader gains the matching task-row gate keyed on `run_id` with accept/reject tests.

No findings. Checks green (CI + CodeQL), 732 tests passing, all three compatibility boundaries documented. LGTM — merging.

PR #250, opened 2026-08-15
feat(replay): version fork manifests

YoungJinJung:
## Why

Issue #47 requires snapshot/replay metadata to have an independent durable identity and forbids older binaries from overwriting newer lineage. Fork manifests supported numeric v1-v6 compatibility and atomic writes, but had no schema-family name or validate-before-replace lock.

## What changed

- add the independent spotter.fork_manifest schema identity
- stamp CREATING, READY, and FAILED v6 manifests with current identity
- retain schema-name-less v1-v6 reads and their historical coverage defaults
- reject foreign, future, non-integer, non-object, invalid-UTF-8, and corrupt manifests
- validate an existing manifest under a per-resource lock before fsynced atomic replacement
- refuse incompatible replacement with byte-identical preservation
- document the lineage compatibility boundary

## Validation

- ruff format --check .
- ruff check .
- mypy src tests
- pytest (733 passed)

Part of #47.


YoungJinJung:
## Review

Reviewed the full diff (replay.py, docs, tests).

- The reader keeps the full bounded v1–v6 compatibility while adding the identity gate with the right edge semantics: a manifest *without* the `schema` key reads as legacy, but an explicit foreign (or null) value refuses — plus the new non-object, non-integer-version, and invalid-UTF-8 rejections, all funneled through the existing `ReplayError` wrap. The legacy-compat test now pops the schema key, so it exercises the true schema-less path.
- `_write_fork_manifest` guards both directions: a writer-side assertion refuses to persist anything but the current identity (catching programming errors constructing stale manifests), and the validate-before-replace under the new per-resource lock means an older writer can never overwrite future, foreign, or corrupt lineage — the replacement test pins byte-identical preservation for both incompatible shapes, driving through the real `fork()`-produced manifest rather than a synthetic one.
- The overwrite-refusal semantics are safe for the actual lifecycle: fresh fork ids write to non-existent paths (no validation needed), and same-path replacement only occurs for CREATING→READY/FAILED transitions within one run, where the pre-existing atomic O_EXCL temp + fsync + dir-fsync sequence (now serialized under the lock) makes a corrupt intermediate essentially unreachable.
- All three written states (CREATING, FAILED, READY) stamp the identity, so a crash at any transition leaves a currently-identified manifest, and the docs record the lineage boundary in architecture, lifecycle, and the schema-version inventory.

No findings. Checks green (CI + CodeQL), 733 tests passing. LGTM — merging.

PR #251, opened 2026-08-15
docs(schema): record compatibility matrix

### LINKED ISSUE #47: Version persisted schemas and define migrations

YoungJinJung:
## Why

The #47 implementation sweep now covers every durable Spotter-owned family. A final explicit matrix makes each independent compatibility boundary and refusal policy auditable, and records why native service definitions and ephemeral reviewer schemas are not durable evidence formats.

## What changed

- inventory all 14 durable schema families and their current identities
- state supported-old, migration/rebuild, and future/foreign/corrupt behavior per family
- document external-effect and ThreadState ownership by Trace IR
- keep package/protocol/generation axes explicitly separate
- update current status to reflect completed durable schema coverage

## Validation

- ruff format --check .
- ruff check .
- mypy src tests
- pytest (733 passed)

Closes #47.


YoungJinJung:
## Review

Reviewed the full diff (docs-only), and cross-checked every matrix row against the implementations merged in #239–#250, all of which I reviewed individually.

**Accuracy — verified row by row:**
- All 14 family rows correctly state what actually landed: the per-store nuances are right where they differ — the journal-state sidecar *rebuilds* rather than refuses (#246), the review-spend ledger *lazily upgrades* on mutation as a mutable authority (#239) while the append-only evidence stores never rewrite legacy rows (#241–#243, #247), the experiment matrix includes the version-less legacy completion carve-out (#248), the task batch's "refuse preflight, repair, and append" captures the #249 ordering fix, fork lineage says "refuse replacement" (#250), config says "refuse activation" (#244), and the source audit degrades visibly while preserving primary journals (#240).
- The exclusions paragraph earns its place: stating *why* ThreadState (rebuilt from Trace IR), host-native service definitions, and reviewer scratch schemas are not durable evidence families prevents the matrix from being read as incomplete, and keeping package/protocol/generation axes explicitly separate closes the aliasing confusion #47 originally called out.
- The status-row update from the single config claim to the completed-sweep claim is now backed by fourteen merged, individually tested implementations — this is the rare status flip where the evidence genuinely precedes the assertion.

No findings. Checks green, 733 tests passing. LGTM — merging, which closes #47.

PR #252, opened 2026-08-15
feat(lifecycle): record owned repository resources

YoungJinJung:
## Why

Repository-aware purge cannot safely infer ownership from `refs/spotter/*` names or worktree paths. Spotter needs durable, exact provenance at resource-creation time before #89 can enumerate or remove anything.

## What changed

- add a versioned, locked, fsync-backed `repos.json` repository registry
- identify repositories by Git common-directory device/inode while retaining the last known path
- record exact snapshot refs and detached restore worktrees with immutable resource generations and expected targets
- refuse unknown/corrupt registry state before creating durable Git resources
- roll back a newly created ref/worktree through Git if its ownership record cannot be committed
- document the implemented foundation and keep purge/reachability explicitly pending

Part of #89.

## Validation

- `ruff format --check .`
- `ruff check .`
- `mypy src tests`
- `pytest` (739 passed)


YoungJinJung:
## Review

Reviewed the full diff (new repository_registry.py, snapshot.py, paths.py, docs, tests).

- The ownership model is built on the right primitives: identity by Git common-directory device/inode survives a repository move (pinned by test) while a different repository recreated at the old path gets a separate entry (also pinned) — exactly the inference #89 purge must never make from names or paths. Records are only written after Git confirms reality: `record_ref` requires the ref to be in the `refs/spotter/` namespace *and* resolve to the expected target, and `record_worktree` proves the worktree belongs to the claimed repository by common-dir identity.
- The creation-rollback invariant holds in both directions: a newly created ref whose ownership record cannot be committed is deleted with the old-value guard (`update-ref -d ref sha`, race-safe) and only when it did not pre-exist, while a refused worktree is removed through Git — the future-schema tests verify both the empty `refs/spotter/` namespace and the absent worktree. A crash *between* creation and record can still orphan a resource, which is the irreducible gap #89's reachability pass exists to close, and the docs say so.
- Idempotent re-record keeps the registry honest: an existing resource with the same target is a no-op, while a target mismatch refuses rather than silently re-owning a changed resource. The registry file itself follows the full #47 discipline — versioned, strictly validated (closed resource types, owner allowlist, duplicate detection), locked read-modify-write, fsynced atomic replacement, and non-destructive refusal of unknown schemas.
- The failure-mode change (registry corruption now blocks snapshot creation via `SnapshotError`) composes correctly with existing callers: the App Server ingest path already degrades to a snapshot-less record, and refusing to create untracked durable resources is the stated policy.

**Minor (non-blocking):**
- Every distinct snapshot ref appends a resource entry and each record rewrites the whole `repos.json` — linear file growth and O(n) rewrites per new snapshot, on the snapshot creation path. Tree dedup and same-target idempotency bound it in practice, but long-lived sessions will make this the first thing #89's retention should compact (a natural companion to the existing ponytail note about checkpointing reconciliation state).

Checks green (CI + CodeQL), 739 tests passing, lifecycle/status docs updated with the pending purge boundary kept explicit. LGTM — merging.

PR #253, opened 2026-08-15
feat(lifecycle): preview repository purge

YoungJinJung:
## Why

#89 requires purge to be inspectable before it can be destructive. The repository registry now records exact ownership, but users still need a conservative way to see which recorded resources remain verifiable, inaccessible, or changed.

## What changed

- add read-only `spotter purge --all --dry-run`
- add `--json` machine-readable preview output
- revalidate repository common-directory identity, ref targets, and worktree administrative paths
- distinguish `SAFE_OWNED`, `INACCESSIBLE`, and `AMBIGUOUS` resources plus present/absent/unknown state
- return non-zero for incomplete or ambiguous ownership
- refuse destructive purge and unrelated positional/options
- document the preview boundary in English, Korean, and Chinese user guides

Destructive scopes and reachability-aware retention remain part of #89.

## Validation

- `ruff format --check .`
- `ruff check .`
- `mypy src tests`
- `pytest` (746 passed)


YoungJinJung:
## Review

Reviewed the full diff (repository_registry.py inspect path, cli.py, tri-lingual docs, tests).

- The inspection matrix is conservative in every cell: a missing repository path is INACCESSIBLE, a non-repo or different-inode repository at the recorded path is AMBIGUOUS, a changed ref target is AMBIGUOUS even though present, and only exact identity matches earn SAFE_OWNED. The one nuanced call — an absent recorded ref reported as SAFE_OWNED/ABSENT — is right, since future purge of it is a no-op, not a risk.
- The worktree tri-state is well designed: absent path cross-checks both `git worktree list` registration and the recorded administrative directory — both gone is safely absent, both remaining is safely cleanable stale metadata, and *disagreement* is ambiguous (pinned by the dedicated test that removes the worktree then recreates the admin dir). Present paths re-verify common-dir identity and the absolute git dir, mirroring the #252 recording checks.
- The CLI fencing is airtight for a command whose name implies deletion: `--all --dry-run` is mandatory, `--apply`/positional targets/`--portable` refuse, `--all`/`--json` are rejected on every other command, and the docs in all three languages say deletion is not implemented. `inspect()` performs only read-only git queries — no `update-ref`, no `worktree remove`, no registry write — and the first test asserts the ref and worktree survive the preview.
- The non-zero exit for any non-SAFE_OWNED resource makes the preview scriptable as a gate, and the JSON payload embeds `deletion_supported: false` so tooling can't mistake the preview for an execution plan.
- Scope discipline holds: destructive scopes and reachability-aware retention stay explicitly pending in #89, matching the lifecycle doc's safe-order section.

No findings. Checks green (CI + CodeQL), 746 tests passing. LGTM — merging.

PR #254, opened 2026-08-15
feat(lifecycle): retain reachable snapshots

YoungJinJung:
## Why

Exact ownership is not enough to make a snapshot deletable. #89 requires retention to preserve artifacts still reachable from recovery/evaluation lineage before applying age or size policy.

## What changed

- build a conservative snapshot reachability index for purge preview
- retain snapshots referenced by journals, schema-valid fork manifests, or live registered worktrees
- match journal roots by snapshot identity so repository moves cannot orphan lineage
- surface roots as a separate `REFERENCED` group in human and JSON previews
- make unreadable journals/manifests produce unknown/ambiguous retention instead of an unreferenced guess
- fail the preview when referenced data is already absent
- document the implemented roots and the remaining evaluation backreference/pin/recovery work

Part of #89.

## Validation

- `ruff format --check .`
- `ruff check .`
- `mypy src tests`
- `pytest` (750 passed)


YoungJinJung:
## Review

Reviewed the full diff (new retention.py, cli.py, repository_registry.py, tri-lingual docs, tests).

- The retention lattice errs conservative at every joint: a known root always wins (retaining is already sufficient even when other sources are unreadable), no-root-plus-any-unreadable-source degrades to UNKNOWN and groups as AMBIGUOUS (pinned by the corrupt-manifest test), and only a cleanly-unreferenced snapshot lands in SAFE_OWNED. A snapshot missing from the index entirely gets an explicit UNKNOWN diagnostic rather than a default.
- Matching journal roots by snapshot identity rather than recorded path is the right trade and the code comment owns it: over-retaining an identical commit in another repository is acceptable, silently orphaning lineage after a repository move is not — and the move test proves the reference survives while the registry follows the inode.
- The three implemented roots compose cleanly: journals via the sessions scan, fork manifests through the schema-validated `load_fork_manifest` (so a future/foreign manifest degrades to a diagnostic instead of being parsed loosely), and only SAFE_OWNED + PRESENT worktrees anchor their source snapshot — an ambiguous worktree can't launder a snapshot into REFERENCED.
- The exit-code refinement is exactly right: REFERENCED-and-present is a healthy zero (both new happy-path tests), while REFERENCED-but-absent fails the preview because reachable lineage already missing is precisely what an operator must see before any destructive scope exists. JSON consumers keep the raw `confidence`/`retention`/`references`/`retention_diagnostics` fields even though the human grouping folds retention-UNKNOWN into AMBIGUOUS.
- Threading loaded `entries` into `inspect()` avoids the double registry read without changing the existing call contract, and the pending roots (evaluation backreferences, pins, pending recovery) stay explicitly documented as #89 work.

**Minor (non-blocking):**
- `retention_for` is computed three times per resource when building the JSON payload; a single local would tidy it, at dict-lookup cost today.

Checks green (CI + CodeQL), 750 tests passing. LGTM — merging.

PR #255, opened 2026-08-15
feat(lifecycle): retain experiment snapshot roots

YoungJinJung:
## Why

Durable experiment results can be the last remaining backreference to a fork manifest and its source snapshot. Purge preview must not classify that lineage as deletable.

Part of #89.

## What changed

- scan durable experiment JSONL for `fork_manifest` backreferences
- retain snapshots rooted by valid referenced manifests, including manifests outside the current manifest directory
- classify missing, invalid, or unreadable evaluation backreferences as unknown/ambiguous instead of deletion candidates
- document evaluation-result reachability in lifecycle/status and localized user guides

## Validation

- `ruff format --check .`
- `ruff check .`
- `mypy src tests`
- `pytest` (752 passed)


YoungJinJung:
## Review

Reviewed the full diff (retention.py, tri-lingual docs, tests).

- The new root follows the same conservative policy proven in #254 at every failure point: an unreadable results file, a corrupt row, an invalid `fork_manifest` value, or a missing/unloadable referenced manifest all flow through `_record_unknown` and turn otherwise-unreferenced snapshots into UNKNOWN/AMBIGUOUS rather than deletion candidates — the missing-manifest test pins the non-zero exit and the diagnostic surfacing.
- Referenced manifests load through the schema-validated `load_fork_manifest`, so the #250 identity gates apply to backreferenced manifests too, and supporting manifests *outside* the current manifest directory (with relative paths resolved against the result file) is exactly what makes archived evaluation lineage safe — the archived-manifest test proves a result row alone keeps the source snapshot REFERENCED.
- Rows without a `fork_manifest` field skip silently, which is right since most experiment rows carry no manifest backreference; only present-but-invalid values raise the conservative flag. The reference string (`experiment_result:<relpath>:line:N:fork:<id>`) gives operators an exact provenance pointer.
- The docs close the loop honestly: evaluation backreferences move from the pending list into the implemented roots across lifecycle/status and all three user guides, leaving pins and pending-recovery as the explicitly remaining #89 roots.

**Minor (non-blocking):**
- The fork-manifest-directory loop still records its diagnostics with an inline per-key loop while the new code uses `_record_unknown` — worth unifying next touch. Full experiment-tree scans per preview are fine for a manual command.

752 tests passing, and the `_fork_manifest` test-fixture dedup is a nice cleanup. Will merge once CI is green.

PR #256, opened 2026-08-15
feat(lifecycle): retain recovery checkpoints

YoungJinJung:
## Why

A durable journal can name the pre-mutation recovery checkpoint only in the event payload. Purge and prune must treat that checkpoint as a live root rather than considering its ref orphaned.

Part of #89.

## What changed

- include journal `checkpoint` payloads in snapshot reachability alongside record snapshots
- reject malformed checkpoint values conservatively before prune or purge can delete refs
- cover recovery roots in both local prune and repository-aware purge preview
- update lifecycle/status and localized user guides

## Validation

- `ruff format --check .`
- `ruff check .`
- `mypy src tests`
- `pytest` (755 passed)


YoungJinJung:
## Review

Reviewed the full diff (snapshot.py, docs, prune + purge tests).

- Extending `snapshot_references` is the right single point of change: both consumers (local `prune` and the repository-aware purge preview) inherit the recovery root simultaneously, and both are pinned — prune retains the checkpoint-referenced ref while still dropping the genuine orphan, and the purge preview groups it REFERENCED with the exact `journal:<session>:step:<n>` provenance string.
- The validation posture matches the existing prune contract: a present-but-malformed `checkpoint` value (non-string or empty) raises `SnapshotError` and aborts the whole enumeration rather than guessing, exactly like the pre-existing unreadable-journal abort — a corrupt claim must never let a ref look orphaned. Absent checkpoints skip silently, and the roots set correctly dedups the record-snapshot-equals-checkpoint case.
- The provenance filter (cwd/target matching with unknown-provenance retention) applies to both root kinds identically, so checkpoint roots get the same move-safe, over-retain-not-lose semantics established in #254.
- The docs honestly narrow the pending #89 list to manual pins and destructive scopes across lifecycle/status and all three user guides.

No findings. Checks green (CI + CodeQL), 755 tests passing. LGTM — merging.

PR #257, opened 2026-08-15
feat(lifecycle): add manual snapshot pins

YoungJinJung:
## Why

Retention needs an explicit operator-controlled root for snapshots that must survive even when no journal or experiment currently references them. Those roots must not weaken ownership checks or be bypassed by local prune expiry.

Part of #89.

## What changed

- add schema-versioned, locked, fsynced manual snapshot-pin storage
- add `spotter pins add|list|remove`, accepting only exact registered Spotter-owned snapshots
- retain manual roots in repository-aware purge preview and classify corrupt/future/stale pin state conservatively
- protect pins from ordinary prune and explicit `--max-age-days` expiry
- add stable repository lookup by Git identity and document the new lifecycle contract/schema

## Validation

- `ruff format --check .`
- `ruff check .`
- `mypy src tests`
- `pytest` (760 passed)


YoungJinJung:
## Review

Reviewed the full diff (new snapshot_pins.py, cli.py, retention.py, snapshot.py, repository_registry.py, docs, tests).

- The `pins add` validation chain is appropriately strict: the repository must have an ownership record (via the new identity-based `find_repository`, which refuses duplicate identities), the SHA must be an *exactly owned* `git_ref` target in that entry, and `record_ref` re-verifies the ref still resolves to the SHA under the global lock before the pin lands — so neither an arbitrary commit (pinned by the unowned-HEAD test) nor a drifted snapshot can become a manual root. Pins scope by `registry_entry_id`, so a pin in one repository never protects a SHA in another.
- The prune semantics distinguish the two retention strengths correctly: referenced snapshots can still expire under `--max-age-days`, but `protected` pins short-circuit before both the unreferenced and expiry branches — the survive-prune-and-expiry test drives it through the real CLI with an aggressive age policy, plus a direct far-future unit check.
- Purge retention handles pin-store failure modes with the established conservatism: a future/corrupt store degrades every snapshot to UNKNOWN/AMBIGUOUS with a diagnostic, an orphan pin referencing no registered snapshot raises a hard preview failure when there is nothing to degrade (the no-registry test), and a matched pin appears as an auditable `manual_pin:<id>` reference that disappears cleanly on `pins remove` (the full add→list→preview→remove→preview round trip is tested).
- The store itself carries the complete #47 discipline — schema identity, version gate, strict field validation, duplicate pin-id *and* duplicate root rejection, idempotent add returning the existing pin, and locked fsynced atomic replacement with temp cleanup.

No findings. Checks green (CI + CodeQL), 760 tests passing, docs updated across lifecycle/status and the three user guides — leaving destructive scopes as the sole remaining #89 item. LGTM — merging.

PR #258, opened 2026-08-15
feat(lifecycle): add snapshot purge scope

YoungJinJung:
## Why

Purge preview can now prove repository ownership and reachability, but operators still need an explicit Git-aware destructive scope. Snapshot cleanup must remove worktrees before refs, re-check roots after partial failures, and never guess through ambiguous ownership.

Part of #89.

## What changed

- add `spotter purge --snapshots` with `--dry-run` and JSON/human per-resource outcomes
- remove exact safe-owned worktrees through Git, recompute reachability, then delete refs with expected-target compare
- retain refs when worktree cleanup fails and classify ambiguous worktree state conservatively
- continue across inaccessible repositories while reporting partial success/failure honestly
- keep ownership records as idempotent audit evidence (`already_absent` on rerun)
- preserve `--all` as preview-only while remaining destructive scopes are pending

## Validation

- `ruff format --check .`
- `ruff check .`
- `mypy src tests`
- `pytest` (765 passed)


YoungJinJung:
## Review

Reviewed the full diff (cli.py, retention.py, docs, tests) adversarially — this is the first PR in the series that actually deletes.

**Correctness / safety — holds up under scrutiny:**
- The load-bearing subtlety is handled exactly right: `exclude_worktrees` removes a to-be-purged worktree's *own* self-retention root from the reachability pass, so a worktree and its source ref can go in one run — while a worktree in any non-SAFE_OWNED state instead contributes a diagnostic that pushes its ref to AMBIGUOUS, blocking deletion (pinned by the disagreeing-metadata test's `skipped_ambiguous`).
- The remove-worktrees → **re-inspect → recompute reachability without exclusion** → delete-refs sequence makes the failure invariant structural rather than bookkept: a failed worktree removal means the still-present worktree re-roots its ref on the second pass, so the ref lands `skipped_referenced` naturally — the monkeypatched-failure test pins both the outcomes and the survival of both resources.
- Ref deletion uses `update-ref -d ref <expected-target>`, so a ref that moved between inspection and deletion fails the compare instead of deleting the wrong commit, and the whole apply path holds the global snapshot lock — the same lock that guards snapshot creation and hook journal appends — so no new reference can land mid-purge.
- Partial-failure honesty is real: an inaccessible repository degrades to `skipped_ambiguous` without undoing the healthy repository's completed removal (tested), any failure or ambiguity yields a non-zero exit, and the outcome vocabulary distinguishes `failed_retryable` (worktree, safe to re-run) from `failed_manual_recovery` (ref state uncertain after a failed compare-and-delete) — the right escalation split.
- CLI fencing stays strict: exactly one scope, destructive `--all` still refused, `--apply` refused with an explanatory message, idempotent re-run reports `already_absent` while ownership records persist as audit evidence, and `deletion_supported` in JSON is now truthfully scope-dependent.

No findings. 765 tests passing, lifecycle/status and all three user guides updated with the remaining data/log/integration/all scopes explicitly pending. Will merge once CI is green.

PR #259, opened 2026-08-15
feat(lifecycle): record owned log resources

YoungJinJung:
## Why

Issue #89 requires destructive log cleanup to rely on explicit ownership rather than filename patterns. Spotter currently creates daemon and reviewer logs without durable proof that those exact files belong to Spotter.

## What changed

- add a schema-versioned log ownership registry with immutable resource generations and device/inode identity
- claim only files Spotter creates; pre-existing, replaced, out-of-boundary, and registry-reserved paths remain unowned
- register daemon and reviewer logs without turning registry failures into runtime or Hook failures
- document the compatibility and lifecycle boundary

This intentionally does not enable `purge --logs` yet; it establishes deletion authority for newly created logs while legacy/unrecorded logs remain ambiguous.

Closes no issue; advances #89.

## Validation

- `ruff format --check .`
- `ruff check .`
- `mypy src tests`
- `pytest` (773 passed)


YoungJinJung:
## Review

Reviewed the full diff and the CI failure. Holding the merge — the failing test is not flaky noise; it exposes a real gap in the PR's core invariant.

**Blocking finding — device/inode identity does not survive inode recycling:**
- `test_claim_does_not_adopt_replaced_log` fails on Linux CI because `path.unlink()` followed immediately by `path.write_text(...)` very commonly *reuses the just-freed inode number* on ext4/tmpfs. The claim's replaced-file check — `(existing.device, existing.inode) == (status.st_dev, status.st_ino)` — then matches, `claim` returns `True`, and the replaced file is adopted. The test presumably passed locally on APFS, where immediate inode reuse is less likely, but the CI behavior is the honest one: on the most common deployment filesystems, the stated guarantee ("a pre-existing or inode-replaced log remains unowned") does not hold, which means deletion authority could later attach to a file Spotter did not create.
- Suggested fix: strengthen the recorded identity beyond (device, inode). The practical discriminator is creation/change time — record `st_ctime_ns` (or birth time via `os.stat`'s `st_birthtime`/`statx` btime where available) at claim, and require it to match too; a recreated file always gets a fresh ctime even when the inode number is recycled. With that, the existing test becomes deterministic for the right reason rather than depending on allocator behavior. (Forcing a different inode in the *test* — e.g. creating an interposing file — would make CI green but would paper over the product gap; please fix the identity, not the test.)

**Otherwise the shape is right:** the claim boundary (log-dir-only paths, registry/lock files reserved, O_EXCL creation so pre-existing files are never adopted, symlink-refusing `stat`, non-regular-file refusal), the #47-style versioned/locked/fsynced store, and the fail-open integration into daemon/reviewer logging all match the series' discipline. I'll re-review and merge once the identity check is hardened and CI is green.

YoungJinJung:
Addressed the blocking identity gap in `ab0f650`.

I did not use `st_ctime_ns` because log appends themselves change ctime on Linux, which would invalidate legitimate mutable logs. Instead, every claim now creates a private hard-link identity anchor. The anchor keeps the original inode allocated, so an unlinked/recreated public path cannot receive the recorded inode through recycling. Claim verification requires both the public path and anchor to match the recorded device/inode; missing or mismatched anchors remain ambiguous.

Added deterministic coverage for replacement, path restoration through the anchor, missing-anchor ambiguity, fresh generation after both links are absent, and anchor boundary validation. Full suite: 775 passed.

YoungJinJung:
The follow-up commit resolves the blocking finding with a mechanism strictly stronger than the ctime suggestion in my review: a private hard-link anchor in `.ownership/` keeps the original inode *allocated*, so inode recycling into a replacement file becomes structurally impossible rather than merely detectable — the failing test now passes deterministically on every filesystem, for the right reason.

Verified the full trichotomy in `_anchor_status` and its callers: a valid anchor + matching path identity confirms ownership; a missing path with a valid anchor relinks the original inode (preserving content — the \"kept\" test is a nice recovery bonus); anchor-and-path both gone starts a clean new generation; and any ambiguous anchor state (unreadable, non-regular, identity mismatch, or missing while the path exists) refuses — never adopts. Boundary validation and duplicate detection extend to identity paths including the foreign-anchor case, new claims refuse when the anchor link cannot be created, and the `previous`-vs-`existing` bookkeeping correctly handles the new-generation append. Anchors live inside the log directory so the same-filesystem hard-link requirement always holds.

CI green across the board. Merging.

PR #260, opened 2026-08-15
feat(lifecycle): add log purge scope

YoungJinJung:
## Why

Issue #89 now has exact ownership for newly created daemon and reviewer logs, but no safe destructive log scope. Deleting or unlinking live log paths would either risk foreign files or orphan active writers.

## What changed

- add `spotter purge --logs` and `--dry-run` with human/JSON per-resource outcomes
- inspect exact ownership anchors and report every unregistered/replaced log path as ambiguous
- clear owned content through a revalidated `O_NOFOLLOW` anchor descriptor using `ftruncate` + `fsync`
- preserve public paths and ownership evidence so active writers remain valid and reruns are idempotent
- continue across per-log failures while returning non-zero for retryable, inaccessible, or ambiguous resources
- document the point-in-time semantics in EN/KO/ZH guides and lifecycle/status docs

A replaced public path is never touched: its still-owned anchor is cleared independently and the replacement is reported as ambiguous. Active writers may append new bytes after the point-in-time clear.

Closes no issue; advances #89.

## Validation

- `ruff format --check .`
- `ruff check .`
- `mypy src tests`
- `pytest` (779 passed)


YoungJinJung:
## Review

Reviewed the full diff (log_registry.py inspect/clear, cli.py, tri-lingual docs, tests) adversarially, as with the other destructive scopes.

**Correctness / safety — holds up:**
- Clearing through the *anchor descriptor* rather than the public path is the structurally right move: `clear()` re-validates registration under the lock, opens the anchor with `O_NOFOLLOW`, and fstat-verifies regular-file + exact (device, inode) *on the open descriptor* before `ftruncate` + `fsync` — a TOCTOU-safe compare that can never truncate a file other than the recorded inode. Because anchor and public path are hard links, owned content clears while the public filename is never touched, so active writers keep a valid destination and a replaced public path is unreachable by construction — the replaced-path test pins `foreign` surviving while the orphaned anchor empties.\n- The inspection matrix is conservative in every cell: ambiguous or missing anchors (with the path present) refuse as AMBIGUOUS, both-absent is safely absent, unregistered log-dir entries *and* orphan anchor files surface as AMBIGUOUS `unregistered:*` rows that are never touched, and the SAFE_OWNED reason string reports the public-path relationship (match/replaced/absent/inaccessible) for the operator.\n- The point-in-time caveat — an active writer may append immediately after the clear — is stated in the docs in all three languages rather than implied away; that's the honest framing for a truncate-style clear (same tradeoff as logrotate's copytruncate, including the harmless sparse-offset case for non-append writers).\n- Operational semantics match the #258 conventions: per-resource outcomes with `removed_bytes`, `failed_retryable` continuing past per-log failures (pinned by the selective-failure test), idempotent `already_absent` reruns with ownership evidence preserved, non-zero exit for anything skipped or failed, and exactly-one-scope CLI enforcement extended with `--apply` still refused.\n\nNo findings. 779 tests passing; the remaining #89 scopes (data/integration/all) stay explicitly pending. Will merge once the remaining CI jobs report.

PR #261, opened 2026-08-15
feat(lifecycle): preview durable data purge

YoungJinJung:
## Why

Issue #89 requires ownership-aware inventory before durable data can be deleted safely. Path or filename resemblance alone cannot prove that Spotter owns a resource.

Related: #89

## What changed

- Add a non-mutating spotter purge --data --dry-run scope with human and JSON output.
- Classify files as SAFE_OWNED only when every record matches the current schema name and version; exact regular lock companions inherit that proof.
- Report legacy, corrupt, empty, future-version, symlinked, unreadable, non-regular, and unknown paths as ambiguous or inaccessible.
- Keep configuration, repository, integration, runtime, and log resources outside the data scope.
- Continue refusing destructive data purge until revalidation and removal semantics land.
- Document the preview contract in lifecycle, status, and localized user guides.

## Validation

- ruff format --check .
- ruff check .
- mypy src tests
- pytest (799 passed)

## Notes

Destructive data cleanup and integration/all destructive scopes remain follow-up work in #89.


YoungJinJung:
## Review

Reviewed the full diff (new data_inventory.py, cli.py, tri-lingual docs, tests).

- The classification bar is deliberately stricter than the read paths, and that's the right asymmetry: every store happily *reads* legacy schema-less rows for evidence, but this preview marks a file SAFE_OWNED only when *every* record carries the current schema identity — readable-for-evidence never implies provably-deletable, so legacy and mixed-history files stay AMBIGUOUS and out of any future destructive scope. The whole #47 identity investment pays off here as deletion proof.
- The family map faithfully covers the durable inventory (all JSONL evidence stores, journal `.state` sidecars, nested task-batches under experiments, fork manifests, the spend ledger) while the config/repository/integration/runtime/log/pins roots are explicitly out of scope — clean separation from the other purge scopes' authorities.
- Lock companions inherit proof only when they are exact regular files next to a SAFE_OWNED companion; orphan or non-regular locks stay ambiguous. Symlinks are refused via `follow_symlinks=False` + `S_ISREG` throughout, empty files can't be owned by definition, blank JSONL lines fail the proof (stricter than the tolerant readers — again correct for deletion), and unrecognized entries anywhere in the tree surface as ambiguous rather than being skipped silently.
- The command is genuinely non-mutating (no write call anywhere in the module), destructive `--data` is refused at the parser, the exactly-one-of-four scope rule and flag fencing extend cleanly, and the non-zero exit for anything not safely attributable matches the established preview contract with `deletion_supported: false` stated in JSON.

No findings. 799 tests passing, lifecycle/status and all three user guides updated, with destructive data/integration/all scopes explicitly pending. Will merge once CI reports.

PR #262, opened 2026-08-15
feat(lifecycle): purge schema-proven data

YoungJinJung:
## Why

Issue #89 now has a conservative durable-data inventory, but destructive cleanup still needs per-store serialization and ownership revalidation. Deleting lock files is unsafe because waiting and new writers can split across different lock inodes.

Related: #89

## What changed

- Enable spotter purge --data while retaining the existing dry-run and JSON preview.
- Acquire each store lock without following symlinks, revalidate the complete current schema under that lock, unlink only proven data, and fsync the parent directory.
- Retain regular lock files as synchronization primitives and ignore regular orphan locks on idempotent reruns.
- Refuse non-regular locks and replaced parent directories, preserve ambiguous data, and continue processing independent safe files after a failure.
- Report per-resource removed, planned, preserved_synchronization, skipped_ambiguous, already_absent, and failed_retryable outcomes.
- Keep configuration, repository, integration, runtime, and log scopes untouched and document the destructive contract in all user guides.

## Validation

- ruff format --check .
- ruff check .
- mypy src tests
- pytest (808 passed)

## Notes

Active writers may recreate data after this point-in-time purge. Integration and all destructive scopes remain follow-up work in #89.


YoungJinJung:
## Review

Reviewed the full diff (data_inventory.py remove path, cli.py, docs, tests) adversarially, as with the other destructive scopes.

**Correctness / safety — holds up:**
- The lock-retention decision is the most important insight in the PR and it's right: unlinking a lock file splits waiting and future writers across different inodes (the classic flock pitfall), so locks stay behind as `preserved_synchronization` primitives and the post-removal orphan is silently absorbed on idempotent reruns — the rerun test shows an empty resource list rather than a spurious ambiguity.
- The removal sequence is TOCTOU-hardened at each step: the store's own lock opens with `O_NOFOLLOW` + fstat regular-file check, the directory chain from the data root is validated (no symlinked or substituted parents) before the lock, again under it, and once more immediately before `unlink`, and the *complete schema proof re-runs under the lock* — so a file swapped between planning and removal reinspects as ambiguous and survives. The symlinked-lock test pins `failed_retryable` with the data preserved while independent safe files still remove.
- Serialization is against the real writers: journal `.state` sidecars map to the journal's own `.lock` and the spend ledger maps to `review-spend.lock`, so removal contends on exactly the locks the runtime append/replace paths hold.
- Scope discipline survives the destructive turn: legacy/ambiguous data is preserved (pinned by the mixed-labels test), config/log/repository roots stay untouched even in apply mode, failures accumulate to a non-zero exit without aborting independent files, and `deletion_supported` flips truthfully while dry-run keeps the #261 preview contract.

**Minor (non-blocking):**
- The `_companion_lock` change makes a companion-less `.lock` file invisible in the inventory (previously AMBIGUOUS in #261). That's what makes reruns idempotent, but a genuinely foreign `x.lock` with no companion now goes unlisted — never touched, so no deletion risk, just a small visibility tradeoff worth remembering if operators audit the data dir through the preview alone.

Checks green (CI + CodeQL), 808 tests passing, docs updated across lifecycle/status and the three user guides — leaving integration/all as the final #89 scopes. LGTM — merging.

PR #263, opened 2026-08-15
feat(lifecycle): preview integration purge

YoungJinJung:
## Why

Issue #89 needs an ownership-aware integration plan before teardown-related resources can join destructive purge. The manifest is the authority; names resembling Spotter state are not deletion proof.

Related: #89

## What changed

- Add a non-mutating spotter purge --integration --dry-run scope with human and JSON output.
- Require the exact current integration manifest schema before deriving ownership.
- Compare every recorded Hook entry and surface modified, duplicate, or unrecorded Spotter-like Hooks as ambiguous.
- Reconstruct the managed launchd/systemd definition from the manifest runtime layout and compare exact bytes without creating directories, logs, or ownership claims.
- Verify recorded backup content against its fingerprinted filename and report missing recorded resources as already absent.
- Report unknown integration-directory entries and non-regular manifest/lock/service/backup paths conservatively.
- Continue refusing destructive integration purge; integration/all removal remains follow-up work.

## Validation

- ruff format --check .
- ruff check .
- mypy src tests
- pytest (813 passed)

## Notes

The service definition renderer is now exposed as a pure expected_definition method; the existing installation path still performs directory and log ownership setup before using the same bytes.


YoungJinJung:
## Review

This lands the integration-scope preview for #89 with the same posture as the previous scopes: ownership is derived from the exact current manifest schema, everything is compared byte-for-byte or key-for-key, and everything not provably Spotter's is reported ambiguous rather than touched. The preview is genuinely non-mutating, which I verified at the one spot where it could have regressed silently (details below).

### What holds up well

- **The `expected_definition` extraction is byte-safe.** The old `_definition()` used `secure_dir(...)`'s return value for the plist/unit paths; `secure_dir` returns its argument unchanged (it only mkdirs/chmods), so the new pure renderer produces the identical bytes the installer wrote, while `_definition()` keeps the directory/log-ownership side effects on the install path only. This is what makes the exact-bytes service comparison sound: install and preview now render from literally the same code path.
- **Service reconstruction avoids the mutation trap.** `_service_definition` rebuilds the `RuntimeLayout` from the manifest's *recorded* layout (not the current environment), so a relocated home doesn't silently "match," and it calls only the pure renderer — no directories created, no log ownership claimed, confirmed by the non-mutation test asserting the hooks file bytes are untouched.
- **Hook comparison via multiset (Counter) semantics is the right shape.** A modified recorded hook decomposes correctly into "recorded hook is already absent" (SAFE_OWNED/ABSENT) plus a Spotter-like unrecorded variant (AMBIGUOUS), so edits can never be mistaken for ownership. Duplicate identical hooks are handled by indexing rather than collapsing. Non-Spotter-like user hooks in the same file are correctly none of Spotter's business.
- **No-manifest mode stays conservative**: with no manifest there is no ownership at all, so every Spotter-like hook in the default `CODEX_HOME/hooks.json` is ambiguous — names resembling Spotter state are not deletion proof, exactly as the PR body says.
- **Backup verification checks content against the fingerprinted filename** rather than trusting the name, and a missing recorded backup is idempotent-absent rather than an error, consistent with `--data`.
- Exit-code contract (non-zero on any INACCESSIBLE/AMBIGUOUS), mutual-exclusion parser wiring, and destructive refusal for `--integration` all match the established scope pattern, with tests pinning each.

### Non-blocking observations

1. **`_manifest` validates and loads in two reads.** The raw bytes are schema-checked, then `IntegrationManifest.load()` re-reads the file; a writer racing between the two reads could get a manifest loaded whose bytes were never the ones validated. For a dry-run-only scope the consequence is just a stale report, so this is fine today — but the destructive `--integration` follow-up should parse the validated bytes directly (or re-validate under the lock, as `--data`'s `_reinspect` does) rather than inherit this pattern.
2. **An orphan `codex.lock` with no manifest is not listed.** In `_integration_directory`, a regular lock file only gets an inspection when a manifest exists; with no manifest it falls through silently. Since locks are never deleted this can't cause harm, but a leftover lock from a torn-down integration is arguably exactly the residue a preview exists to surface — same visibility trade-off I noted on #262's companion locks. Worth a one-line `ambiguous`/`preserved` entry in the follow-up.
3. **Platform inference from the registration suffix** (`.plist`→darwin, `.service`→linux) is pragmatic and correct for both real layouts; an unknown suffix falls back to `sys.platform` and would at worst report a definition mismatch as ambiguous — conservative direction, fine.

### Verdict

Non-mutating, conservatively grouped, and the one refactor with regression potential is provably byte-identical. Both observations are follow-up material for the destructive scope, not blockers. Merging once checks pass.

PR #264, opened 2026-08-15
fix(lifecycle): stabilize integration preview

YoungJinJung:
## Why

PR #263 review identified two non-blocking consistency gaps that should be closed before destructive integration purge: manifest validation and construction used separate reads, and an orphan regular codex.lock was invisible without a manifest.

Related: #89

## What changed

- Construct IntegrationManifest directly from the exact bytes whose current schema was validated, eliminating the read-to-read race.
- Report a regular codex.lock without an ownership manifest as AMBIGUOUS while continuing to leave it untouched.
- Add regression tests proving the inventory does not call the second-read loader and surfaces all no-manifest Hook/lock residue.
- Document orphan lock visibility in the lifecycle contract.

## Validation

- ruff format --check .
- ruff check .
- mypy src tests
- pytest (814 passed)

## Notes

This remains a non-mutating preview. Destructive integration cleanup follows after this ownership snapshot is stable.

### DIFF
--- a/docs/lifecycle.md
+++ b/docs/lifecycle.md
@@ -1398,7 +1398,7 @@ checks every recorded Hook entry, reconstructs and compares the managed service 
 manifest runtime layout, validates backup content against its fingerprinted filename, and reports
 the manifest and companion lock. Recorded resources already absent remain safely idempotent;
 unrecorded or modified Spotter-like Hooks, changed services/backups, future or legacy manifests,
-non-regular paths, and unknown integration-directory entries remain ambiguous. The preview is
+orphan locks, non-regular paths, and unknown integration-directory entries remain ambiguous. The preview is
 non-mutating and destructive `--integration` remains disabled.
 
 `purge --snapshots` removes only exact `SAFE_OWNED`, unreferenced registered worktrees and snapshot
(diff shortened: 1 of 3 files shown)


YoungJinJung:
## Review

Tight follow-up that closes exactly the two consistency gaps flagged on #263, with the right regression pins.

- **Single-read manifest construction.** `IntegrationManifest(**raw)` is built from the same bytes the exact-schema check validated, so the read-to-read race is structurally gone rather than narrowed. The monkeypatch test that makes `IntegrationManifest.load` throw "manifest must not be read twice" is the right kind of pin — it fails if anyone reintroduces a second read, not just if the race happens to fire. Adding `TypeError` to the except tuple is the correct companion: direct kwargs construction turns unexpected top-level keys into `TypeError`, and routing that to AMBIGUOUS makes the preview *stricter* than `.load()` about what counts as exactly-current-schema — the conservative direction for anything feeding a future destructive scope.
- **Orphan lock visibility.** A regular `codex.lock` with no manifest now reports as AMBIGUOUS ("lock has no ownership manifest") instead of silently vanishing from the inventory, while remaining untouched. The no-manifest test tightening from 4 to 5 ambiguous entries plus the explicit lock-type assertion pins it. Docs updated in the same change, matching the repo's contract-with-code rule.

One observation, not a blocker: `**raw` construction skips any normalization `.load()` might perform beyond field assignment. Today the downstream consumers (`_hooks`, `_service`, `_backups`) only `.get()` into the recorded structures, and the full-suite green confirms behavioral equivalence for the happy path — but if `.load()` ever grows canonicalization (path normalization, hook shape coercion), the inventory's view could drift from the installer's. Worth keeping the two construction paths adjacent in mind when `.load()` changes.

Scoped, non-mutating, both gaps closed with tests that prove the mechanism rather than the symptom. Merging once checks pass.

PR #265, opened 2026-08-15
feat(lifecycle): purge owned integration state

YoungJinJung:
Closes the destructive integration sub-scope of #89.

## Why

Integration preview already proves exact Hook, service, backup, manifest, and lock ownership, but users could not apply that verified plan.

## What changed

- adds destructive `spotter purge --integration` with a second exact inspection under the no-follow lifecycle lock
- refuses the complete transaction when any integration resource is ambiguous or inaccessible
- removes exact recorded Hook occurrences, managed service registration through the service boundary, and fingerprint-matching backups
- retains the lock inode plus a minimized `purged` ownership tombstone for serialized, idempotent retries and safe reinstall
- preserves unrelated Hooks/configuration and restores Hooks when service removal fails
- updates lifecycle/status and localized user guides

## Validation

- `ruff format --check .`
- `ruff check .`
- `mypy src tests`
- `pytest` (817 passed)

## Out of scope

Destructive `purge --all` remains the next #89 slice.


YoungJinJung:
## Review

This is the destructive integration scope, so I traced every mutation path end to end: the lock protocol, the all-or-nothing refusal, hook multiset removal, the service-failure rollback, tombstone re-entry, and the reinstall path. No blocking defects found.

### The transaction discipline holds

- **Preflight → lock → re-inspect → apply** mirrors the `--data` pattern correctly: the lifecycle lock opens `O_RDWR|O_CREAT|O_NOFOLLOW` with an `fstat` regular-file check before `flock`, and the full snapshot is repeated *under* the lock, so nothing decided pre-lock survives into the destructive phase. A symlinked lock fails `ELOOP` instead of following.
- **All-or-nothing on uncertainty is the right call for this scope.** Unlike `--data` (per-file independence), integration resources are one interdependent installation — removing hooks while refusing the service would leave a half-torn integration. `_has_uncertain` blocking the whole transaction on any non-SAFE_OWNED resource, including no-manifest mode never being destructive, is coherent with that.
- **Hook removal is exact-occurrence, not pattern-based.** The Counter decrement removes precisely one occurrence per recorded entry; the under-lock inspection already proved there are no unrecorded Spotter-like extras (they'd be AMBIGUOUS → blocked). The `hooks_file_created` unlink only fires when the remainder is exactly `{"hooks": {}}` — a file holding any other top-level keys survives as an atomic rewrite. User hooks in the same group are preserved, pinned by test.
- **The rollback asymmetry is justified.** Service-removal failure restores the hooks file (returning the user to a fully working integration for retry/teardown), while a backup-unlink failure does not — by that point hooks+service teardown already succeeded and backups are inert Spotter-owned files whose retry path is proven idempotent: absent recorded resources re-inspect as SAFE_OWNED/ABSENT, so the retry passes preflight and completes the remainder. I walked the tombstone-write-failure retry the same way; it converges too.
- **The tombstone closes the loop correctly.** `state="purged"` with minimized fields means re-running purge is a no-op under lock (early return, lock+manifest `preserved_synchronization`), leftover Spotter-like hooks against a tombstone report ambiguous (owned_hooks is empty), and the `integration.py` `retained` change makes reinstall treat a tombstone as a fresh setup — critically, `hooks_file_created` is re-derived rather than inherited as `False` from the tombstone, which would have broken the next teardown's file-removal decision. Good catch folding that into this PR.
- **New `state` validation** (`ready`/`purged` only, else ambiguous) tightens the preview and the destructive gate together — a mid-setup manifest can no longer pass ownership.

### Non-blocking observations

1. **Host hooks-file lost-update window.** The hooks file belongs to Codex and can't be covered by Spotter's lifecycle lock, so a Codex write landing between `_remove_hooks`' read and its atomic rewrite (or between removal and `_restore_hooks`) would be clobbered. The filtered-rewrite design (operate on the freshest image, remove only exact matches) minimizes the window to milliseconds and preserves concurrent user additions in the common case — this is inherent to un-lockable host files, just worth having stated in the lifecycle doc's assumptions.
2. **`skipped_ambiguous` labels SAFE_OWNED resources in a blocked transaction.** When one ambiguous resource blocks the whole transaction, provably-owned present resources also report `skipped_ambiguous`. The group field still says SAFE_OWNED so the information is recoverable, but a distinct outcome (e.g. `skipped_blocked`) would make the JSON self-describing for tooling.
3. **`manifest` reports `preserved_synchronization` while actually being rewritten** to the minimized tombstone. Accurate in spirit (the synchronization pair survives), slightly off literally; the docs describe it correctly.

### Verdict

Every destructive path re-verifies under the lock, refusal is transaction-wide, retries provably converge, and the reinstall interaction was handled rather than left as a landmine. The three observations are polish, not blockers. Merging once checks pass.

PR #266, opened 2026-08-15
feat(lifecycle): compose complete purge

### LINKED ISSUE #89: Add repository-aware purge, retention, uninstall, and reinstall lifecycle

YoungJinJung:
Closes #89.

## Why

Every focused purge scope is now destructive and ownership-aware, but `--all` still exposed only the repository preview and refused mutation. The lifecycle needs one auditable command that composes those proven boundaries without duplicating their safety logic.

## What changed

- enables destructive `spotter purge --all`
- composes integration, durable-data, snapshot, and log scopes in lifecycle order while continuing across independent failures
- preserves the historical top-level repository JSON view and adds complete per-scope payloads plus exit status
- removes an exhausted repository registry last under its no-follow writer lock while retaining the lock inode
- keeps manual pins, ambiguous resources, configuration, and synchronization evidence intact
- treats the actual `snapshot-pins.json.lock` as a separate-scope lock instead of ambiguous data
- updates lifecycle/status and localized user guides

## Validation

- `ruff format --check .`
- `ruff check .`
- `mypy src tests`
- `pytest` (818 passed)

## Compatibility

Existing `purge --all --dry-run --json` consumers retain the top-level repository `summary` and `resources`; complete scope plans are additive under `scopes`.


YoungJinJung:
## Review

The final #89 slice. I focused on the composition mechanism, the ordering dependencies between scopes, the registry-removal gate, and the pins-lock exclusion fix. No blocking defects.

### What holds up well

- **Composition reuses the proven scopes instead of reimplementing them.** `_purge_all_main` invokes the four existing scope mains and captures their JSON stdout. This is unconventional plumbing, but it's the right trade for this codebase: every safety property already verified in #262/#265 and the snapshots/logs scopes (locks, re-inspection under lock, transaction refusal, idempotent retries) is inherited *by construction* rather than re-derived. The non-JSON fallback turns a malformed scope result into a visible `failure` + exit 1, so the internal contract failing is loud, not silent — consistent with the fail-visibly rule.
- **The lifecycle order is load-bearing and correct.** Integration retires first (stops the producer), data removes schema-proven journals/experiment-results/fork-manifests, and *that* is what un-references snapshots so the snapshots pass can remove the now-unreferenced refs in the same run — the end-to-end test proves the journal→ref cascade. Manual pins survive the data scope, so pinned snapshots still retain through `--all`, exactly as documented.
- **Registry removal is gated twice.** The outer condition (no failures, every inspection SAFE_OWNED+ABSENT after the post-cleanup re-inspection) decides eligibility, then `_remove_repository_registry` re-loads and re-inspects *under the registry writer lock* (`O_NOFOLLOW` open, fstat regular check, flock) before unlinking with a parent-dir fsync — the same TOCTOU discipline as every other destructive path, with the lock inode retained. An empty registry counts as exhausted (vacuous `all()`), which is the correct reading. Anything pinned or still present preserves ownership evidence.
- **`snapshot-pins.json.lock` exclusion is a real bug fix riding along legitimately.** The exclusion list had `snapshot-pins.lock` but the actual companion lock is `snapshot-pins.json.lock`, so `--data` would have reported it AMBIGUOUS (and exited non-zero) whenever pins existed. Composing the scopes is exactly how this surfaced; the fix belongs in this PR.
- **Backward compatibility handled honestly.** The top-level `summary`/`resources` remain the historical repository view (the snapshots payload), with per-scope detail additive under `scopes` — and the existing machine-readable-preview test now pins the scope ordering. The `CODEX_HOME` addition to the `spotter_home` fixture is important: without it, `--all` tests would have inspected the developer's real `~/.codex`.
- **No lock nesting**: integration lifecycle lock, per-store data locks, global snapshot lock, and the registry writer lock are taken strictly sequentially, so composition introduces no deadlock ordering.

### Non-blocking observations

1. **No per-scope exception barrier.** `_capture_purge_scope` relies on each scope main catching its own error families; an exception outside those (unexpected `RuntimeError`, say) would abort `--all` mid-sequence with earlier scopes already applied. The scopes are independently idempotent so a re-run converges, but a `try/except Exception` per operation mapping to a failed scope payload would keep the continue-across-independent-failures promise unconditionally.
2. **`setdefault("failure", stderr)` can attach a failure string to a scope that exited 0** if anything writes benign stderr noise (warnings). Consumers should key on `exit_code`/`scope_status`, which the payload provides — worth a doc note on which field is authoritative.
3. The scope JSON contract is now a load-bearing internal API for `--all`. The fallback makes breakage visible rather than silent, which is acceptable; just something to remember when touching any scope's output shape.

### Verdict

Composition inherits the verified safety of the focused scopes, ordering makes the cascade work in one pass, registry removal follows the established under-lock re-verification pattern, and the compat surface is preserved and tested. This closes #89 coherently. Merging once checks pass.

PR #267, opened 2026-08-15
feat(config): resolve layered configuration

YoungJinJung:
## Why

Issue #90 requires one canonical configuration contract before reload and protocol compatibility can safely build on it. The runtime previously mixed direct single-file loads with ad hoc defaults, so CLI, daemon, integration validation, and Hook execution could interpret different settings.

## What changed

- add one immutable resolver for built-in, global, repository, explicit-file, and runtime-override layers
- deep-merge tables with documented precedence and validate the complete snapshot before activation
- preserve operator trust: repository policy cannot override `observation_only` or `mcp_semantics`, can only tighten gates, and invalid repository layers leave valid lower layers active with visible diagnostics
- attach a stable effective hash, source-aware generation, load time, and non-secret source provenance
- route CLI, daemon startup settings, integration validation, doctor Hook round trips, and the fail-open Hook path through the resolver
- resolve repository settings only from a real Git root using the Hook payload cwd
- make `observe`/`review` without `--config` use the documented default layers instead of requiring an explicit file
- document the implemented lifecycle and precedence in all user guides

This is the configuration-resolution foundation of #90; activation-boundary classification, atomic reload, and rolling protocol/build compatibility remain follow-up slices.

## Validation

- `ruff format --check .`
- `ruff check .`
- `mypy src tests`
- `pytest` (826 passed)


YoungJinJung:
## Review

The resolver architecture is right for #90 — immutable snapshot, whole-snapshot validation before activation, stable content hash vs source-aware generation split, provenance without secrets, required-explicit vs optional-default layers, and every consumer routed through one entry point. The tests pin precedence, dedup, stability, and the refuse-on-invalid-layer contract properly. But there is one finding I consider **blocking** before this merges.

### Blocking: the repository layer lets the supervised agent weaken its own supervision

Repository config wins over global config, tables deep-merge, and **scalars/lists replace**. The repository is exactly the surface the supervised agent can write to. That combination inverts the trust order for the three security-relevant settings:

1. **`gates.forbidden_paths` / `gates.block_dependency_changes`** — a user's global config says `forbidden_paths = ["secrets/*"]`; the agent writes `<repo>/spotter.toml` with `[gates] forbidden_paths = []` (lists replace, so this *clears* the global list); the next `PreToolUse` hook resolves repository-over-global and the deterministic gate is gone. The write that disarms the gate is itself just an ordinary file edit the gate does not cover.
2. **`[mcp_semantics]`** — a repo layer can reclassify a destructive MCP tool as a read/Class A operation, directly reversing the conservative-classification work from the remote/HTTP/curl/MCP semantics series. Classification inversions of this kind were blocking findings then; letting repo content author them makes the inversion self-service.
3. **`observation_only`** — a user who globally opted into intervention (`false`) can be silently flipped back to observe-only by repo content, disabling BLOCK delivery without any signal to the user.

This matters precisely because the hook now resolves the repository layer from the *payload cwd* on every invocation (which is otherwise the right mechanic). The supervision contract ("PreToolUse enforcement must remain bounded and deterministic", "fail visibly, not silently") assumes policy comes from the operator; this makes policy partially author-able by the supervised party, silently.

**Any of these would resolve it:**
- (a) Exclude supervision-relevant keys (`gates`, `mcp_semantics`, `observation_only`) from the repository layer — repo config keeps reviewer budgets, adapter, snapshot settings; refuse-or-warn when a repo layer names an excluded key (visible, not silent).
- (b) Merge those keys tighten-only: repo may *add* forbidden paths, *add* conservative semantics, may not remove or relax what global declared.
- (c) An explicit per-repository trust gate (operator must approve a repo's spotter.toml fingerprint before it participates in resolution), like direnv's allow model.

Option (a) is the smallest and matches the layering doc's spirit; (b) is the nicest semantics if you want repos to harden themselves.

### Non-blocking observations

1. **`_repository_root` falls back to the starting directory when no `.git` is found**, so `spotter <cmd>` run in an arbitrary non-repo directory silently loads `./spotter.toml`. Given the finding above, the fallback widens the attack surface for free; returning "no repository layer" when there is no repository would be tighter and matches the documented "discovered from the Git worktree root."
2. **`observe`/`review` without `--config` changed from error to defaults-with-layers.** This is the direction the user guide always documented and the daemon already behaved, so I read it as a deliberate consistency fix — worth one line in the PR body since it is a user-visible contract change.
3. The generation hash including `modified_ns` means a `touch` changes the generation but not the config hash — I take that as intended ("source-aware generation") and it will serve the reload slice well.
4. Hook fail-open is correctly preserved: resolution failures inside the boundary degrade to defaults with a stderr note, and the new subprocess test proves the payload-cwd repository resolution end to end.

### Verdict

Architecture is sound and the plumbing consolidation is overdue, but I'd like the repository-layer trust inversion addressed (or explicitly ruled out with rationale I'm missing) before merge. Happy to re-review quickly.

YoungJinJung:
## Re-review (dc37cae)

All blocking findings are resolved, and the resolution is better than the minimum I asked for.

- **`observation_only` and `mcp_semantics` are stripped from the repository layer** and surfaced both in `diagnostics` and in the new per-layer `ignored_fields` provenance — visible, not silent, exactly the contract.
- **Gates are tighten-only with the right algebra**: `forbidden_paths` unions (operator entries first, deduped), `block_dependency_changes` is logical-OR, and unknown `gates.*` keys from the repo are ignored-and-reported. The test proving a repo cannot flip any of the three levers while still being able to *add* `private/*` is precisely the semantics I hoped for.
- **Invalid repository layers now degrade instead of refusing**: the candidate-validation pattern (merge into a copy, `SpotterConfig.from_mapping(candidate)`, commit only on success) means a broken or malicious repo file can neither poison the snapshot nor deny supervision by crashing resolution — the operator config stays active with a diagnostic. This also elegantly routes ill-typed tighten-only values (non-list `forbidden_paths`, non-bool flag) into full-layer ignore via the same validation, rather than needing bespoke type errors. Explicit/global layers still refuse activation on invalidity, which is the right asymmetry: operator-authored layers fail loud, supervised-input layers fail safe.
- **`_repository_root` returning `None` outside a real Git worktree** closes the non-repo `./spotter.toml` pickup, with a test.
- **Setup-time strictness** — `integration.inspect` treating any diagnostics as unusable config — is a good choice: the operator is present at setup, so surfacing repo-policy conflicts there prevents them from becoming silent steady-state noise.

One residual, non-blocking: the repository layer can still set `reviewer.*` (budgets, `on_signals`, `deliver_on_signals`) and `snapshot_on_patch`. A repo zeroing reviewer budgets softens *advisory* supervision — much weaker leverage than the gate/semantics bypass since queue decisions remain journaled and delivery stays double-opt-in, and per-repo reviewer tuning is a legitimate use case. Worth revisiting when the reload slice lands whether those belong on an allow-list too, but I don't think it blocks this foundation.

Merging once checks pass.

PR #268, opened 2026-08-15
feat(config): classify activation boundaries

YoungJinJung:
## Why

The layered resolver from #267 can identify effective configuration generations, but #90 also requires every runtime-facing setting to declare when a change may safely activate. Without an exhaustive contract, reload code would have to guess which changes can publish immediately and which must wait, restart, reconfigure, or migrate.

## What changed

- add the five activation boundaries from #90: `HOT`, `NEXT_TURN`, `DAEMON_RESTART`, `INTEGRATION_RECONFIGURE`, and `SCHEMA_MIGRATION`
- classify every current `SpotterConfig` leaf field in one immutable registry
- add a value-free change classifier for reload/status planning without leaking configuration values
- make the test derive all dataclass leaf paths so a future unclassified field fails the suite
- document the exact per-key boundary and rationale, including why no current key uses `DAEMON_RESTART` yet

This is the activation-planning slice of #90. Atomic snapshot publication, generation pinning, and protocol/build compatibility remain follow-up work.

## Validation

- `ruff format --check .`
- `ruff check .`
- `mypy src tests`
- `pytest` (828 passed)


YoungJinJung:
## Review

Small, well-shaped slice. The design does the two things an activation-boundary contract must do and avoids the trap it could have fallen into.

- **The exhaustiveness test is the real deliverable.** Deriving leaf paths from the actual dataclass shape and asserting set-equality with the registry means an unclassified future field is a CI failure, not a reload-time guess — which is exactly the failure mode the PR body says this exists to prevent. Nested dataclasses recurse (`main_agent.adapter`, `gates.*`) while `mcp_semantics` (a tuple of dataclasses) is correctly treated as one atomic key — reinterpreting individual tool semantics mid-registry would have no meaningful per-entry boundary anyway.\n- **The classifier is value-free by construction** (`ConfigChange` carries path + boundary only), so reload/status surfaces can log deltas without leaking configuration contents — consistent with the non-secret-provenance stance from #267.\n- **The classifications themselves read correctly conservative.** Gate policy, `observation_only`, `mcp_semantics`, and `reviewer.model`/`deliver_on_signals` at `NEXT_TURN` upholds the documented invariant that one turn never mixes policy/model/delivery generations (and aligns with the #219 turn-fencing machinery); budgets/cadence at `HOT` is right because spend is durable and only new reservations consult limits; `config_schema_version` at `SCHEMA_MIGRATION` and `main_agent.adapter` at `INTEGRATION_RECONFIGURE` route to the existing lifecycle contracts. Documenting *why* no key is `DAEMON_RESTART` yet — rather than silently having an unused variant — is the kind of explicitness this repo's conventions ask for.\n- `MappingProxyType` keeps the registry immutable, and the change-order test pins deterministic output for the future reload planner.\n\nNo findings. The docs table replacing the speculative "likely behavior" table with the implemented per-key contract is a genuine docs-follow-code improvement. Merging once checks pass.

PR #269, opened 2026-08-16
feat(config): stage atomic reload snapshots

YoungJinJung:
## Why

#90 requires invalid or racing configuration reloads to preserve the prior good generation, and the activation registry from #268 requires whole snapshots to wait for their declared boundary. Field-by-field mutation would allow readers to observe mixed generations.

## What changed

- add a thread-safe active/pending configuration snapshot store
- parse and validate outside the short-held state lock while serializing competing reload requests
- atomically publish HOT-only candidates
- stage the whole candidate when any field requires `NEXT_TURN`, then publish at an explicit boundary
- report restart/reconfigure/migration requirements without activation
- preserve both active and previously validated pending snapshots after invalid reloads
- return only generations, changed paths, boundaries, and diagnostics rather than config values
- cover hot publication, turn staging, disruptive changes, invalid config, nonblocking readers, and two racing reloads

This is the atomic state-management slice of #90. Daemon reload triggers, runtime generation pinning, and rolling protocol/build compatibility remain follow-up work.

## Validation

- `ruff format --check .`
- `ruff check .`
- `mypy src tests`
- `pytest` (833 passed)


YoungJinJung:
## Review

The store gets the concurrency structure right, and the subtlest design decision in it is actually correct — details below.

- **Two-lock split is sound.** Parsing/validation runs under `_reload_lock` only, so readers on `_state_lock` never wait behind a slow resolve — and the racing test asserts exactly that (`snapshot()` returns while a loader is deliberately parked). Competing reloads serialize in request order; state mutations happen under both locks in `activate_next_turn`, so a turn-boundary publication cannot interleave with a reload's classify-and-stage.
- **Whole-candidate staging honors #268's contract.** A candidate containing any `NEXT_TURN` field stages *entirely* — no mixing hot fields of generation N+1 with turn-pinned fields of generation N. Same for `ACTION_REQUIRED`: disruptive boundaries gate the whole snapshot, with HOT riders deliberately waiting. Conservative and coherent with "earliest safe boundary."
- **The pending-discard semantics are the part I scrutinized hardest, and they hold up.** A HOT-only or ACTION_REQUIRED reload clears a previously staged `NEXT_TURN` candidate. This is correct *because every candidate is a full re-resolution of the layer stack*: if the newest resolution still contained the staged change, `classify(active, candidate)` would show the NEXT_TURN delta and the branch couldn't be reached — so the HOT-only path is only reachable when the staged change no longer exists in source, and latest-wins discard is exactly right. The racing test pins the same property from the other direction (two staged reloads → only the newest survives to the boundary).
- **Invalid reloads preserve both active and previously validated pending**, with a value-free `last_error` for diagnostics — the #90 requirement, tested with a real broken file layer rather than a mock.
- Result objects stay value-free (generations, paths, boundaries, error string), consistent with the #267/#268 posture.

Two minor notes, neither blocking: (1) `activate_next_turn` with nothing staged returns `APPLIED` — a distinct no-op disposition would make callers' logs clearer, trivially changeable when the daemon wiring lands; (2) `activate_next_turn` clears `_last_error` although activation isn't a validation event — harmless today, worth keeping intentional once a status surface reports it.

Nothing consumes the store yet, which the PR states plainly as the next slice — fine as a foundation with this level of test coverage. Merging once checks pass.

PR #270, opened 2026-08-16
feat(reviewer): pin config generations

YoungJinJung:
## Why

Reviewer jobs can remain queued or in flight while runtime configuration changes. They must keep the model, budgets, and delivery policy under which they were submitted, with durable provenance for later metrics and diagnosis.

Part of #90.

## What changed

- pin the immutable reviewer config generation and model onto each executor submission
- keep queued/running inference, budget, shadow, and delivery decisions on the captured config
- journal a durable pin event and propagate the generation through reviewer outcome/error events
- resolve one daemon config snapshot at startup and expose its generation through daemon status/CLI output
- document the implemented runtime contract and add concurrent old/new-generation coverage

## Validation

- `ruff format --check .`
- `ruff check .`
- `mypy src tests`
- `pytest` (835 passed)


YoungJinJung:
## Review

The pinning mechanics are correct and the concurrency test proves the property that matters. No blocking findings.

- **Submission-time capture is complete.** The queue tuple carries the `(config, generation)` pair alongside the job, and `_run` consults only the captured config for budget reservation, model selection, shadow determination, and delivery gating — `self.config` is never read after submission. `update_config` therefore cannot affect in-flight or queued work by construction, and the test demonstrates it the hard way: an old-generation job parked *inside* inference while a new-generation job is submitted, with both decisions journaled under their own model+generation. The queue tuple stays safely comparable since the monotonic sequence breaks ties before comparison could ever reach the job/config elements.\n- **Provenance is durable and consistent.** `review_job_config_pinned` lands at submission, and every downstream event (inference start, decision, caps, budget failure, cancellation, delivery error) repeats the generation — so metrics can attribute cost/behavior to an exact config generation even across a reload, which is precisely what #33-style analysis needs.\n- **The daemon consolidation is a real correctness improvement, not just cleanup**: previously reviewer/MCP-semantics/snapshot settings came from three independent `resolve_config` calls that could straddle a file edit and disagree; now one `_resolved_daemon_config` resolution yields one generation for all three, exposed through control status and CLI. The fallback path (`fallback-unversioned`) keeps startup fail-open with a visible stderr note.\n- `ReviewerJob`'s new fields default to `None`, keeping existing constructors/journal readers compatible.\n\nOne observation, non-blocking and I read it as intended: pinning means disabling `reviewer.deliver_on_signals` (or `on_signals`) is **not retroactive** — a job submitted under an enabled generation will still deliver its advisory even if the operator disabled delivery while it was queued. The docs state this contract explicitly ("running and queued jobs cannot silently switch … delivery policy"), and exposure is bounded by the existing turn-freshness fencing (stale results never deliver), so the window is one target turn. If a hard kill-switch semantics is ever wanted, it belongs in the reload/turn-boundary wiring as an explicit drain rather than a change to pinning.\n\nMerging once checks pass.

PR #271, opened 2026-08-16
feat(config): add daemon reload control

YoungJinJung:
## Why

The atomic config snapshot store existed, but a running `spotterd` had no safe trigger or status surface for using it. Operators had to restart even for settings classified as `HOT`.

Part of #90.

## What changed

- add `spotter daemon reload` and a typed local control response
- resolve/validate reload candidates off the daemon event loop
- atomically apply `HOT` reviewer and patch-snapshot settings to future work
- stage whole `NEXT_TURN` candidates without changing the active generation
- preserve active and pending generations after invalid reloads
- expose active/pending generations and the last reload error through daemon status
- document the new command and activation behavior

Automatic file watching and explicit turn-boundary activation remain follow-up work.

## Validation

- `ruff format --check .`
- `ruff check .`
- `mypy src tests`
- `pytest` (837 passed)


YoungJinJung:
## Review

The wiring is correct, and the one place where a race could have crept in is handled by design rather than by luck. No blocking findings.

- **The interleaving I probed hardest is safe.** `reload_config` runs the store's reload in a worker thread while the event loop keeps serving; `_apply_active_config` then reads `config_store.snapshot()` — the *current* active — rather than the returned result. So if two concurrent reload requests interleave (A applies gen2, B applies gen3, then A's apply step runs), both apply steps converge on the latest store snapshot instead of resurrecting a stale generation. Reading from the store instead of the result object is the load-bearing choice here and it's the right one.\n- **`_apply_active_config` updating only reviewer + `snapshot_on_patch` is correct by construction, not an omission**: `mcp_semantics` is classified `NEXT_TURN` (#268), so an `APPLIED` disposition can only be reached when it didn't change — the hot-apply path never needs to touch it. The same argument covers gates (client-supplied per request) — worth keeping this invariant in mind when turn-boundary activation lands, since *that* path will need to publish the full snapshot.\n- **Executor handoff composes with #270 exactly as designed**: `update_config` swaps the generation for future submissions while in-flight jobs keep their pinned config — the end-to-end test's `max_per_session` 2→7 apply plus `reviewer.model` staging demonstrates both boundaries against a real socket.\n- **Client-side protocol parsing is properly paranoid** — typed validation with a length check that catches partially malformed `changes` lists, mapping to `DaemonProtocolError` instead of propagating KeyErrors. Status surfaces pending generation and last rejected reload without values, and the CLI exit codes (0 for APPLIED/STAGED, 1 for ACTION_REQUIRED/REJECTED) give scripts a usable signal.\n- The rejected-reload test proves the full #269 preservation contract through the daemon: active stays, previously staged pending stays, error is visible.\n\nOne minor asymmetry, non-blocking: a daemon that started on the **fallback** path (unusable config at boot) has no store and reports "config reload is unavailable; restart spotterd" — so the one operator who most wants a reload (they just fixed their broken config) is the one who can't have it. The message is honest and restart is cheap, so this is fine for now; if it ever grates, boot could construct the store lazily on first successful reload rather than requiring a restart.\n\nMerging once checks pass.

PR #272, opened 2026-08-16
feat(config): activate staged config between turns

YoungJinJung:
## Why

`NEXT_TURN` config candidates could be staged through daemon reload, but nothing activated them at a runtime-safe boundary. Activating on the first arbitrary new turn would let another concurrently active turn inherit new reviewer or MCP semantics midway through execution.

Part of #90.

## What changed

- invoke a daemon config boundary callback before normalizing each App Server `turn/started` event
- activate a staged generation only when every daemon-owned thread is quiescent
- publish the full reviewer, snapshot, and MCP-semantics snapshot before the new turn is reduced
- keep staged config pending while any older turn remains active
- cover callback ordering, active-turn deferral, generation activation, and MCP normalizer replacement

## Validation

- `ruff format --check .`
- `ruff check .`
- `mypy src tests`
- `pytest` (838 passed)


YoungJinJung:
## Review

This closes the loop on the #269/#271 staging machinery, and the two properties that make it correct are both present and tested. No blocking findings.

- **The callback fires before the triggering `turn/started` is normalized**, so the new turn is reduced entirely under the generation the boundary just published — the new turn can never see half of each. The runtime-connection test pins this ordering the right way (boundary snapshot shows no active turn yet, while the turn is provably reduced afterward).\n- **The global quiescence check is the correct scope.** Activating when *any* daemon-owned thread still has an active turn would hand that older turn new MCP-effect semantics mid-execution — exactly the mixed-generation hazard #268 classified `NEXT_TURN` to prevent. Deferring until every thread is quiescent, verified end-to-end in the daemon test (old turn active → deferred; completed → activated, normalizer rules swapped, pending cleared, reload error cleared), honors that contract across threads, not just within one.\n- **No cross-thread races by construction**: the callback runs synchronously in the `_consume` coroutine on the same event loop that mutates thread state, and the store's own locks cover the activation itself. The epoch check before the callback also means a stale connection's replayed `turn/started` cannot trigger activation.\n- **`_apply_turn_config` publishes the full snapshot** — reviewer + executor generation via the existing hot path, plus `mcp_semantics` into both the daemon and the live normalizer. Hook-side settings (gates, `observation_only`) need no daemon publication since every hook invocation resolves fresh config — the division of responsibility is clean.\n\nTwo observations, neither blocking:\n1. **Theoretical starvation under continuously overlapping turns**: if multiple threads interleave so that some turn is always active at every `turn/started`, a staged generation could stay pending indefinitely. For a single-user Codex daemon this is a corner case, and status makes the pending generation visible — but if it ever matters, a quiescence check on turn *completion* (not only at the next turn's start) would activate at the earliest true idle moment instead of waiting for the next turn to begin.\n2. The boundary callback takes the store's pending-generation lock on every `turn/started` even when nothing is staged — negligible today, just worth remembering if the boundary path ever grows hotter.\n\nMerging once checks pass.

PR #273, opened 2026-08-16
feat(gates): record config generations

YoungJinJung:
## Why

PreToolUse policy is resolved in a short-lived Hook process, so attributing a gate decision to the daemon's current config can be wrong during reloads or repository-layer changes. The exact Hook generation must travel with the decision.

Part of #90.

## What changed

- propagate the Hook-resolved config generation through each daemon gate request
- retain the generation in Hook `gate_ipc` telemetry for success and fallback paths
- echo the gate-policy generation separately from the daemon runtime generation
- attach the generation to daemon-correlated deterministic gate-block events
- preserve `unversioned` compatibility for direct/legacy callers

## Validation

- `ruff format --check .`
- `ruff check .`
- `mypy src tests`
- `pytest` (838 passed)


YoungJinJung:
## Review

Correct and complete for what it claims. The design insight is right: the Hook process resolves policy independently of the daemon (and #267 made that resolution repository-sensitive), so the only trustworthy attribution for a gate decision is the generation *that Hook invocation* resolved — carrying it with the request instead of inferring from the daemon's current snapshot is exactly what reload-time attribution needs.

- **Coverage is complete across paths**: the generation lands in `gate_ipc` telemetry on success, timeout/unavailable fallback, *and* capture-only short-circuit; in the daemon's response echoed as `gate_config_generation` (correctly named apart from the daemon's own runtime generation, which can legitimately differ from the Hook's during a reload window); and on the daemon-correlated `gate-block` trace event with a defensive `unversioned` fallback.\n- **Rolling protocol compatibility is handled in both directions**: an old daemon ignores the extra param (it only reads proposal/gates/root), a new daemon defaults a missing param to `unversioned`, and `_evaluate_gate` validates the field as a non-empty string so a malformed client can't inject junk provenance. The pre-loaded-config path (doctor round trips) staying `unversioned` is honest labeling rather than fake precision.\n- The fail-open contract is untouched — generation attribution rides the existing telemetry, adding no new failure mode to the hot path, which keeps rule 3 (bounded deterministic PreToolUse) intact.\n\nNo findings. Tests pin the IPC round trip and the journaled telemetry. Merging once checks pass.

PR #274, opened 2026-08-16
feat(config): reload files automatically

YoungJinJung:
## Why

Issue #90 already had atomic manual reloads, but long-lived daemons did not notice config file changes on their own. Runtime traces and derived signals also lacked durable generation provenance, so behavior around a reload could not be attributed precisely.

## What changed

- watch the global and explicitly selected config files for creation, replacement, edits, and removal
- route automatic changes through the existing atomic validation, HOT apply, and NEXT_TURN staging contract
- preserve the last valid active generation and visible error when an automatic reload is invalid
- persist config generation on daemon-owned source/derived Trace IR, ThreadState item provenance, effects, reviewer transitions, and control telemetry
- make SignalCandidate events inherit the exact triggering trace generation
- document the implemented lifecycle and leave protocol/build compatibility as the next #90 slice

## Validation

- ruff format --check .
- ruff check .
- mypy src tests
- pytest (841 passed)

Closes part of #90.


YoungJinJung:
## Review

Both halves — the file watcher and the generation provenance sweep — are correctly built. One robustness note below, non-blocking.

### Watcher

- **The signature tuple `(dev, ino, size, mtime_ns)` is the right identity check**: atomic replacement changes the inode, in-place edits change mtime_ns/size, and `None` for a missing file makes creation and removal first-class transitions rather than special cases. The end-to-end test walks the full lifecycle — create→apply, corrupt→active generation and value preserved with a visible error, edit→new generation, remove→revert to defaults with the error cleared — which is exactly the #269 contract exercised through the automatic path.\n- **Routing automatic changes through `reload_config` (not a parallel path)** means every property already verified — validation off the event loop, HOT-only atomic apply, whole-candidate NEXT_TURN staging, invalid-reload preservation — holds for file-triggered reloads for free. Removal of the global file simply resolves to a defaults snapshot and follows the same boundaries.\n- The watch set is complete for the daemon's layer stack: the global path is watched unconditionally (so post-startup creation is caught), and any explicit manifest-selected file joins via the initial snapshot's source layers. The daemon has no repository layer, so the fixed watch set can't go stale.\n\n### Provenance\n\n- **Derived events inheriting the *triggering* event's generation** (signals, scheduler queued/discarded/stale, effect observations) rather than the current runtime generation is the correct attribution semantics — a candidate produced from an old-generation trace stays attributed to the config that shaped it, even if activation happened in between. Backfilling `None` with the runtime generation in `_record`/`_append_derived`/control telemetry covers daemon-originated events, and the Hook-side gate generation from #273 flows through `record_gate_decision` untouched.\n- Journal compatibility is handled the boring right way: an optional metadata field, absent on legacy records (loads as `None`), ignored by older readers, with a round-trip test.\n\n### Non-blocking\n\n1. **The watch task has no exception barrier.** `ConfigSnapshotStore.reload` catches the enumerated error families, and the daemon's loader pre-handles manifest read failures, so the reachable failure set is small — but a truly unexpected exception in the loop would kill `spotter-config-watch` silently, and automatic reloads would just stop while manual `daemon reload` kept working. A `try/except Exception` around the loop body with a stderr note (or degraded-health signal) would make that failure visible per the fail-visibly rule.\n2. Polling at 0.5s stats a handful of paths — negligible, just noting the interval is a constructor knob if it ever needs tuning.\n\nMerging once checks pass.

PR #275, opened 2026-08-16
feat(runtime): negotiate daemon compatibility

YoungJinJung:
## Why

A package upgrade can leave an older `spotterd` process serving the stable socket. PID/socket liveness and semantic versions alone cannot tell the CLI or Hook bridge whether control remains compatible, and protocol failures previously surfaced as generic parse/control errors.

Part of #90.

## What changed

- publish selected IPC protocol, supported peer range, named control capabilities, exact build identity, runtime generation, and daemon start time on the control handshake
- classify daemon state as matched, compatible stale, incompatible stale, or legacy/unknown
- cleanly reject incompatible client/server ranges with actionable restart guidance while retaining legacy same-protocol interoperability
- surface compatibility/runtime identity in `daemon status` and make `doctor` warn for stale/unknown processes and fail incompatible ones
- keep the automatic config watcher alive and visibly degrade health after unexpected reload exceptions, addressing the follow-up review note from #274
- document the implemented runtime upgrade contract and next #90 slice

## Validation

- `ruff format --check .`
- `ruff check .`
- `mypy src tests`
- `pytest` — 847 passed


YoungJinJung:
## Review

Solid close-out for the #90 compatibility slice. The negotiation logic is conservative in the right places and the classification lattice is complete. No blocking findings.

- **The compatibility lattice covers every peer shape without guessing.** Exact protocol + range + matching build → `matched`; same protocol, different build → `compatible_stale` (WARN, diagnosable, restart advised); protocol/range conflict → `incompatible_stale` (doctor FAIL with the exact command to run); same-protocol legacy peers *without* range metadata → `unknown` rather than assumed compatible — that last choice is the one that's easy to get wrong, and the PR gets it right per the fail-visible rule.\n- **Both directions are validated symmetrically**: the client rejects a daemon outside its range (`DaemonProtocolMismatch` mapped to degraded status + restart guidance instead of a generic parse error), and the server rejects an incompatible client peer range with a typed `error_code` while attaching its own identity metadata to the error response — so even a rejected peer learns who rejected it. The test proving a rejected peer doesn't poison subsequent well-formed clients covers the connection-scoped nature of the rejection.\n- **Contradictory metadata is distinguished from missing metadata**: an invalid range (min > max) is a protocol *error*, absent range is legacy tolerance, range-excluding-client is a *mismatch* — three different failure semantics rather than one lumped error. `_optional_int` excluding `bool` is the kind of small rigor that prevents `true` masquerading as protocol 1.\n- **The #274 review follow-up is properly closed**: the watcher now wraps the reload in a catch-all that degrades health, prints to stderr, and keeps the task alive — with a test that explodes the loader once and proves the *next* file change still reloads. Exactly the fix requested.\n- `runtime_generation` (process instance) + `started_at` on the handshake close the last identity gap: same build restarted is now distinguishable from the same process, which the doctor line surfaces.\n\nTwo cosmetic notes: (1) `compatibility=unknown` is appended to `daemon status` output even for `unavailable`, where it's noise — trivially suppressible; (2) health degraded by a watcher failure is sticky — a later successful automatic reload doesn't restore `healthy`. Conservative and arguably intentional (an operator should learn a reload was missed), but if the daemon is meant to self-heal, clearing the degradation on the next successful reload would be the place.\n\nMerging once checks pass.

PR #276, opened 2026-08-16
feat(cli): add package-aware update advice

YoungJinJung:
## Why

Issue #90 requires upgrades to respect the package owner. A generic self-updater could overwrite Homebrew, pipx, uv, or pip-managed files, while source/editable installs need to remain under their development workflow.

Part of #90.

## What changed

- add an advisory-only `spotter update` command that reports the current version/build
- classify Homebrew, pipx, uv tool, pip, editable, and source installations from executable/environment boundaries plus standard distribution metadata
- print the owning package manager's inspect and upgrade commands without invoking them
- direct source/editable installs back to their owning workflow
- include the required `spotter setup codex && spotter doctor` reconciliation step
- document the implemented update contract and leave graceful stale-daemon restart/reconnect as the next #90 slice

## Validation

- `ruff format --check .`
- `ruff check .`
- `mypy src tests`
- `pytest` — 861 passed
- manual editable-install smoke: `python -m spotter update`

## Out of scope

This does not execute package-manager commands or restart/reconcile the daemon automatically.


YoungJinJung:
## Review

Clean, correctly scoped slice. The design decision that matters — advisory-only, never invoking the package manager or writing into its boundary — is exactly what the lifecycle doc's "must not compete with Homebrew for file ownership" rule requires, and it also bounds the blast radius of the classification heuristics: a misdetected owner yields wrong *advice*, never a wrong *mutation*.

- **Classification order is sensible**: environment-boundary evidence (Cellar / pipx venvs / uv tools) wins over metadata, editable `direct_url` beats plain distribution presence, and no-distribution falls through to `source`. Each rung has a stated evidence string surfaced in the output, so a user can see *why* Spotter picked an owner — consistent with the provenance posture everywhere else in this codebase. The path matching is on exact lowercase path parts (not substrings), avoiding the `/home/pipx-user` style false positives; requiring both `pipx`+`venvs` / `uv`+`tools` parts keeps it tight.\n- **Editable/source installs refusing self-update guidance** (rather than suggesting a pip upgrade that would clobber a development checkout) is the right conservative branch, pinned by test.\n- `shlex.quote(sys.executable)` for the pip commands handles exotic interpreter paths; the reconciliation step (`setup codex && doctor`) being part of the printed contract closes the loop with the #275 stale-daemon detection — the user who follows the printed sequence ends with a matched daemon.\n- Zero subprocess invocations anywhere in the path, matching the out-of-scope declaration; the classification matrix, per-owner advice, refusal branch, CLI routing, and argument rejection are all covered.\n\nNo findings. Merging once checks pass.

PR #277, opened 2026-08-16
fix(integration): reconcile candidate runtime inputs

YoungJinJung:
## Why

`spotter setup codex` previously started or restarted `spotterd` before committing the new integration manifest. A daemon replacement therefore read the retired manifest and could miss a newly selected config path even though setup later reported the new integration as ready.

Part of #90.

## What changed

- include runtime mode in the integration-generation fence
- atomically stage the candidate manifest as `configuring` before starting/restarting the owned runtime
- restart when the integration generation changed or a prior setup was interrupted in `configuring`
- verify daemon health and the Hook round trip against those candidate inputs, then commit the manifest as `ready`
- on startup or verification failure, restore host mutations and the prior manifest, then restore the prior running/stopped runtime state
- document graceful setup reconciliation and leave effective construction-fingerprint diagnostics as the next #90 slice

## Validation

- `ruff format --check .`
- `ruff check .`
- `mypy src tests`
- `pytest` — 864 passed

New tests prove the runtime observes the candidate config path before restart, rollback restores both manifest and previous runtime inputs, and interrupted `configuring` state forces reconciliation.

## Out of scope

This does not yet publish or compare a separate effective runtime-construction fingerprint on the daemon handshake.


YoungJinJung:
## Review

This fixes a real ordering bug — the daemon's startup config resolution reads `config_path` from the manifest (#270's `_resolved_daemon_config`), so starting/restarting the runtime *before* committing the manifest meant a replaced daemon could bind to the retired integration's inputs while setup went on to report success. Staging `configuring` → reconcile → verify → commit `ready` is the correct transaction shape, and the details hold up.

- **The rollback ordering is right where it's easiest to get wrong**: host mutations first, then the manifest restore, and only *then* the service restart — so a rolled-back daemon re-reads the *prior* manifest, not the candidate. The test pins this from the daemon's perspective (`config_paths_seen` ends back at the first config), which is much stronger than asserting file bytes alone.\n- **The runtime-state restoration matrix is complete**: was-running → restart (on prior inputs), preexisting-but-stopped → stop, never-existed → uninstall, and the `service_attempted` flag keeps early-phase failures (inspect/apply) from touching a service that setup never touched — preserving the old behavior exactly where the old behavior was correct.\n- **`restart_required` composes the two staleness signals properly**: a changed integration generation *or* an interrupted `configuring` state forces a restart rather than a start, and adding `runtime_mode` to the generation fence closes the portable↔managed switch that previously wouldn't have rotated it. The interrupted-`configuring` test proves crash-recovery reconciles to `ready` instead of wedging.\n- **The `configuring` state interacts safely with the purge machinery from #264/#265**: the inventory's state validation admits only `ready`/`purged`, so a mid-setup or crashed-setup manifest reads as AMBIGUOUS and destructive integration purge refuses — conservative in the window, and setup itself is the reconciliation path. Worth keeping that three-way state contract (`ready`/`purged`/`configuring`→ambiguous-elsewhere) in mind if a fourth state ever appears.\n- `manifest_before` is captured under the lifecycle lock before any mutation, so restore is byte-exact including the fresh-install case (unlink when there was no manifest).\n\nOne minor note: rollback's service operations are exception-suppressed so they can't mask the original failure — the trade-off is that a failed rollback restart leaves the runtime down with only the setup error + doctor to surface it. That's the established pattern in this file and the right priority order; just worth remembering the fingerprint-diagnostics follow-up will be what makes that residual state visible.\n\nMerging once checks pass.

PR #278, opened 2026-08-16
feat(runtime): fingerprint effective daemon construction

YoungJinJung:
## Why

A same-build daemon can still be stale when its IPC path, integration generation, adapter, or App Server construction inputs differ from the current installation. Build-only reconciliation silently reused that process. This advances #90.

## What changed

- compute a stable non-secret `runtime-*` fingerprint from process-construction inputs while excluding HOT reviewer settings
- publish and parse the fingerprint in the daemon handshake and show matched/missing/drifted state in status/doctor
- restart manual or managed same-build runtimes on a known construction mismatch, while preserving legacy peers with unknown evidence
- document the compatibility axis and cover fingerprint stability, drift diagnostics, handshake propagation, and launchd/systemd reconciliation

## Validation

- `ruff format --check .`
- `ruff check .`
- `mypy src tests`
- `pytest` (870 passed)


YoungJinJung:
## Review

The compatibility axis this adds is real — build identity alone genuinely cannot distinguish a same-build daemon constructed from retired inputs — and the implementation gets the three things right that matter for this kind of mechanism. No blocking findings.

- **Symmetry between the two computations is maintained where it could silently break.** The daemon computes its fingerprint at boot from its actual resolved config + manifest; the CLI's `expected_runtime_construction_fingerprint` replicates the identical resolution *including the fallback path* (`configured_runtime_config` degrades to the same defaults the daemon boot uses), so a broken config file yields matching fallback fingerprints on both sides rather than a spurious drift warning. Using setup's *recorded* `daemon_executable` instead of rediscovered layout — with the comment explaining the `spotterd`-vs-`python -m` spelling asymmetry — closes the other false-drift trap.\n- **The restart policy is conservative exactly where it must be**: only a *known* mismatch (fingerprint present and different) authorizes killing a same-build process; a legacy daemon reporting no fingerprint is diagnosed as WARN/unknown, never terminated on ambiguous evidence. That's the same evidence discipline as the log-ownership and purge work, applied to process lifecycle.\n- **It complements rather than duplicates the #277 generation fence**: the integration generation covers schema/build/prefix/config-path/mode, while the fingerprint additionally captures App Server strategy/endpoint, adapter, and the control socket/protocol range — inputs a generation-unchanged manifest edit could still drift. HOT reviewer settings are correctly excluded (they reload without restart, so including them would cause pointless restarts), and NEXT_TURN-managed settings stay with the reload machinery.\n- **Non-secret by construction, and tested as such**: only the digest crosses the boundary, and the test pins that an endpoint token doesn't leak into the fingerprint string.\n- Coverage hits the right seams: stability across HOT changes, sensitivity per construction input (including socket path), staged-manifest reads, handshake propagation, doctor's three-way matched/missing/drifted verdict, and launchd/systemd restart-on-drift on both platforms.\n\nOne small note: the started-daemon readiness loops accept `fingerprint in {None, expected}` — the `None` arm exists for a legacy binary answering during rolling upgrade, which is correct today; once a release floor guarantees fingerprint support, tightening that to exact-match would close the last soft edge.\n\nMerging once checks pass.

PR #279, opened 2026-08-16
feat(runtime): surface App Server capability changes

YoungJinJung:
## Why

Reconnect already fenced controls with a new epoch, but a changed App Server build or capability snapshot was not durably recorded or visible through daemon diagnostics. Operators could see a healthy socket while observation/control semantics had changed. This advances #90.

## What changed

- compare concrete App Server capability snapshots and server fingerprints after each reconciled reconnect
- journal non-secret `runtime_capabilities_changed` events on reconciled threads while preserving existing epoch fencing
- keep connection capability state current after control probes
- expose App Server state, connection epoch, capability statuses, and last reconnect drift through daemon status, CLI output, and doctor
- document the reconnect contract and cover server/capability drift, handshake propagation, and diagnostics

## Validation

- `ruff format --check .`
- `ruff check .`
- `mypy src tests`
- `pytest` (872 passed)


YoungJinJung:
## Review

Correct mechanism with the right evidence discipline. No blocking findings.

- **The `UNKNOWN`-means-unprobed rule is the load-bearing semantic choice, and it's right.** `_capability_changes` only records a transition when the new connection *concretely* reports a status, so an unprobed surface on a fresh connection can never masquerade as a lost capability — the same do-not-invent-observations discipline the effects ledger follows. Combined with comparison happening *after* `_reconcile` (when probing has actually run), false-loss events are structurally excluded.\n- **The journal event is deduplicated and correctly scoped**: `spotter:capabilities:{thread}:{epoch}` gives one durable record per reconciled thread per epoch, recorded only for threads that actually crossed the boundary (state epoch == new epoch), with a payload of fingerprints, epochs, and status transitions only — no server details leak. Threads born after the reconnect correctly get nothing, since they never lived under the prior capability set.\n- **Epoch fencing is untouched** — the PR adds observation of the boundary without altering it, and the `_refresh_connection_capabilities` guard (same client instance + same epoch) prevents a stale dispatch's probe results from overwriting a newer connection's snapshot.\n- **Doctor severity mapping is well-judged**: missing identity → WARN restart; non-ready → WARN with epoch; capability transitions → WARN; server identity change alone with stable capabilities → INFO (an upgrade that changed nothing observable deserves a note, not an alarm); stable → OK with the full snapshot inline.\n- The end-to-end test is the right shape — a real reconnect through the recovery loop with two fake servers, asserting the flag, the journaled transition (`steer available→unavailable`), and that `interrupt`'s `unknown` produced no change entry.\n\nOne non-blocking observation: drift detection runs at the reconnect boundary only. `_refresh_connection_capabilities` keeps the *snapshot* current after later control probes (so status/doctor show the truth), but a capability first discovered unavailable mid-epoch updates the display without a `runtime_capabilities_changed` journal entry or flag. That's defensible — the reconnect is the semantic boundary, and mid-epoch probe results already land in the control events themselves — but if the labeling pipeline ever wants durable capability history, the refresh path is where the gap is.\n\nMerging once checks pass.

PR #280, opened 2026-08-16
feat(runtime): enforce Codex host bootstrap compatibility

YoungJinJung:
## Why

Advance #90 by enforcing the evidence-backed Codex bootstrap floor consistently before setup or runtime subscription, while leaving operational compatibility to capability negotiation.

## What changed

- add a dependency-free parser for supported Codex CLI/App Server identities
- reject missing, malformed, prerelease, and pre-0.147.0 hosts before integration mutation or App Server subscription
- publish the negotiated Codex host version through daemon status, CLI status, and doctor diagnostics
- document the bootstrap-only version policy with no upper stable allow-list
- cover install, runtime, and diagnostic failure paths

## Validation

- `ruff format --check .`
- `ruff check .`
- `mypy src tests`
- `pytest` (888 passed)


YoungJinJung:
## Review

Clean bootstrap-floor slice with the policy boundary drawn in the right place. No blocking findings.

- **The version-string-for-bootstrap-only policy is correctly held**: the floor gates only entry (setup mutation, App Server subscription), while everything operational stays with capability negotiation — and the explicit no-upper-allow-list stance means a newer stable Codex is never rejected by a table that would inevitably go stale. This matches the architecture doc's existing per-capability degradation principle rather than fighting it.\n- **All four enforcement points apply the same parser**, which is what prevents the classic drift where setup accepts what runtime rejects: binary detection (`--version`), `inspect` (defense in depth for an injected `CodexInstall`), the App Server client (validated after `initialize` but *before* the `initialized` notify and thread subscription — so an unsupported server is never subscribed to), and doctor against the *recorded* manifest version with concrete `upgrade + spotter setup codex` guidance.\n- **The parser is appropriately strict without being fragile**: only the three known identity prefixes (no guessing arbitrary text), anchored semver with optional patch, prerelease (`-`) rejected because the floor's evidence covers stable builds only, build metadata (`+homebrew`) accepted per semver precedence rules. Dependency-free, honoring the `websockets`-only runtime surface. The `order=True` dataclass gives correct tuple comparison for the floor check.\n- **Setup rejection happens before any host mutation** — pinned by the test asserting neither hooks nor manifest exist afterward — and a rejected App Server connect lands the client in `UNAVAILABLE` where the recovery loop's normal backoff + doctor guidance take over, rather than inventing a new failure mode.\n- Threading `host_version` → `ConnectionIdentity.server_version` → handshake → status/doctor rides the #279 plumbing naturally, and clearing it on disconnect keeps the identity honest.\n\nOne minor forward-compat note: existing manifests recorded real `codex --version` output so they should parse, but any manifest recorded from an exotic/dev build (e.g. a bare `development` string) will now FAIL doctor's integration check until re-setup — which is arguably exactly the diagnosis you want, just worth knowing it's a behavior change for previously-quiet configurations.\n\nMerging once checks pass.

PR #281, opened 2026-08-16
feat(runtime): diagnose mixed Codex upgrade windows

### LINKED ISSUE #90: Implement runtime configuration and upgrade compatibility lifecycle

YoungJinJung:
## Why

Complete the remaining mixed-version acceptance gap in #90. Setup already records the selected Codex host and runtime negotiation already exposes the connected App Server version, but doctor did not compare them.

## What changed

- report an explicit Codex host-version check from setup/runtime identities
- keep matched versions healthy and warn in either upgrade direction with `spotter setup codex` guidance
- prove setup records a newer stable Codex without rotating Spotter-owned integration identity
- add the mixed-version diagnostic matrix and mark the implemented lifecycle acceptance gates

## Validation

- `ruff format --check .`
- `ruff check .`
- `mypy src tests`
- `pytest` (892 passed)

Closes #90


YoungJinJung:
## Review

Small, correct close-out for #90. No findings.

- **Both sides of the comparison are normalized through the same #280 parser** — the recorded `codex-cli 0.147.0` and the runtime's canonical `0.147.0` both reduce to `str(CodexHostVersion)` — so the check compares semantic versions, not string spellings, and can't false-positive on prefix/format differences. That's the detail that makes a version-comparison diagnostic trustworthy.\n- **No double reporting**: an unparseable recorded version is silently skipped here because the integration check from #280 already FAILs it with its own guidance — one defect, one check.\n- **Warning in *both* upgrade directions is right**: a downgrade window (setup newer than connected server) is just as much a mixed-version state as the common upgrade case, and the matrix test pins all three cells.\n- **The companion integration test locks in the important non-behavior**: re-running setup after a stable Codex upgrade records the new host version *without* rotating the Spotter-owned integration generation (host version is correctly not a construction-fence input) and therefore without a restart — so reconciling the manifest after a Codex upgrade is cheap, exactly what the WARN guidance asks users to do.\n- The lifecycle acceptance checkboxes flipped to done are backed by the #275–#280 series plus this diagnostic, and the status row's next step correctly shifts from building to keeping the matrix in the release regression gate.\n\nThis completes the #90 arc coherently: bootstrap floor at the edges, capability negotiation in the middle, and now visibility across the seam between them. Merging once checks pass.

PR #282, opened 2026-08-16
test(runtime): cover terminal-before-steer-ack race

YoungJinJung:
## Why

Advance #22 by pinning the enqueue/consume race where the target turn becomes terminal after `turn/steer` is sent but before its RPC response reaches Spotter.

## What changed

- force the terminal notification to be consumed before the steer ACK
- prove the accepted request closes exactly once as durable `rpc_accepted_only`
- retain the precise `target_completed_without_observed_input` reason
- narrow the remaining #22 work to concurrent steer ordering and later-history contamination

## Validation

- `ruff format --check .`
- `ruff check .`
- `mypy src tests`
- `pytest` (893 passed)


YoungJinJung:
## Review

Good test, and built the right way. No findings.

- **The ordering is enforced deterministically, not probabilistically**: the fake server holds the steer ACK until it has *observed* (via `_wait_until` on the ingestor's records) that Spotter consumed the `turn/completed` notification, then releases the response. That makes the terminal-before-ACK interleaving a guaranteed property of every run rather than a timing coincidence — exactly what a race regression test must do to be worth having.\n- **The assertions pin the two things that matter** from the #219/#221 contracts: the accepted-but-unobserved steer closes as durable `rpc_accepted_only` with the precise `target_completed_without_observed_input` reason (not a generic stale bucket), and it closes **exactly once** — the `len(terminals) == 1` check is what guards the event-id dedup in `_record_control_event` against this specific interleaving, where both the acceptance-reconciliation path and the terminal path could plausibly try to write.\n- The status-row update is honest bookkeeping: it claims only what this test proves and narrows the remaining #22 work to concurrent steer ordering and compaction/later-turn contamination, keeping the implementation-vs-evidence separation the repo insists on.\n\nTest-only, deterministic, contract-anchored. Merging once checks pass.

PR #283, opened 2026-08-16
test(review): pin concurrent advisory ordering

YoungJinJung:
## Why

Advance #22 by pinning deterministic behavior when multiple live review results target the same active turn.

## What changed

- hold the first delivery open while a duplicate and a second job are submitted
- prove duplicate durable job identity does not create a second advisory
- prove distinct same-turn advisories receive unique control IDs and are delivered in queue order
- prove delivery calls never overlap
- narrow the remaining #22 evidence gap to compaction/later-turn history contamination

## Validation

- `ruff format --check .`
- `ruff check .`
- `mypy src tests`
- `pytest` (894 passed)


YoungJinJung:
## Review

Same deterministic construction quality as #282, aimed at the executor side of the #22 evidence gap. No findings.

- **The hold-open-then-flood pattern makes every claim structural**: parking the first delivery inside `steer` while a duplicate of the in-flight job and a distinct second job are submitted forces the exact contention window, then the assertions each guard a different mechanism — `controls` order pins FIFO queue ordering; `maximum_active_deliveries == 1` pins that the single queue worker serializes deliveries (no overlapping steers to one turn); unique control IDs pin per-advisory identity; and `review_job_config_pinned == 2` catches the subtle one — the duplicate submission of an *in-flight* job must be dropped by durable job identity (`_submitted`), not just by queue dedup, and must not even emit a second pin event.\n- Wiring the executor's `deliver` through the real `runtime.deliver_review_decision` with only the terminal `steer` faked keeps the delivery-side plumbing (control-ID minting, job attribution) under test rather than mocked away.\n- The status row again claims exactly what the test proves and honestly narrows the remaining #22 evidence to compaction/later-turn history contamination — the last cell of the matrix.\n\nTest-only, deterministic, one assertion per mechanism. Merging once checks pass.

Issue #23, opened 2026-08-11
Measure wrong-nudge susceptibility


YoungJinJung:
### Question / hypothesis

Can Main reject plausible but incorrect Spotter guidance when available evidence contradicts it? A useful supervisor must avoid both failure directions: being ignored when correct and being obeyed when wrong.

This experiment is now additionally important because the chosen live-delivery primitive, Codex `turn/steer`, is persisted/model-visible as user input rather than a distinct native supervisor role. A wrong Spotter nudge can therefore do more than distract Main momentarily: it can become the apparent latest user request or remain in later history/compaction.

### Implementation / failure references — study, do not depend on

Current upstream Codex reports expose exactly the failure mode this issue needs to measure:

- a steer can look like a standalone latest user request and make Codex drop the unfinished task: https://github.com/openai/codex/issues/29582
- a completed steered prompt can survive compaction and later be treated as the current task again: https://github.com/openai/codex/issues/30859
- a one-off goal steer can remain sticky across later goal continuations or even a new goal because it lives in ordinary thread history: https://github.com/openai/codex/issues/34602
- `clientUserMessageId` correlates steer input but does not currently propagate immutable provenance into descendant native tool calls: https://github.com/openai/codex/issues/36994

OpenClaw has also identified steering as a distinct privileged injection boundary that can bypass ordinary inbound security hooks unless explicitly screened:

- https://github.com/openclaw/openclaw/issues/78830

Spotter is not accepting arbitrary external steering input, so the threat model differs, but the architectural lesson is relevant: **steering is its own privileged actuator path and must have its own provenance/policy/evaluation boundary.**

These are references only. No OpenClaw dependency should be introduced.

### Experiment design

Use the mechanically scored task corpus (#21) and a replay instrument whose fidelity is measured in #42. From healthy or recoverable prefixes, inject a versioned corpus of plausible wrong nudges, including:

- false causal premise
- incorrect target/module
- unjustified scope expansion
- premature stop or validation advice
- advice that contradicts an explicit user constraint
- evidence request whose premise is already falsified
- "helpful" local optimization that conflicts with the task's real acceptance criterion

Compare against a neutral/control continuation. Once the live supervision contract exists in #22, test through the real `turn/steer` path so task-ownership/history effects are measured rather than simulated away.

### Make the wrong-nudge corpus immutable/versioned

Each wrong nudge should record:

```text
wrong_nudge_id
schema/version
source task/prefix family
nudge kind
false/unsupported premise
[...]

PR #284, opened 2026-08-16
test(runtime): fence expired advisories from later goals

### LINKED ISSUE #22: Deliver live VERIFY/NUDGE via turn/steer

YoungJinJung:
## Why

Complete the remaining #22 regression cell: a persisted/compacted Spotter steer can resurface as user-role history in a later turn and must not become that turn's active user goal.

## What changed

- continue the existing correlated-intervention fixture into a distinct second turn
- establish a new authoritative user goal in that turn
- replay the expired advisory identity against the later turn
- prove it is classified as `observed_outside_target` / `expired_advisory_visible`
- prove the later user goal remains authoritative

## Validation

- `ruff format --check .`
- `ruff check .`
- `mypy src tests`
- `pytest` (894 passed)

Closes #22


YoungJinJung:
## Review

The right final cell for the #22 matrix, and built as a continuation of the existing correlated-intervention fixture rather than a parallel scenario — so the contamination case is tested against the same lifecycle state a real session would carry. No findings.

- **The scenario models the actual hazard precisely**: the advisory's durable identity (`client_user_message_id = spotter:intervention:job-1`) resurfaces as a *user-role* prompt inside a distinct later turn that already has its own authoritative goal — which is exactly how persisted/compacted steer history would leak back. The assertions split the two protections correctly: classification (`outside_target` relation, `observed_outside_target` outcome, the precise `expired_advisory_visible` reason rather than a generic stale bucket) and, separately, **goal authority** — the later turn's goal remains `\"Document the completed fix\"`, proving the leaked advisory neither replaced nor reverted the goal to the old turn's \"Fix the original bug\".\n- Establishing the new goal *before* replaying the leak matters — it proves the fence protects an already-authoritative later goal, not merely that a leaked advisory fails to become a goal in a vacuum.\n- The status update is again evidence-honest: the `VERIFY`/`NUDGE` row now claims the diagnosed-and-fenced behavior this test proves, and the next step correctly moves from building regression cells to measuring advisory framing and wrong-nudge susceptibility in #23/#34 — the right hand-off now that #282/#283/#284 close the enqueue/consume race, ordering, and contamination cells.\n\nWith this, #22's regression surface is covered end to end: terminal-before-ACK, concurrent same-turn ordering/dedup, and later-turn history contamination. Merging once checks pass.

PR #285, opened 2026-08-16
feat(experiments): freeze wrong-nudge corpus

YoungJinJung:
## Why

Wrong-nudge susceptibility needs immutable, falsifiable experiment inputs before paid control/raw/advisory arms can produce trustworthy evidence.

Refs #23.

## What changed

- add a dependency-free, versioned wrong-nudge manifest validator with strict schema, enum, containment, identity, and hash checks
- freeze seven v1 manifests covering every required wrong-nudge family, each with the false premise and injection-time contradicting evidence
- document the no-rewrite/new-version rule and distinguish corpus input from experimental evidence
- update current status to identify equivalent-prefix arm execution as the next evidence gate

## Validation

- `ruff format --check .`
- `ruff check .`
- `mypy src tests`
- `pytest` (898 passed)


YoungJinJung:
## Review

Good foundation for the #23 susceptibility runs. The design choices that make this corpus scientifically usable are all present. No blocking findings.

- **Falsifiability is built into the schema, not left to convention**: every manifest must declare both the false premise *and* the contradicting evidence already visible to Main at injection time, plus a typed expected healthy response. That's what turns \"did Main resist the nudge\" from a judgment call into a checkable outcome — the healthy response taxonomy (refute-from-evidence / verify-then-refute / preserve-user-constraint) maps cleanly onto distinct failure modes worth measuring separately.\n- **The freeze mechanism follows the established evidence discipline**: per-manifest sha256 pinned in the set file, validated in CI by the repo-corpus test (so the seven recorded digests are machine-verified, not trusted), duplicate-ID rejection, set-vs-manifest identity cross-check, and the documented no-rewrite/new-version rule. An observed item can't be silently edited after arms have run against it — the property that makes later results attributable to exact inputs.\n- **The validator is appropriately paranoid for untrusted-ish input**: schema identity per the #47 pattern, enum membership with the allowed list in the error, containment check via `resolve()` + `is_relative_to` (which also catches symlinks escaping the corpus root), and `bool`-excluding integer checks. Dependency-free, honoring the runtime-surface rule.\n- **The corpus content is honest about what it is**: seven families spanning the full `NudgeKind` enum (pinned by set-equality in the test, so adding a kind without a fixture fails CI), each anchored to an existing fixture task family, and the README states plainly that this is experiment *input*, not evidence Main rejects bad supervision — with the equivalent-prefix arms named as the actual evidence gate. Exactly the implementation-vs-evidence separation the repo's rules require.\n- The premise≠evidence sanity assertion is a nice guard against a degenerate manifest that would make the arm unfalsifiable.\n\nMerging once checks pass.

PR #286, opened 2026-08-16
feat(experiments): plan wrong-nudge framing arms

YoungJinJung:
## Why

The frozen #23 corpus needs mechanically distinct framing conditions with identical source provenance before real-steer execution can compare task replacement and bad-supervision resistance.

Refs #23.

## What changed

- build neutral control, raw wrong imperative, Spotter-scoped advisory, and VERIFY-first arm plans from one corpus item
- pin every arm to the same source session/step, prefix identity, environment fingerprint, manifest hash, and payload version
- make neutral control explicitly emit no steer payload and reject incomplete prefix provenance
- document that execution must independently fork and preflight arms before delivery

## Validation

- `ruff format --check .`
- `ruff check .`
- `mypy src tests`
- `pytest` (904 passed)


YoungJinJung:
## Review

Right shape for the arm-planning layer, and the provenance discipline carries through from #285. No blocking findings.

- **Equivalence is enforced structurally, not by convention**: all four arms are minted from one call with a single source session/step, prefix identity, environment fingerprint, manifest hash, and payload version — and the test asserts each of those is a singleton set across arms. Combined with the manifest hash coming from the *already-verified* set entry (threaded through `WrongNudge` rather than recomputed later), a result row can always be traced to the exact frozen input that produced it.\n- **The neutral control emitting `payload=None`** — rather than an empty-string steer — is the correct baseline: the condition is \"no delivery at all\", and making that unrepresentable as an accidental empty steer removes a whole class of execution bugs before the execution slice exists.\n- **The advisory framing matches the production contract**: the supervision header (\"not a new user requirement… preserve the user's original task\") is the same posture the live `turn/steer` advisories carry, so the SPOTTER_ADVISORY arm measures the framing Spotter actually ships, not a synthetic one.\n- **Provenance validation refuses incomplete prefixes** (empty/whitespace strings, negative or boolean steps) at plan time — cheap fail-fast that keeps garbage out of the durable experiment record.\n- The README/status updates keep the boundary honest: these are *plans*; forking, preflight, and delivery remain the explicitly named next slice.\n\nOne observation worth carrying into the analysis phase, non-blocking: the VERIFY_FIRST arm embeds the `false_premise` while RAW/ADVISORY embed the `payload_template`, so that arm differs from the others in *content basis*, not only in framing wrapper. That's a legitimate design — VERIFY-vs-NUDGE is a real strategy choice Spotter makes, and testing it matters — but it means the four-way comparison isolates pure framing only across NEUTRAL/RAW/ADVISORY; conclusions about VERIFY_FIRST should be phrased as strategy comparison, not framing comparison. Worth one line in the eventual experiment writeup so the distinction doesn't get lost.\n\nMerging once checks pass.

PR #287, opened 2026-08-16
feat(experiments): preflight wrong-nudge forks

YoungJinJung:
## Why

Wrong-nudge framing plans are not comparable unless every condition starts from an independent worktree restored from the same recorded prefix and environment.

Refs #23.

## What changed

- create one Git-aware replay fork per neutral/raw/advisory/VERIFY-first condition
- pin prepared arms to their fork session, worktree, rollout, and durable fork manifest
- refuse duplicate sessions, shared worktrees, missing manifests/provenance, source-environment drift, prefix mismatch, or restored-environment mismatch before delivery
- pass declared environment resources, variables, and venv/cache inputs through every fork
- reject invalid source identity before creating any fork

## Validation

- `ruff format --check .`
- `ruff check .`
- `mypy src tests`
- `pytest` (913 passed)


YoungJinJung:
## Review

Correct bridge between the #286 arm plans and the proven replay-fork machinery. No blocking findings.

- **Reusing `fork()` instead of a bespoke experiment fork path is the right call** — every equivalence guarantee built up in the #42 replay series (Git-aware worktrees, declared environment resources/variables/venv inputs, source-environment preflight, durable fork manifests) is inherited rather than reimplemented, and the test asserts the environment declarations actually pass through to all four calls.\n- **The preflight refusal set is the complete comparability contract**: duplicate fork sessions, shared worktrees (compared on *resolved* paths, catching symlinked aliases), any non-MATCHED source-environment preflight, missing manifests, and missing-or-mismatched prefix/environment fingerprints — with distinct machine-readable reason codes for each, consistent with the classified-refusal style of the fork-coverage work. A single drifted fork poisons the whole batch rather than silently degrading one arm.\n- **Condition↔fork pairing is positional over two same-ordered iterations of `FramingCondition`**, which is safe even against future reordering because preflight has already proven all four forks byte-equivalent in prefix and environment — any pairing of equivalent forks is valid, so the implicit coupling has no failure mode.\n- **Invalid source identity rejects before any fork is created** (pinned by the assert-raising fake), so garbage input can't leave artifacts.\n\nOne non-blocking note: when preflight refuses after forking (e.g. the fourth fork drifts), the three-plus created forks are abandoned rather than torn down. They're fully owned/registered artifacts, so the existing prune/purge machinery reclaims them and nothing leaks invisibly — but a batch of refused experiments will accumulate worktrees until then. If refusals turn out to be common in practice, best-effort cleanup on the refusal path would keep the workspace tidy; as long as they're rare, purge-later is fine.\n\nMerging once checks pass.

PR #288, opened 2026-08-16
feat(app-server): expose turn start

YoungJinJung:
## Why

The #23 real-steer experiment runner must start the neutral continuation on each prepared replay fork before it can inject a wrong nudge into that exact active turn.

Refs #23.

## What changed

- add a typed `start_turn` adapter to the Codex App Server client
- keep JSON-RPC method/shape details behind the adapter boundary
- override both cwd and runtime workspace roots to the prepared fork worktree
- support durable client user-message correlation for the continuation input
- reject empty turn input, cwd, or correlation IDs before dispatch

## Validation

- `ruff format --check .`
- `ruff check .`
- `mypy src tests`
- `pytest` (916 passed)

### DIFF
--- a/src/spotter/app_server.py
+++ b/src/spotter/app_server.py
@@ -289,6 +289,33 @@ class CodexAppServerClient:
     async def resume_thread(self, thread_id: str) -> Mapping[str, Any]:
         return await self._object_request("thread/resume", {"threadId": thread_id})
 
+    async def start_turn(
+        self,
+        thread_id: str,
+        text: str,
+        *,
+        cwd: str | None = None,
+        client_user_message_id: str | None = None,
+    ) -> Mapping[str, Any]:
+        """Start a turn without exposing Codex JSON-RPC shapes to callers."""
+
+        if not text.strip():
+            raise ValueError("turn input must be non-empty")
+        params: JsonObject = {
+            "threadId": thread_id,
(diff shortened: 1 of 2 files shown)


YoungJinJung:
## Review

Small, correct adapter slice. No findings.

- **Adapter boundary respected**: `turn/start`'s JSON-RPC method name and input shape stay inside `CodexAppServerClient`, so the experiment runner will consume a typed call rather than Codex wire details — the same discipline the existing `steer`/`interrupt` adapters follow, and rule 2 of the working rules.\n- **Overriding both `cwd` and `runtimeWorkspaceRoots` to the fork worktree is the load-bearing detail**: setting only `cwd` would let runtime workspace resolution leak back to the source checkout, which is precisely the absolute-path-drift class of contamination the #42 replay work classifies as disqualifying. Coupling them in one parameter makes the safe configuration the only expressible one.\n- **`clientUserMessageId` correlation** reuses the same durable-identity mechanism the advisory pipeline uses (#218/#284), so the neutral continuation input will be exactly attributable — and later distinguishable from injected nudges — in the journal.\n- Empty text/cwd/correlation-ID rejection happens before dispatch (pinned by tests), and the wire-shape test asserts the full params object byte-for-byte, so a silent protocol drift fails loudly.\n\nMerging once checks pass.

PR #290, opened 2026-08-16
feat(experiments): deliver wrong nudges via steer

YoungJinJung:
## Why

#23 must measure the actual Codex `turn/steer` history and task-ownership behavior, not simulate wrong guidance as an ordinary continuation prompt.

Refs #23.

## What changed

- resume each independently prepared fork and start the same neutral continuation through App Server `turn/start`
- invoke real `turn/steer` only for raw/advisory/VERIFY-first arms; neutral control sends no steer
- assign separate durable correlation IDs to continuation and steer inputs
- revalidate all prepared-arm provenance before any connection
- classify control, RPC acceptance, rejected/stale delivery, and unknown acceptance separately from observed turn completion
- filter shared App Server notifications by exact fork thread and turn with a bounded completion deadline

## Validation

- `ruff format --check .`
- `ruff check .`
- `mypy src tests`
- `pytest` (919 passed)


YoungJinJung:
## Review

The delivery runner carries the evidence discipline from the runtime work into the experiment harness, which is exactly what makes the eventual measurements trustworthy. No blocking findings.

- **The error taxonomy is the important design and it's right**: a control-rejected or stale steer is `DELIVERY_FAILED_OR_STALE` (with the precise reason code), an RPC rejection likewise, but an ambiguous transport failure is `ACCEPTANCE_UNKNOWN` — and that arm **skips the completion wait entirely**, because scoring a turn whose input set is unknown would poison the comparison. Meanwhile a stale steer still observes completion (`completion_observed=True`), correctly recording \"the turn ran *without* the nudge\" as distinct from both successful delivery and Main resistance. This is the `rpc_accepted_only`/#231 do-not-invent-outcomes discipline applied to experiment arms, and the dedicated test pins the stale case against misclassification as compliance.\n- **Separate durable correlation IDs for continuation vs steer** (`spt-exp-start-*` / `spt-exp-steer-*`) mean the journal can attribute exactly which input was the neutral continuation and which was the injected nudge — feeding the same correlation machinery the advisory pipeline already uses.\n- **`_preflight_prepared` revalidates everything before any connection**: full condition coverage, distinct sessions and resolved worktrees, and a single provenance tuple across arms — so prepared arms that were tampered with or mixed between batches refuse before touching the App Server (pinned by the assert-raising factory test).\n- **Completion filtering is correctly paranoid for a shared notification stream**: exact threadId+turnId match, bounded deadline, and a protocol error (not a silent default) when `turn/completed` omits status. Timeout/transport/protocol failures each append a distinct diagnostic rather than collapsing into one bucket.\n- Reusing `CONTROL_PROMPT` keeps the neutral continuation byte-identical to the existing experiment harness's, and the `turn/start` cwd pinning from #288 puts each continuation in its own fork worktree.\n\nOne observation, non-blocking: arms execute **sequentially**, so later arms run minutes after earlier ones — a temporal confound (backend/model drift across the batch window) that parallel execution would avoid. Since each arm is an isolated fork/thread, sequential is the safer first implementation and the window is small relative to the effect sizes #23 cares about; just worth stating in the experiment writeup, and revisiting if batches grow.\n\nMerging once checks pass.

PR #291, opened 2026-08-16
feat(experiments): score wrong-nudge arms

YoungJinJung:
## Why

Wrong-nudge delivery is now exercised through the real App Server path, but #23 still needs objective task outcomes that cannot be confused with steer acceptance or failure.

Refs #23

## What changed

- reuse the frozen task-corpus checks against each arm worktree after its terminal event
- reject delivery/fork/task provenance mismatches before writing results
- keep stale/rejected delivery separate from the mechanical PASS/TASK_FAIL classification
- leave arms without observed completion UNJUDGEABLE instead of racing checks against a live turn
- fsync a versioned header, each arm row, and completion row with task, prefix, fork, delivery, and check provenance
- expose the existing experiment append and task-check boundaries for reuse

## Validation

- `ruff format --check .`
- `ruff check .`
- `mypy src tests`
- `pytest` (922 passed)


YoungJinJung:
## Review

The scoring layer keeps the three classifications that must never be conflated — delivery outcome, mechanical task outcome, and judgeability — properly orthogonal. No blocking findings.

- **`run_task_checks` is a pure extraction, not a reimplementation**: the existing arm-classification logic (timeout/error/pass/fail over required checks) moves verbatim from `_run_task_arm` into a shared boundary, so wrong-nudge scoring is mechanically identical to the established task-corpus scoring — the property that makes cross-experiment comparisons meaningful. Same for exposing `append_experiment_result`: the locked, fsynced, schema-checked append path is reused rather than duplicated.\n- **`UNJUDGEABLE` instead of racing a live turn** is the right refusal: an arm without an observed terminal event gets no checks run (empty tuple, explicit `turn_completion_not_observed` diagnostic), never a speculative score against a possibly-still-mutating worktree.\n- **Delivery and mechanical outcomes stay independent columns**: the test's stale-delivery arm is scored `TASK_FAIL` *and* carries `DELIVERY_FAILED_OR_STALE` + `no_active_turn` — so analysis can stratify \"turn ran without the nudge\" separately from \"nudge landed and the task failed\", exactly the distinction the README warns about.\n- **All three provenance gates fire before any write**: prepared-arm consistency (now including `source_task`), delivery coverage + per-arm fork-session match, and task-vs-corpus identity — with the test proving a mismatch leaves no partial result file. The refuse-on-existing-path check preserves the no-rewrite discipline for result files themselves.\n- **Result rows are self-contained for later audit**: task manifest and fixture hashes, fork manifest, both correlation IDs, delivery diagnostics, and a nested `wrong_nudge_result_schema_version` riding inside the existing experiment-result schema so legacy readers stay compatible.\n\nOne non-blocking note: `~/.spotter/experiments/wrong-nudges/` is a new durable location. The `purge --data` inventory classifies unknown entries conservatively (reported, never deleted), so nothing is at risk — but after real runs, purge previews may start reporting these files as ambiguous and exiting non-zero until the data inventory learns the new family. Worth folding into the inventory's known-store set in a follow-up so experiment artifacts get first-class schema-proven handling like the rest.\n\nMerging once checks pass.

PR #292, opened 2026-08-16
feat(experiments): record wrong-nudge annotations

YoungJinJung:
## Why

Mechanical task outcomes alone cannot explain whether Main preserved the user task, refuted a bad steer with evidence, or merely acknowledged it. #23 needs those secondary judgments without rewriting primary experiment evidence.

Refs #23

## What changed

- add an independently versioned wrong-nudge annotation store beside experiment results
- fingerprint the exact arm result and retain task ownership, conservative behavior relations, rater identity, notes, and non-exclusive susceptibility classes
- require concrete evidence references for evidence-backed refutation
- reject semantic labels for stale/rejected or incomplete delivery and reject classifications inconsistent with mechanical PASS/TASK_FAIL
- initialize shared result headers once under the existing locked, fsynced append boundary
- prove nested `experiments/wrong-nudges` artifacts are already SAFE_OWNED by the data inventory

## Validation

- `ruff format --check .`
- `ruff check .`
- `mypy src tests`
- `pytest` (935 passed)


YoungJinJung:
## Review

The annotation layer encodes label semantics as machine-checked invariants rather than rater conventions, which is what will make the eventual susceptibility rates defensible. No blocking findings.

- **The consistency lattice is the core value and it's thorough**: control arms cannot carry susceptibility classes; semantic robustness labels require `RPC_ACCEPTED` delivery (you cannot claim Main resisted a nudge that never landed — the exact confusion #290's outcome taxonomy exists to prevent); `REFUTED_WITH_EVIDENCE` demands concrete evidence refs *and* `REFUTED_AND_CONTINUED` ownership *and* excludes contradictory compliance classes; `TASK_REPLACED` is biconditional with its ownership outcome; and compliance-degraded/recovered classes must agree with the mechanical TASK_FAIL/PASS. Every one of these is pinned by a rejection test that also proves no partial file is written.\n- **Annotations are structurally incapable of rewriting run evidence**: separate JSONL, exact-result sha256 fingerprint so a drifted result invalidates `annotation_matches`, corrections append as new rows with rater identity (the #224 no-rewrite pattern), and an unknown schema version refuses rather than reinterprets (#47 pattern).\n- **`initialize_experiment_result` is the right generalization** — idempotent header under the existing lock, with a real ThreadPoolExecutor test proving one header under 20 concurrent initializers, and the existing `run_experiment` callers untouched.\n- **The #291 review note is closed here**: the data-inventory test now proves `experiments/wrong-nudges/*.jsonl` classifies as schema-proven `experiment_result` rather than ambiguous, so purge previews stay clean after real runs. Appreciated.\n- `getuser()` default with a trimmed, bounded explicit override keeps rater identity durable without inventing an identity system.\n\nMerging once checks pass.

PR #293, opened 2026-08-16
feat(experiments): report wrong-nudge susceptibility

YoungJinJung:
## Why

#23 now records delivery, mechanical outcomes, and secondary annotations, but the evidence is not decision-usable until equivalent-prefix control/nudge pairs are aggregated with explicit coverage.

Refs #23

## What changed

- reload versioned wrong-nudge mechanical rows for offline analysis
- revalidate task, corpus, prefix, and environment provenance before pairing each framing arm with its neutral control
- report delivery, completion, mechanically judgeable pair, and semantic annotation coverage per framing condition
- quantify control-pass to nudge-fail harm, evidence-backed refutation, compliance, recovery, task replacement, constraint loss, and persistent contamination
- select the latest correction per rater while exposing rater conflicts, stale labels, and orphan labels instead of silently resolving them
- refuse future in-memory or durable schemas

## Validation

- `ruff format --check .`
- `ruff check .`
- `mypy src tests`
- `pytest` (941 passed)


YoungJinJung:
## Review

This is the layer where subtle denominator mistakes would silently corrupt the #23 conclusions, and the design avoids all the ones I checked for. No blocking findings.

- **The denominator cascade is explicit and correct**: attempts → delivery accepted → completion observed → mechanically judgeable *pairs* (both arms completed with a PASS/TASK_FAIL classification) → control-pass→nudge-fail harm over judgeable pairs only. Harm is never computed over arms whose control was unjudgeable, and the render shows every rate with its own numerator/denominator so a reader can't mistake a 1/1 for strong evidence.\n- **Annotation handling refuses to launder disagreement**: latest-per-rater selection (dict-overwrite over file order, so append-only corrections resolve correctly per rater), then *current* raters must agree on (ownership, classification-set) or the arm counts as a conflict and drops out of labeled coverage — surfaced, not majority-voted. Stale labels (fingerprint no longer matching the result) and orphan labels (no matching result) are counted separately rather than silently dropped. This is the double-label agreement discipline from the #33 metrics work applied here.\n- **Pair provenance is re-verified at read time** across the full eleven-field tuple (corpus identity, prefix, environment, task/fixture hashes) even though the write path already enforced it — the right posture for offline analysis of durable files that could have been mixed from different runs, with duplicates refused at both load and measure layers.\n- **Schema discipline both durable and in-memory**: rows and annotation objects with future versions refuse rather than reinterpret, per the #47 pattern; the loader accepts partial files (no invented rows for missing arms) so interrupted runs remain analyzable with honest coverage.\n- The semantic-eligibility gate (accepted delivery + observed completion) mirrors #292's write-side rules, so read-side counts can never include labels the write side would have refused.\n\nThe status row's next step — persistence-cohort execution and operator-facing commands, then actually running the measurements — keeps the implemented-vs-evidence line clean: this PR builds the instrument, and nothing here claims a susceptibility result yet. Merging once checks pass.

PR #294, opened 2026-08-16
feat(cli): report wrong-nudge susceptibility

YoungJinJung:
## Why

Wrong-nudge measurements already persist mechanical results and independent annotations, but operators need a stable read-only boundary to render those durable artifacts without rerunning paid agent arms.

Refs #23

## What changed

- add `spotter wrong-nudge report <results.jsonl>` with optional `--annotations <annotations.jsonl>`
- render the existing coverage-aware susceptibility report from durable rows
- return an actionable non-zero error for unreadable or invalid evidence
- lazy-load experiment reporting so dependency-free bundled Hook entrypoints remain isolated
- document the offline reporting workflow and current implementation status

## Validation

- `ruff format --check .`
- `ruff check .`
- `mypy src tests`
- `pytest` (944 passed)


YoungJinJung:
## Review

Thin, correct operator boundary over the #293 instrument. No findings.

- **Strictly read-only**: the command loads durable rows and renders — it cannot rerun paid arms, mutate results, or write anything, which is exactly the property an operator-facing report command over experiment evidence needs. All validation (schema refusal, duplicate detection, pair-provenance recheck, conflict/stale/orphan accounting) comes from the #293 layer unchanged, so the CLI adds no second interpretation of the data.\n- **Error handling matches the house style**: unreadable or invalid evidence produces an actionable one-line stderr message and exit 1 — pinned by a test feeding it literal non-JSON — rather than a traceback.\n- **Argument plumbing is complete in both directions**: `wrong-nudge` requires `report <results.jsonl>`, and `--annotations` outside `wrong-nudge report` is a parser error rather than a silently ignored flag, consistent with how `--resume`/`--dry-run` are fenced.\n- **Lazy imports for the reporting modules** keep the dependency-free bundled Hook entrypoint isolation intact — the same pattern the CLI already uses for heavy paths.\n- The end-to-end test goes through real durable files including a genuine `add_wrong_nudge_annotation` round trip, so the CLI is proven against the actual write-side format, not a hand-built fixture of it.\n\nStatus honestly narrows the remaining #23 work to persistence-cohort execution and the paid run command. Merging once checks pass.

PR #295, opened 2026-08-16
feat(cli): run wrong-nudge experiments

YoungJinJung:
## Why

The wrong-nudge instrument had preparation, real steer delivery, mechanical scoring, and offline reporting, but no operator boundary for intentionally launching a frozen paid run.

Refs #23

## What changed

- add `spotter wrong-nudge run <wrong-nudge-set.toml>` for one exact frozen corpus item
- require an explicit task set, source session/step, controlled App Server endpoint, and `--run` paid-execution confirmation
- validate the immutable nudge corpus and frozen scorer preflight before creating any experiment forks
- execute the existing four equivalent-prefix framing arms, persist mechanical results, and print the durable result path
- reject run-only arguments on report and unrelated commands
- document the operator workflow and narrow the remaining #23 implementation gap to persistence cohorts

## Validation

- `ruff format --check .`
- `ruff check .`
- `mypy src tests`
- `pytest` (948 passed)


YoungJinJung:
## Review

Correct operator boundary for the paid path. No findings.

- **The paid gate follows the established convention**: `--run` is a hard parser requirement (pinned by test), matching how `spotter experiment`/`tasks run` already treat paid execution — an operator cannot stumble into four agent arms by omitting a flag.\n- **Every validation runs before any fork exists**: corpus-set hash validation, exact `wrong_nudge_id` membership, full task-set preflight (any non-READY task aborts with the per-task detail), and source-task membership — and the unknown-nudge test proves `prepare` is never reached, so a bad invocation costs nothing and leaves no artifacts. This front-loads all the free checks before the only expensive step.\n- **`run_wrong_nudge_experiment` is pure composition** of the already-reviewed prepare → deliver → score layers (order pinned by test, arguments threaded through verbatim including environment declarations) — no new safety-relevant logic rides in with the CLI, so everything verified in #287/#290/#291 holds unchanged.\n- **Argument fencing is bidirectional and complete**: run requires its five inputs, run-only flags are parser errors on `report` and on unrelated commands, `--annotations` stays report-only — no flag is ever silently ignored.\n- Error families (corpus, experiment, task-corpus, replay, snapshot, OS) all map to one actionable stderr line + exit 1, and the success path prints the durable result path so the operator can go straight to `wrong-nudge report`.\n\nWith this, the #23 instrument is operator-complete end to end: freeze → prepare → deliver → score → annotate → report, all behind explicit paid confirmation. Only persistence cohorts remain before the actual measurements. Merging once checks pass.

PR #296, opened 2026-08-16
feat(experiments): run wrong-nudge persistence cohorts

YoungJinJung:
## Why

A wrong steer can survive as ordinary user-role history and be promoted again at a later model boundary. The primary four-arm run ends after the injected turn, so #23 still needs a durable follow-up cohort before that failure mode can be annotated or measured.

Refs #23

## What changed

- add `spotter wrong-nudge persist <results.jsonl> --endpoint ... --run` as an explicitly paid follow-up boundary
- refuse partial, provenance-mismatched, undelivered, uncompleted, future-schema, or unavailable-worktree source arm sets before connecting
- start the same versioned original-task follow-up on each control/raw/advisory/VERIFY fork at the next model boundary
- persist each arm immediately under an independent wrong-nudge persistence sub-schema, pinned to the exact source-result fingerprint, fork/worktree, client message ID, and completion observation
- keep start rejection, unknown acceptance, transport loss, timeout, and protocol failure visible instead of inferring semantic persistence
- add strict durable reloads for later human outcome annotation/reporting
- document the lifecycle contract and operator command

This PR records execution evidence only. Semantic outcomes such as harmless history versus stale-advisory re-promotion remain a separate annotation/report step.

## Validation

- `ruff format --check .`
- `ruff check .`
- `mypy src tests`
- `pytest` (954 passed)


YoungJinJung:
## Review

The design question for a persistence cohort is whether the follow-up can contaminate its own measurement, and this implementation avoids that. No blocking findings.

- **The follow-up prompt is neutral and versioned** — \"Continue the user's original task and constraints\", identical across all four arms, containing no nudge content and no reference to the prior steer. That's the only valid probe for the hypothesis (does *history alone* re-promote expired advice at the next model boundary?); any arm-specific phrasing would confound it. Versioning the prompt and validating the version on every durable row means a future prompt change can never silently mix cohorts.\n- **Preflight refuses everything before a token is spent**: exactly one complete four-condition arm set from one experiment, the full eleven-field provenance singleton, exact source schema versions, per-condition *expected* delivery outcomes (control → `CONTROL_NO_STEER`, nudge arms → `RPC_ACCEPTED`) plus observed completion, and worktree availability. A follow-up on an arm whose nudge never landed would measure nothing, and the preflight makes that unrepresentable rather than a footnote.\n- **Each row is pinned to the exact source-result fingerprint** (reusing #292's fingerprint), so annotation/reporting can join follow-ups to precisely the mechanical result they extend — and a re-scored or drifted source breaks the join loudly instead of silently.\n- **Execution evidence only, stated and enforced**: the outcome taxonomy (pre-delivery / start-rejected / start-accepted / acceptance-unknown, with completion and distinct timeout/transport/protocol diagnostics separate) records what happened operationally; whether the resurfaced advice was harmless history or a re-promoted stale advisory is explicitly deferred to the annotation step. No semantic persistence is inferred here — the same discipline as #290/#291.\n- Durable handling follows the established pattern throughout: per-row fsync via the shared append boundary, refuse-existing-path, deterministic arm ordering, strict loader with duplicate and future-sub-schema refusal, and the lifecycle compatibility table updated with the new sub-schema. CLI fencing (subject+endpoint+`--run` required; inapplicable flags rejected) matches `run`'s conventions.\n\nSame non-blocking observation as #290 applies: arms execute sequentially, so the batch-window temporal confound is worth one line in the writeup. With this, the full #23 instrument — freeze through persistence — is in place; only the persistence-outcome annotation/report coverage remains. Merging once checks pass.

PR #297, opened 2026-08-16
feat(experiments): annotate wrong-nudge persistence

YoungJinJung:
## Why

Operational persistence rows show whether a follow-up was delivered and completed, but cannot distinguish harmless historical context from stale advisory re-promotion or new-goal contamination. This adds the independent human evidence layer required by #23 without rewriting execution results.

Refs #23

## What changed

- add a separately versioned wrong-nudge persistence annotation schema pinned to the exact follow-up-result fingerprint
- classify `NO_PERSISTENCE`, `HISTORICAL_BUT_HARMLESS`, `STALE_ADVISORY_REPROMOTED`, `NEW_GOAL_CONTAMINATED`, and `UNJUDGEABLE` outcomes
- require accepted, completed delivery plus trajectory evidence for semantic outcomes
- reject contamination claims on neutral controls and refuse foreign/future schemas
- document the evidence contract and narrow the remaining status item to report integration

Persistence annotation aggregation in `spotter wrong-nudge report` intentionally remains the next small PR.

## Validation

- `ruff format --check .`
- `ruff check .`
- `mypy src tests`
- `pytest` (960 passed)


YoungJinJung:
## Review

This applies the #292 annotation discipline to the persistence layer with the same rigor, and the outcome taxonomy is the right shape for the question. No blocking findings.

- **The five-outcome taxonomy separates exactly what the follow-up turn can reveal**: no persistence, harmless historical presence, stale-advisory re-promotion, and new-goal contamination are genuinely distinct failure severities (the #284 fence test targets the third; the fourth is worse), and `UNJUDGEABLE` is the only outcome permitted when the follow-up wasn't operationally judgeable — so an operational gap can never be laundered into a semantic claim in either direction.\n- **The evidence gates mirror #292 consistently**: semantic outcomes require `START_ACCEPTED` + observed completion *and* concrete trajectory evidence refs; neutral controls cannot claim wrong-nudge contamination (there was no nudge to persist); notes are mandatory; the source result's schema versions are validated before accepting a label against it.\n- **Fingerprint pinning, append-only corrections with rater identity, idempotent header initialization, and future-sub-schema refusal** all reuse the established mechanisms (`initialize_experiment_result`, the shared append boundary, the exact-result sha256 join) rather than reinventing them — annotations remain structurally incapable of rewriting execution evidence.\n- The lifecycle compatibility table gains the annotation sub-schema alongside the persistence one, keeping the durable-family inventory complete, and the status row honestly narrows the remaining work to report integration — explicitly deferred as the next small PR rather than bundled here.\n\nMerging once checks pass.

PR #298, opened 2026-08-16
feat(metrics): report neutral replay failures

YoungJinJung:
## Why

Issue #42 requires neutral replay noise, environment preflight failures, and infrastructure failures to remain separately measurable. The immediate experiment summary exposed those failures, but the durable aggregate metrics projection dropped them, so later qualification reports could silently lose the reason neutral arms were unjudgeable.

Refs #42

## What changed

- preserve exact non-matched neutral preflight categories at pair granularity
- report neutral infrastructure-failure classifications at arm granularity
- render both with explicit pair/arm denominators alongside mechanical disagreement
- reject malformed persisted `environment_preflight` fields instead of silently ignoring them
- update status to keep broader natural-failure/noise execution as the remaining #42 evidence step

This does not change the existing v1 NO-GO qualification or claim a measured representative noise floor.

## Validation

- `ruff format --check .`
- `ruff check .`
- `mypy src tests`
- `pytest` (962 passed)


YoungJinJung:
## Review

Small and correct. This closes a real evidence-loss gap: the immediate experiment summary knew *why* neutral arms were unjudgeable, but the durable aggregate projection dropped the reason — so a later #42 qualification report reading only the metrics could not distinguish \"replay is noisy\" from \"the environment restore failed.\" Those demand opposite responses, and conflating them would have corrupted the noise-floor interpretation. No findings.

- **Granularity choices are right**: preflight failures count once per *pair* (a set-union of non-MATCHED categories, so a pair with both arms failing the same way isn't double-counted), while infrastructure failures count per *arm* (each arm's classification is an independent event) — and the render states both denominators explicitly (`/N pairs`, `/2N arms`), continuing the #293 discipline of never letting a rate float free of its base.\n- **UNJUDGEABLE is correctly excluded from infrastructure categories** — it's an observation gap, not an infrastructure event — so the three failure families (preflight, infrastructure, unjudgeable) stay separately measurable as #42 requires.\n- **Legacy tolerance vs malformed rejection is split properly**: an absent `environment_preflight` (older rows) loads as `None` and simply contributes nothing, while a present-but-non-string value raises with the file path — silent coercion of malformed evidence is exactly what the fail-visibly rule forbids in a metrics reader.\n- Sorted category dicts keep the render deterministic, and the status update preserves the v1 NO-GO verdict while claiming only the reporting capability, not a measured noise floor.\n\nMerging once checks pass.

Issue #304, opened 2026-08-17
Productize ordinary Codex launch and validate concurrent runtime isolation


YoungJinJung:
## Problem / outcome

The configured explicit `codex --remote <endpoint>` path proves that a Codex TUI and Spotter can observe and control the same real thread. It is not yet an ordinary product launch contract:

- plain `codex` does not discover the configured external App Server;
- starting Spotter at the first Hook is too late for complete observation/control;
- the lifecycle checklist still leaves ordinary managed launch and concurrent session isolation incomplete;
- multi-TUI, reconnect, and embedded-server degraded behavior remain unresolved product boundaries;
- no open Issue currently owns these remaining Runtime exit conditions.

Define and validate the smallest honest user-facing launch contract without making Spotter implicitly own or stop a shared App Server.

## Scope

- choose and document one supported managed launch surface, preferring a thin wrapper/explicit command around the already validated `--remote` path before adding new runtime machinery;
- ensure `spotterd` and the configured external endpoint are ready before the TUI attaches;
- preserve explicit endpoint identity, capability negotiation, and external process ownership;
- validate at least two concurrent TUI/thread attachments with isolated ThreadState, reviewer jobs, control targets, and journals;
- validate daemon/App Server disconnect, reconnect, and restart behavior without fabricated continuity;
- define the embedded/plain-`codex` path as working, degraded, or unsupported with actionable `status`/`doctor` output;
- update setup/status/doctor/lifecycle documentation and regression coverage for the chosen contract.

## Constraints

- Do not claim ownership of or stop a shared App Server process.
- Do not hide which endpoint/process is shared.
- Do not use a Hook to retrofit the observation/control plane after the TUI has selected an embedded server.
- Do not weaken fail-open deterministic `PreToolUse` enforcement when observation/control are degraded.
- Do not add a general service orchestration abstraction unless the existing service/endpoint boundaries cannot support the chosen contract.

## Validation matrix

- clean install → setup → managed launch → active observation/control;
- second concurrent TUI/thread remains isolated;
- TUI resume/late attachment marks pre-attachment coverage unknown;
- daemon restart recovers durable state and fences stale control targets;
- App Server reconnect creates a new epoch and records the observation gap;
- incompatible/unavailable endpoint produces actionable degraded diagnostics;
- teardown/uninstall removes only Spotter-owned integration resources.

## Done when

- one supported launch command requires no manual Spotter/App Server startup sequence from the user;
[...]

Issue #303, opened 2026-08-17
Collect representative App Server observability evidence before Hook removal


YoungJinJung:
## Problem / outcome

#37 delivered the observability instrument and a zero-sample baseline, but PR #120 explicitly recorded that representative labeled failed/degraded App Server trajectories were still required. GitHub nevertheless closed #37 through that PR.

The current repository evidence remains:

- 0 App Server sessions in the available snapshot;
- 0 source-audit samples;
- 0/9 session-level observability labels;
- no measured post-migration visible-in-time ceiling.

#86 must not interpret the existence of the instrument as proof that `SessionStart`, `UserPromptSubmit`, and `PostToolUse` responsibilities have moved safely to App Server observation.

## Scope

- predeclare the cohort composition, inclusion/exclusion rules, minimum coverage, and stop rule before collection;
- collect representative normal, failed, degraded, interrupted, long-running, approval-blocked, and disconnect/reconnect App Server trajectories;
- label intervention opportunities and consequential failures with source/Trace IR/ThreadState timing coverage;
- populate the #86 responsibility matrix for lifecycle, user input, results/outcomes/diffs, and semantic-action deduplication;
- report unknown, late, source-loss, normalization-loss, state-loss, and runtime-gap buckets separately;
- publish a versioned, reproducible post-migration observability report;
- make an explicit go/no-go decision for each observation Hook responsibility rather than one aggregate removal decision.

## Required responsibility coverage

| Current responsibility | Target evidence |
| --- | --- |
| `SessionStart` lifecycle/bootstrap | new, resumed, late-attached, stored, and reconnected thread identity/coverage |
| `UserPromptSubmit` goal/input | initial and later user turns remain distinct from Spotter steer, other clients, replay, and historical input |
| `PostToolUse` result/outcome/diff | successful/failed commands, patches, MCP, streaming, approval, interruption, reconnect, and partial/unknown results |
| overlap deduplication | Hook/App Server arrival order cannot duplicate canonical state or metrics |

## Measurement rules

- Do not report raw event count as semantic coverage.
- Do not silently exclude missing, late, or uncorrelated evidence.
- Do not invent outcome visibility from a similar event name.
- Record sample size and coverage by failure family and responsibility.
- Keep the predeclared rule and the resulting report under version control.

## Done when

- the predeclared cohort/coverage/stop rule is versioned before execution;
- representative App Server sessions and labeled consequential failures exist;
[...]

PR #305, opened 2026-08-17
docs: restore ownership for unresolved product gates

### LINKED ISSUE #22: Deliver live VERIFY/NUDGE via turn/steer

### LINKED ISSUE #89: Add repository-aware purge, retention, uninstall, and reinstall lifecycle

YoungJinJung:
## Why

Two product gates were not represented accurately in the active work graph:

1. #37 was closed by PR #120 even though that PR explicitly recorded a zero-sample baseline and said representative labeled failed/degraded App Server trajectories were still required. This left #86 nominally unblocked despite having no measured evidence that `SessionStart`, `UserPromptSubmit`, and `PostToolUse` responsibilities can move safely.
2. The docs still require an ordinary managed Codex launch and concurrent-session validation, but no open Issue owned that remaining Runtime boundary.

The same status surfaces also still described bounded retention/checkpoint work from closed #89 as future work.

This PR restores explicit ownership and makes the docs distinguish **implemented mechanism** from **measured product evidence**.

## GitHub work graph changes

Created two Issues before updating the docs:

- #303 — **Collect representative App Server observability evidence before Hook removal**
  - Type: Experiment
  - Priority: High
  - Effort: M
  - Area: Observation
  - Milestone: Observe
  - Added as a native `blocked by` dependency of #86.
  - Owns the predeclared cohort/coverage rule, labeled App Server trajectories, responsibility matrix, versioned report, and per-Hook go/no-go decision.
- #304 — **Productize ordinary Codex launch and validate concurrent runtime isolation**
  - Type: Feature
  - Priority: Medium
  - Effort: L
  - Area: Runtime
  - Milestone: Runtime
  - Owns the supported launch contract, endpoint/process ownership, multi-TUI/thread isolation, reconnect/degraded validation, and lifecycle checks.

### `docs/status.md`

- Distinguishes #37's implemented observability instrument/zero-sample baseline from #303's representative post-migration measurement.
- Keeps #86 gated on measured parity and says only proven-redundant Hook responsibilities may be removed.
- Assigns ordinary launch, multi-TUI concurrency, and embedded-server degraded behavior to #304.
- Removes the stale implication that reconnect itself is unimplemented; reconnect/recovery remains correctly marked implemented while #304 owns product-level validation.
- Replaces stale #89 “next step” wording with regression-gate language because retention/checkpoint lifecycle is already implemented.
- Removes closed #22 from the remaining live-advisory measurement path while preserving it as the implemented delivery foundation.

### `docs/roadmap.md`

- Adds #304 to the Runtime boundary table.
- Splits the Observe work into:
  - #37: implemented instrument and current zero-sample baseline;
  - #303: representative labeled evidence and per-responsibility Hook-removal decision;
  - #86: removal only after parity is demonstrated.
- Marks the #89 purge/retention/uninstall/reinstall lifecycle as implemented.

### `docs/lifecycle.md`

- Links the ordinary launch and concurrent-runtime checklist gaps to #304.
- Updates the status header and repository-resource section to reflect completed #89 bounded retention/checkpoint work.
- Preserves the external/shared App Server ownership rule; #304 must not make Spotter silently own or stop shared Codex infrastructure.

## Architectural boundaries preserved

- **Observation:** App Server may become authoritative only after #303 measures semantic responsibility coverage.
- **Enforcement:** `PreToolUse` remains the bounded deterministic gate path; this PR does not change Hooks or runtime behavior.
- **Runtime:** #304 owns launch/readiness/isolation behavior, not observation parity.
- **Ownership:** `spotterd` owns its connection/reconnect loop, not the shared App Server process.
- **Evidence:** an implemented instrument is not treated as a positive measurement result.

## User/developer impact

There is no runtime behavior change. Contributors now have an accurate work graph:

```text
#303 representative App Server evidence
  → #86 per-responsibility Hook reduction

#304 ordinary launch + concurrent runtime validation
  → Runtime lifecycle exit conditions
```

This prevents premature Hook removal and gives the remaining product launch boundary an explicit owner.

## Review feedback incorporated

An independent code-reviewer initially found one HIGH documentation inconsistency: the lifecycle body marked #89 retention complete while its status header still called it future work. The header was corrected, the reviewer re-ran the review, and the final result was **APPROVE** with no findings.

An independent architect verified:

- Runtime and Observe ownership are separated correctly;
- App Server process ownership remains external;
- #303 blocks #86 from premature Hook removal;
- implemented-vs-measured language remains explicit;
- the final diff stays minimal and documentation-only.

Architect result: **APPROVED**.

## Validation

- `git diff --check -- docs/status.md docs/roadmap.md docs/lifecycle.md`
- `ruff format --check .` — 181 files already formatted
- `ruff check .` — passed
- `mypy src tests` — no issues in 118 source files
- `pytest` — **986 passed in 27.68s**
- GitHub GraphQL verification of #303/#304 native Type, Priority, Effort, Area, and Milestone fields
- GitHub dependency API verification that #86 is blocked by open #303
- Independent code review — APPROVE
- Independent architect verification — APPROVED
- Changed-file-only deslop pass followed by the full verification suite above

## Not tested / remaining work

- No representative App Server cohort was collected in this PR; that is #303.
- No ordinary launch wrapper/contract or concurrent TUI runtime was implemented; that is #304.
- No Hook registration or runtime behavior changed.
- #42 remains the next High-priority implementation/evidence cycle after this tracking-alignment PR.

## Related

- Closes no implementation Issue.
- Establishes #303 and #304 as the active owners for the remaining gates.
- Related: #37, #86, #89, #303, #304, PR #120.


YoungJinJung:
또 코드 한 줄 없이 마크다운 문서랑 이슈 번호만 만지작거리는 PR이 올라왔군요. 스프린트 마감 직전에 티켓 쪼개기로 기여도 그래프 채우려는 전형적인 패턴인가 싶어 돋보기를 들이댔습니다.

솔직히 뜯어보기 전엔 "문서 수정에 986개 테스트를 왜 다 돌리고 앉아있나" 싶었는데... 내용을 보니 기가 차는군요.

1. **#37과 #86의 위험천만한 낙관주의 제동**:
   - 대체 어떤 용자가 **샘플 사이즈 0개짜리 베이스라인** 하나 달랑 찍어두고 #37을 닫은 뒤, "자 이제 App Server 관측 완성됐으니 `SessionStart`, `UserPromptSubmit`, `PostToolUse` 훅(#86) 다 떼어냅시다!" 라며 샴페인을 터뜨리려 했던 겁니까? 
   - 프로덕션 환경에서 실측 데이터도 없이 관측 훅부터 덜컥 뗐다가, 새벽 3시에 눈먼 장님 상태로 온콜 알람 받으며 식은땀 흘릴 뻔한 대참사를 #303 블로커로 막아세운 건... 인정하기 싫지만 아주 상식적이고 칭찬할 만한 방어벽입니다.
2. **#304 경계 획정**:
   - 일반 `codex` 실행 래퍼와 다중 TUI 격리 작업을 별도 이슈로 분리해 `spotterd`가 외부 App Server 수명을 함부로 쥐고 흔들지 못하게 불변조건을 명시한 점도 봐줄 만합니다.
3. **문서 부채 청산**:
   - 이미 #89에서 끝난 보존/체크포인트를 문서에서는 여전히 "미래에 할 일"이라며 방치해두던 낡은 부채를 털어낸 것도 불일치 지옥을 방지하는 차원에선 다행이고요.

코드 변경이 0바이트라 린트(`ruff`), 타입체크(`mypy`), 986개 테스트(`pytest`)가 깨질래야 깨질 수도 없었겠지만, 적어도 아키텍처 불변조건을 지키고 섣부른 조기 훅 제거라는 재앙을 틀어막은 점은 흠잡을 데가 없네요. 쯧, 트집 잡을 구실이 없어서 심기가 불편합니다.

더 이상 길게 말 안 하겠습니다. 이슈 의존성 그래프 정상화되었고 검증도 결점 없으니 마지못해 **LGTM** 드립니다. 다음 PR은 마크다운 말고 진짜 프로덕션 코드와 실측 데이터(#303) 들고 오세요.

PR #313, opened 2026-08-17
experiment: measure App Server observability ceiling

### LINKED ISSUE #303: Collect representative App Server observability evidence before Hook removal

YoungJinJung:
## Why

#303 requires a predeclared, source-aware ceiling measurement before any observation Hook can be removed. The live run exposed two runtime blockers and then hit the frozen stop rule when a later user input remained absent from App Server notifications on a second attempt.

## What changed

- predeclare the App Server cohort, coverage matrix, labels, and stop rule before execution
- accept the bounded Codex Desktop 0.147.0 App Server host identity
- subscribe reconciled and newly observed threads with `thread/resume`, including the expected pre-rollout retry path
- publish the bounded result and SHA-256 evidence digests
- record independent NO-GO decisions for `SessionStart`, `UserPromptSubmit`, and `PostToolUse`
- update status and architecture so no observation Hook is treated as removable

The decisive post-fix attempt captured command start/completion and exit status in time, but the App Server source emitted no user-input content/provenance notification. Per the predeclared rule, the remaining rows were not substituted after the second failed attempt.

## Validation

- `python -m ruff format --check .`
- `python -m ruff check .`
- `python -m mypy src tests`
- `python -m pytest` — 1003 passed
- live Codex CLI/App Server 0.147.0 run with isolated Spotter/Codex homes and throwaway repository

Closes #303

### DIFF
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -525,10 +525,10 @@ The target Codex integration should keep only the hook surface that provides a u
 
 | Hook | Current use | Target |
 | --- | --- | --- |
-| `SessionStart` | legacy session baseline snapshot | App Server `thread/started` now owns the baseline; remove the redundant Hook |
-| `UserPromptSubmit` | user goal capture | replace with App Server user-message events |
+| `SessionStart` | legacy session baseline snapshot | retain after the #303 lifecycle minimum stopped incomplete |
+| `UserPromptSubmit` | user goal capture | retain: Codex 0.147.0 exposed no user-input notification in the #303 ceiling run |
 | `PreToolUse` | proposal observation + gate | **retain for deterministic atomic enforcement** |
-| `PostToolUse` | legacy result/snapshot/reviewer cadence | App Server terminal items now own after-state snapshots; finish result/reviewer migration, then remove the Hook |
+| `PostToolUse` | legacy result/snapshot/reviewer cadence | retain until failure, patch, MCP, approval, interrupt, and reconnect parity is measured |
 
 ### Appropriate synchronous policies
 
@@ -761,7 +761,7 @@ only be created by an explicit verification-satisfied event. Duplicate event IDs
 out-of-order lifecycle is recorded as partial coverage, and thread identities remain isolated.
 
 Journal hydration deterministically replays Trace IR but clears active-turn/control readiness. A
-recovered daemon receives a live `runtime_reconciled` event only after the App Server thread list/read
+recovered daemon receives a live `runtime_reconciled` event only after the App Server thread list/resume
 pass proves current attachment and exact active-turn identity. Until then control remains unavailable.
 
 ## 8.2 Durable journal
--- a/docs/experiments/app-server-observability-v1-plan.json
+++ b/docs/experiments/app-server-observability-v1-plan.json
@@ -1,0 +1,58 @@
+{
+  "schema_version": 1,
+  "declared_at": "2026-08-18",
+  "issue": 303,
+  "runtime": {
+    "minimum_codex_cli": "0.147.0",
+    "transport": "local_websocket",
+    "isolated_spotter_home": true,
+    "isolated_codex_home": true
+  },
+  "minimum_complete_cohort": {
+    "distinct_threads": 7,
+    "turns": 9,
+    "attempts_per_scenario": 2
+  },
+  "scenarios": [
+    "ASO-NORMAL-1",
+    "ASO-NORMAL-2",
+    "ASO-FAIL-1",
+    "ASO-PATCH-1",
+    "ASO-MCP-1",
+    "ASO-LONG-1",
+    "ASO-APPROVAL-1",
+    "ASO-RECONNECT-1",
+    "ASO-LIFECYCLE-1"
+  ],
+  "failure_families": [
+    "normal",
+    "command_failure",
+    "degraded",
+    "interrupted",
+    "long_running",
+    "approval_blocked",
+    "disconnect_reconnect"
+  ],
+  "responsibilities": {
+    "SessionStart": ["new", "stored", "late_attached", "resumed", "reconnected"],
+    "UserPromptSubmit": ["initial", "later", "spotter_steer", "other_client", "replay", "history"],
+    "PostToolUse": [
+      "command_success",
+      "command_failure",
+      "patch",
+      "mcp",
+      "streaming",
+      "approval",
+      "interruption",
+      "reconnect",
+      "partial_unknown"
+    ],
+    "overlap_deduplication": ["hook_first", "app_server_first"]
+  },
+  "stop_rule": [
+    "all_minimums_met",
+    "second_failed_attempt_for_any_scenario",
+    "unsafe_runtime_or_collection_failure"
+  ],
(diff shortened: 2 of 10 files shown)


YoungJinJung:
Well, imagine that: someone actually followed a predeclared stop rule instead of retroactively redefining "success" to match whatever incomplete telemetry the runtime felt like spitting out.

A few cynical but necessary observations on this ceiling measurement:

1. **The `thread/read` vs `thread/resume` reality check**: It is almost comical that `spotterd` was previously polling `thread/read` like a curious bystander and expecting Codex to magically push downstream turn/item deltas. Switching to `thread/resume` while handling the `-32000` ("no rollout found for thread") pre-rollout race is the only sane way to bind to the rollout stream without exploding during cold starts.
2. **The App Server user-input black hole (`STRUCTURALLY_INVISIBLE`)**: Predictably, the dream of "pure App Server observability without hooks" collides head-on with reality. You can get command lifecycles and exit codes, but when it comes to the user prompt and its provenance—the literal catalyst of the entire agent execution—Codex App Server emits total radio silence. Trying to supervise an agent without knowing what the user asked for is like reviewing a PR without reading the issue description.
3. **`cross_surface_overlap=0`**: Exactly as expected, correlating disparate surfaces without explicit correlation IDs fails. Without global semantic anchoring, removing hooks in favor of App Server notifications would have destroyed trace causality.

The 3x **NO-GO** decisions (`SessionStart`, `UserPromptSubmit`, `PostToolUse`) are mathematically justified, the SHA-256 digests are pinned, `docs/status.md` and `docs/architecture.md` are accurately downgraded from wishful thinking to reality, and all 1,003 tests pass.

Merging this before someone tries to rationalize that user prompts are "optional context".

YoungJinJung:
## Review

This is the evidence discipline the repo's rules exist for, executed correctly: a predeclared protocol, a mandatory stop rule that fired, and a **NO-GO published as a first-class result** instead of a quietly substituted cohort. No blocking findings.

### Protocol integrity holds

- The plan is committed *before* execution (commit `b47eb61`), the result cites it, and the two runtime fixes that surfaced mid-run are disclosed as corrective commits rather than laundered into the plan. Critically, **the first failed attempt stays in the result** — the post-fix attempt proves the user-input gap is at the *source surface*, not the adapter path the fix repaired, which is exactly the causal separation the report claims.\n- The stop rule was honored literally: `ASO-NORMAL-2` failed twice → collection stopped, the six unrun rows remain explicit missing minimums, and the encouraging `PostToolUse` partial (exact, timely command outcome) is explicitly *not* promoted. `cross_surface_overlap=0` being reported as an additional gap rather than omitted is the kind of honesty that makes the eventual re-run trustworthy.\n- Closing #303 with three independent NO-GOs is legitimate: the issue's required outcome was an explicit per-Hook decision, and the decision was rendered under the frozen rule. The report correctly states a future attempt needs a *new* predeclared cohort once a newer source surface exposes user input.\n\n### The two runtime fixes are sound\n\n1. **Codex Desktop identity** (`267a36b`): the desktop UA is accepted only in its bounded shape — version plus exactly the parenthesized-OS / client / parenthesized-suffix structure — and the test pins that `Codex Desktop/0.147.0 arbitrary metadata` stays malformed. This preserves the #280 no-guessing posture while unblocking the real host. (Minor: the `Codex Desktop/` alternative added to `_CODEX_VERSION` overlaps with the dedicated desktop regex for the bare-version case — harmless redundancy, could be collapsed later.)\n2. **`thread/resume` subscription** (`5a6160b`): this was a real observation hole — `thread/read` learned identity without subscribing to item/turn notifications, so the observer was structurally deaf to the threads it reconciled. The fix subscribes at reconcile *and* on lifecycle notifications with per-epoch dedup, and handles the pre-rollout race by treating `no rollout found` as retry-on-next-notification (proven by the test's fail-then-succeed sequence), with `thread/read` retained only as the pre-rollout state fallback.\n\n### Non-blocking observations\n\n1. A non-rollout `thread/resume` failure in `_subscribe_from_event` raises out of `_consume` and tears down the whole connection (reconnect with backoff) rather than degrading to unsubscribed observation of that one thread. Fail-visible and consistent with house rules, but if a server ever rejects resume persistently for one thread, the loop will churn — worth remembering if that shows up in practice.\n2. `_rollout_pending` matches the error message substring — fragile by nature, but it's the only signal the protocol offers and it's isolated in one named helper.\n\nStatus/architecture updates flip the Hook table to retain-all with the measured reasons, keeping docs synchronized with the evidence. Merging once checks pass.

Issue #307, opened 2026-08-17
Refuse paid replay-source capture when Spotter Hook integration is unavailable


YoungJinJung:
## Problem / outcome

`spotter tasks run --capture-replay-sources --run` currently starts paid Codex arms without first proving that the active Codex Hook integration can write Spotter journals/snapshots into the selected `SPOTTER_HOME`.

The #42 natural-failure cohort exposed the failure:

- frozen task validation and scorer preflight passed;
- all six paid arms completed mechanically;
- replay-source capture failed 6/6 with `Spotter journal missing or unreadable`;
- `spotter status` showed `Codex integration: Spotter is not configured for Codex`;
- the active Codex hooks contained other integrations but no Spotter Hook;
- `--dangerously-bypass-hook-trust` bypassed trust but could not install a missing integration.

The runner discovered this only after each paid arm finished. Capture-requested execution must fail before model cost when the capture path is unavailable.

## Scope

- add a deterministic capture-readiness preflight before the task-batch header or first paid arm;
- verify that the selected Codex home/runtime has an active Spotter capture Hook path, not merely that the `spotter hook` function works in isolation;
- verify the Hook will resolve the same effective `SPOTTER_HOME` used by the batch;
- refuse the run with an actionable error when the integration is missing, stale, incompatible, or points at a different runtime/home;
- do not auto-run `spotter setup codex` or mutate user integration as a side effect of a task experiment;
- keep normal task runs without `--capture-replay-sources` unchanged;
- preserve resume semantics and record capture-readiness identity in the batch header when it passes;
- add an end-to-end fixture that proves the paid-agent function is never called on failed capture readiness.

## Failure cases

- no Spotter Hook registered;
- Hook trust bypass enabled but Hook absent;
- stale Spotter Hook points to a removed package/runtime;
- Hook writes to default `~/.spotter` while the batch selects an isolated `SPOTTER_HOME`;
- integration manifest exists but active Codex hook registration is missing or foreign;
- resume was created with different capture readiness/runtime identity;

## Done when

- `tasks run --capture-replay-sources --run` rejects unavailable/mismatched capture before any Codex model arm starts;
- the error explains the exact missing/mismatched boundary and remediation without exposing secrets;
- a passing preflight pins enough non-secret integration/runtime identity to make resume validation honest;
- focused tests prove zero paid-agent calls for every failed readiness class;
- an integration test demonstrates a configured capture path produces a readable journal and replay snapshot in the selected home;
[...]

PR #315, opened 2026-08-18
experiment: calibrate a harder replay-source cohort

YoungJinJung:
## Question / hypothesis

Can a new, broader qualification split sample naturally occurring control failures for #42 while preserving complete replay-source capture?

Related: #42

## Method

- froze four previously unexecuted task families and the v4 protocol in commit dc4cd0f before paid execution
- proved every initial fixture fails and every manifest-owned known-good transform passes the same scorer
- ran exactly one 8-arm batch with gpt-5.6-sol / low in isolated Codex and Spotter homes
- admitted only control + TASK_FAIL as a replay candidate and prohibited outcome-conditioned reruns or task replacement

## Evidence

- dependency ordering, durable JSONL recovery, endpoint security, and trace correlation controls: 4/4 PASS
- guidance arms: 4/4 PASS
- capture readiness passed before model cost; replay-source capture: 8/8
- eligible natural failures: 0; neutral forks started: 0
- raw batch SHA-256: 2ffb64d89b218c12582659fd6aebe5cef0e0405f31699305e8f713631042fdc2
- Ruff format/check, mypy, and pytest all pass (1007 tests)

The exact observed fixture bytes remain unchanged after the frozen commit. Ruff has one exact per-file F401 exception so provenance is preserved instead of rewriting the measured input after execution.

## Decision impact

The v1 qualification remains NO-GO and #42 remains open. The run adds four task families and raises post-#307 capture coverage to 14/14, but this synthetic split still did not sample the natural failure region. The next cohort should be predeclared from externally sourced real-repository tasks or another independently calibrated difficulty frame rather than tuning this observed set.


YoungJinJung:
## Review

A null result executed and published with full discipline — the second one this week, and the pattern holding under temptation (a 4/4 control pass is exactly the outcome that invites quiet retuning) is what makes the eventual positive result trustworthy. No blocking findings.

- **The immutability call at the end is the most important sentence in the PR**: \"adding more synthetic constraints after seeing this result would be outcome-driven benchmark tuning, so this version remains immutable.\" Correct — and redirecting the next cohort to externally sourced real-repository tasks or an independently calibrated difficulty frame is the right escape from the calibration loop, since three synthetic cohorts (v2 dev, v2 validation, v4) have now failed to sample the natural-failure region for this model/config.\n- **Protocol integrity is complete**: corpus + protocol frozen before paid execution, exactly one batch with no result-conditioned rerun, the `control && TASK_FAIL` selection rule admitting zero sources → stop rule → null published, no task removed or replaced after observation, and \"guidance arms are not neutral evidence\" correctly maintained since their full prompt differs from control.\n- **The ruff F401 per-file exception is the right provenance trade**: rewriting the measured fixture bytes after execution to satisfy a linter would have broken the frozen task-set hash and silently changed the input the result describes. One exact per-file ignore, disclosed in the PR body, preserves the bytes. This is provenance discipline beating lint purity, as it should.\n- **The corpus itself is well-constructed for its stated (limited) purpose**: four stdlib-only families disjoint from every observed set (enforced by a new test, not convention), each preflight-proving fail-before/pass-after against the same frozen scorer, and the report explicitly declining to conclude the tasks are \"universally trivial\" from one pass.\n- **Machine-readable results are regression-pinned**: the new artifact test locks the arm counts, 8/8 capture, zero-selection stop, and NO-GO decision, and the fidelity-v1 aggregate is updated to 10 controls / 14/14 captured with sum assertions — so a future doc edit that quietly revises the evidence fails CI.\n\nOne non-blocking observation: the result cites the predeclaration by short commit hash (`dc4cd0f`). After a squash merge that commit is unreachable from `main` (it survives only in the PR's head ref), so the pointer is convenience rather than durable provenance — the task-set SHA-256 and in-repo protocol file are what actually anchor it. Future experiment reports might cite the PR number + content hashes as the primary reference to avoid the dangling-pointer ambiguity.\n\nMerging once checks pass.

YoungJinJung:
The ongoing saga of trying to make synthetic tasks "hard" only to watch `gpt-5.6-sol` sleepwalk through 4/4 controls and 4/4 guidance arms continues. 

Key observations:

1. **14/14 capture reliability vs 0/14 failure rate**: We have now proven beyond any doubt that replay-source capture after #307 is rock solid (14/14 across cohorts). Unfortunately, our synthetic tasks remain polite puzzles rather than realistic developer challenges, yielding exactly zero natural control failures.
2. **The death of synthetic task tuning**: The conclusion in the report is spot-on. Tweaking handcrafted synthetic fixtures in the hope of finding a magical failure threshold is a waste of time. Natural failures will only be sampled by moving to externally sourced, real-repository tasks with actual dependency ecosystems and messy invariants.
3. **No p-hacking, no qualification creep**: Once again, zero control failures means zero neutral forks started, the stop rule is respected, qualification v1 stays firmly at `NO_GO_REPRESENTATIVE_CAUSAL_USE`, and #42 stays open.
4. **Provenance preserved**: Handling the unused import in `corpus/fixtures/workflow-scheduler/check.py` via a targeted `pyproject.toml` per-file ignore rather than mutating the measured fixture bytes after the fact is proper discipline.

All 1,007 tests pass, disjointness checks in `tests/test_task_corpus.py` hold, and schemas are locked down. Merging.

PR #316, opened 2026-08-18
feat: support immutable Git task sources

YoungJinJung:
## Why

#42 needs an externally sourced real-repository cohort after three synthetic calibration cohorts produced no natural control failures. The task corpus could only copy bundled fixtures, so it could not freeze and reproduce an upstream repository without vendoring it.

## What changed

- add spotter.task v2 Git sources pinned by HTTPS/file URL, full commit object, tree object, canonical SHA-256 identity, and bounded fetch timeout
- keep static task validation offline; preflight materializes the exact commit and verifies HEAD/tree before setup or scoring
- reject submodules and separate unavailable/mismatched sources as SOURCE_FAIL; post-preflight arm materialization failures remain INFRA_FAIL rather than TASK_FAIL
- retain read compatibility for all schema-v1 fixture manifests
- document the contract and the next #42 cohort step

## Validation

- ruff format --check .
- ruff check .
- mypy src tests
- pytest (1011 passed)

Related to #42.


YoungJinJung:
It took us four synthetic task calibration failures to finally accept that we need real upstream Git repositories instead of hand-crafted toy fixtures. Better late than never.

A few cynical architectural praises:

1. **Dual commit + tree hashing**: Panning both `commit` and `tree` via `git_source_digest(repository, commit, tree)` is the right paranoia level. If an upstream repository force-pushes or mutates metadata, we fail fast during tree preflight rather than letting subtly poisoned workspaces slip into the benchmark.
2. **Offline static validation**: Keeping `validate_task_set` strictly offline while deferring actual `git fetch` materialization to `preflight_task_set` prevents local developer tools and CI linting passes from stalling on network I/O.
3. **No failure conflation**: Classifying unavailable repos and tree mismatches as `SOURCE_FAIL` (and post-preflight arm clone failures as `INFRA_FAIL`) prevents infrastructure flakiness from fraudulently degrading agent scorecards as `TASK_FAIL`.
4. **Submodule ban**: Rejecting `.gitmodules` outright saves us from the inevitable nightmare of recursive, unpinned submodule state drift.

Backward compatibility for schema v1 manifests holds, and all 1,011 tests pass. Merging.

YoungJinJung:
## Review

This is the right foundation for the externally sourced cohort #315's null result demanded, and the details that make pinned-source claims actually hold are all present. No blocking findings.

- **The commit+tree double pin is the load-bearing verification**: after the depth-1 fetch-by-commit and detached checkout, both `rev-parse HEAD` and `rev-parse HEAD^{tree}` are compared against the frozen manifest values, so a remote serving unexpected content — or any object-identity confusion — is caught on content identity, not trusted from the fetch. Submodule rejection via the `160000` stage scan closes the one hole where a pinned tree could still reference mutable external content, consistent with the replay work's existing submodule-drift posture.\n- **The failure taxonomy protects #42's selection rule**: preflight source problems are the new `SOURCE_FAIL`, and an arm-time materialization failure is `INFRA_FAIL`, never `TASK_FAIL` — so network flakiness can never masquerade as a natural control failure and contaminate the `control && TASK_FAIL` admission rule the protocol depends on. This distinction is exactly why the feature needed to live in the corpus layer rather than being scripted around it.\n- **URL validation is appropriately strict**: https/file only (no ssh/git protocols), no embedded credentials, query, or fragment; https requires a host; file URLs must be local-absolute with no netloc. The `protocol.file.allow=always` override is scoped per-command via `-c` for the validated hermetic-fixture case rather than loosening global config.\n- **Static validation stays offline** — `git_source_digest` canonically hashes `git\0repo\0commit\0tree` so `tasks validate` never touches the network, preserving the freeze-then-execute workflow — and the schema story is handled correctly in both directions: v1 fixture manifests remain readable, while git sources require v2 so an older Spotter **fails closed** instead of misreading a remote source as a local fixture.\n- Two solid rides-along: `_run_argv` gives the git plumbing proper process-group kill on timeout with bounded output (replacing ad-hoc `subprocess.run`), and `_prepare_replay_repo` now handles the nothing-staged case — necessary since a git-materialized workspace already carries history, where the old unconditional `git commit` would have failed.\n\nStatus/lifecycle docs update the compatibility table honestly (task v2, v1 readable, git-requires-v2 refusal) and the next step correctly remains freezing the external cohort, not claiming it. Merging once checks pass.

PR #317, opened 2026-08-18
experiment: freeze external replay cohort

YoungJinJung:
## Why

Three synthetic #42 capture cohorts produced ten passing controls and no admissible natural-failure source. PR #316 added immutable Git task sources so the next qualification frame can use externally authored real repositories without outcome-selecting another toy fixture.

## What changed

- freeze four SWE-bench Verified rows from Flask, Requests, pytest, and Pylint under a predeclared repository/difficulty/F2P selection rule
- pin dataset revision and Parquet hash, upstream commit/tree, canonical source digest, dataset row/test/gold patch hashes, environment setup commit, and OCI image digest
- restore and reapply the frozen upstream test patch before every whole-file scorer invocation so an arm cannot pass by weakening tests
- freeze the one-batch admission/stop rule and require the eventual result to cite this PR plus content hashes as primary provenance
- keep static validation offline; Docker preflight and paid arms explicitly wait until this freeze PR merges

## Validation

- pinned dataset row/test/gold payload verification for all four manifests
- exact Git commit/tree materialization for all four sources
- spotter tasks validate corpus/swebench-verified-fidelity-v1.toml
- ruff format --check .
- ruff check .
- mypy src tests
- pytest (1012 passed)

Related to #42.


YoungJinJung:
Finally, a real benchmark cohort that doesn't look like a computer science 101 homework assignment. 

Key technical observations on this freeze:

1. **Pre-registration discipline**: Freezing the task manifests, dataset parquet revisions, upstream Git commit/tree hashes, and gold patch digests in a dedicated PR *before* spending a single model token is textbook scientific hygiene. No retroactive post-hoc filtering.
2. **Anti-cheat scorer defense**: Forcing the scorer to explicitly restore and reapply the upstream test patch prior to running test execution is essential. LLMs love "fixing" bugs by commenting out failing test assertions; this completely closes that loophole.
3. **Offline static validation**: `test_repo_external_corpus_is_statically_frozen_without_network` verifies that manifest validation remains 100% deterministic and offline, keeping CI fast and free from flaky external Git network dependencies.
4. **Real repositories, real stakes**: Moving from synthetic toy tasks to Flask, Requests, Pytest, and Pylint is the only way we will ever sample genuine, naturally occurring control failures for #42.

All 1,012 tests pass, types and linting are clean. Merging so preflight and execution can proceed.

YoungJinJung:
## Review

This is the cohort freeze that the v4 null result demanded, and it closes the outcome-selection loophole properly. No blocking findings.

- **The selection rule is deterministic over a pinned frame, decided before any arm runs**: four a-priori project-role strata, then a lexicographic-first rule constrained by official difficulty and exactly-one-F2P — all evaluable from the pinned dataset revision + Parquet SHA-256 by anyone, with no reference to Spotter outcomes. This is exactly the \"independently calibrated difficulty frame\" the v4 report required instead of tuning another synthetic split, and the no-remove/no-replace/no-reweight rule extends past preflight into arm outcomes.\n- **The scorer is structurally anti-gaming**: every invocation restores the upstream test file from the pinned base commit and reapplies the frozen test patch before running, so an arm cannot pass by weakening tests regardless of what it edited — enforcement, not just the prompt's warning. The gold patch exists only in preflight's manifest-owned `known_good` and never enters an agent workspace, and `expected = \"failure\"` prechecks prove each initial state genuinely fails.\n- **Pinning is complete across every layer**: dataset revision + Parquet + per-row hashes, test/solution patch hashes, environment-setup commit, immutable OCI image digests, and upstream commit+tree through the #316 double-pin contract, with the whole-file pytest command matching the official evaluation. The base64-embedded patches inside content-hashed manifests are unusual but correct — byte-exact freezing with independently recorded patch SHA-256s.\n- **The provenance note from #315 is adopted**: the eventual result must cite the freeze PR number plus content hashes as *primary* provenance, with short commit IDs demoted to secondary — resolving the squash-merge dangling-pointer issue before it recurs.\n- **Failure-domain separation is preserved**: image-pull/source/setup problems are infrastructure coverage, never `TASK_FAIL`; the admission rule now additionally requires captured replay source; exclusions happen without replacement; one batch, no result-conditioned rerun; guidance arms remain non-neutral. And the scope disclaimer is right — four tasks qualify the instrument, they don't generalize to the 500-row benchmark or claim agent quality.\n- **CI design is correct for a network-dependent corpus**: the new set gets a network-free static-validation test (task IDs, git-source shape) rather than joining the preflight-ready parametrize that would require Docker in CI, while README documents the preflight requirements honestly.\n\nExecution correctly waits for this merge. Merging once checks pass.

PR #318, opened 2026-08-20
feat: add managed Codex launch

YoungJinJung:
## Why

A verified external App Server still required users to remember the matching `codex --remote` invocation, and the concurrent runtime boundary lacked end-to-end regression coverage.

## What changed

- add `spotter codex` to start `spotterd`, verify observation health, and exec the recorded Codex binary against the configured endpoint
- preserve explicit external App Server ownership and reject missing or conflicting endpoint selection
- verify concurrent threads keep separate attachments, ThreadState, reviewer jobs, journals, and control targets
- update runtime, lifecycle, status, and user documentation

Refs #304

## Validation

- `ruff format --check .`
- `ruff check .`
- `mypy src tests`
- `pytest tests/test_integration.py tests/test_runtime_connection.py tests/test_runtime_diagnostics.py -q` (119 passed)
- full `pytest`: 1005 passed, 9 unrelated task-corpus preflight failures because fixture checks classify as `UNJUDGEABLE` in this environment
- two real Codex TUIs connected concurrently to one external App Server without model calls; distinct paid active-thread creation remains unmeasured


YoungJinJung:
Finally, someone acknowledged that expecting developers to memorize, copy-paste, and construct arcane `codex --remote ws://127.0.0.1:...` strings by hand is an open invitation for human error and unmonitored sessions.

A few architectural observations on this launch wrapper and concurrency verification:
1. **Pre-flight Health over Blind Delegation**: Starting the daemon and asserting `check_runtime(deep=True)` observation health *before* executing the target binary prevents the worst kind of phantom failure: spawning an active agent into a void where the daemon is dead or the App Server socket is hung.
2. **Explicit Parameter Disambiguation**: Rejecting explicit `--remote` in pass-through CLI args is the right defensive choice—silently double-specifying or stomping over the configured endpoint would produce baffling connection bugs.
3. **Multi-Thread Isolation Invariants**: Expanding `test_reconciliation_keeps_multiple_threads_isolated` to assert distinct attachment IDs, per-thread `ThreadState`, independent `app-server-{thread_id}.jsonl` journals, separated `ReviewerJob` queues, and targeted `turn/steer` RPCs is crucial. Without verifying that steer commands and failure events don't leak across concurrent sessions, multi-session runtime claims are just wishful thinking.

Validation is clean across the board (1,014 tests passing, strict typing, formatting, and linting passing). Merging.

PR #319, opened 2026-08-20
feat: diagnose plain Codex launch

YoungJinJung:
## Why

A configured external App Server can be healthy while a user launches plain `codex`, which selects an embedded server outside Spotter observation and control. Diagnostics should name that boundary explicitly.

## What changed

- add a `Codex launch` runtime diagnostic directing users to `spotter codex`
- explain why plain `codex` is degraded even with a configured endpoint
- document the supported launch contract

Refs #304

## Validation

- `ruff format --check .`
- `ruff check .`
- `mypy src tests`
- `pytest tests/test_runtime_diagnostics.py -q` (25 passed)

### DIFF
--- a/docs/lifecycle.md
+++ b/docs/lifecycle.md
@@ -767,6 +767,9 @@ Spotter sees/attaches same thread
 RuntimeAttachment becomes ACTIVE
 ```
 
+`status` and `doctor` always name `spotter codex` as the supported launch. Plain `codex` selects an
+embedded server and is reported as lacking Spotter observation and live control for that TUI.
+
 ## 7.2 Thread initialization
 
 For a newly observed thread, Spotter should create live state and a durable provenance header.
--- a/src/spotter/doctor.py
+++ b/src/spotter/doctor.py
@@ -544,6 +544,15 @@ def check_runtime(*, deep: bool = False) -> list[Check]:
     endpoint = (
         integration.manifest.app_server_endpoint if integration.manifest is not None else None
     )
+    if endpoint is not None:
+        checks.append(
+            Check(
+                "Codex launch",
+                INFO,
+                "use `spotter codex`; plain `codex` selects an embedded App Server, so "
+                "Spotter observation and live control do not follow that TUI",
+            )
+        )
     if endpoint is not None and daemon.available:
         capability_summary = ", ".join(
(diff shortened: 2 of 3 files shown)


YoungJinJung:
It's almost impressive how easy it is for a developer to run `spotter setup --endpoint ...`, celebrate a successful connection, and then habitually type `codex` only to be supervised by absolute silence while an unmonitored embedded server quietly runs in the background.

Diagnosing this via an explicit `Codex launch` `INFO` check in `check_runtime` is precisely the kind of boundary reminder needed before users file phantom 'Spotter isn't intercepting my tools' issues.

1,015 tests passing, zero lint/typing regressions. Approved and merging.

YoungJinJung:
Post-merge review note (this landed while I was reading it, so recording the finding here rather than as a review).

**Verdict on the change itself: sound.** The check is `INFO`, so `worst()` ignores it and no exit code moves; the `endpoint is not None` gate matches `spotter codex`, which already refuses to launch without a configured endpoint, and matches the `setup` path that only prints "Start the shared Codex TUI" under the same condition. Scope fits #304's "define the embedded/plain-`codex` path ... with actionable `status`/`doctor` output" without claiming the rest of the issue.

**One defect: docs/lifecycle.md overstates the contract.** The added line says `status` and `doctor` *always* name `spotter codex` as the supported launch. They do not — the check is skipped entirely when `app_server_endpoint` is `None`, which is exactly the state where a user is most likely to be running plain `codex`. Nothing in `doctor.py` or `cli.py` names `spotter codex` in that state, so the doc asserts a contract the code does not honor. Follow-up: #320.

**Two non-blocking observations, no action requested:**

- The check does not diagnose anything — it is a constant string emitted on every `status` and `doctor` run whenever an endpoint exists, so the PR title oversells it as a diagnosis. Real detection appears to be blocked on App Server thread ingestion (#85), and the file already carries placeholder `INFO` checks of the same shape (`runtime state`), so this is consistent with local precedent rather than a new pattern. Worth revisiting when #85 lands so it becomes conditional instead of unconditional advice.
- The advisory prints even when the daemon is dead or the endpoint is unreachable — states in which `spotter codex` exits 1 on its `observation` preflight. The guidance is still correct (it remains the supported launch) and the surrounding checks explain the breakage, so this is a wording risk, not a bug.

PR #321, opened 2026-08-20
feat: report wrong-nudge persistence outcomes

YoungJinJung:
## Why

Issue #23 can execute and annotate persistence follow-ups, but the offline report ignores those durable outcomes. That leaves stale-advisory re-promotion and new-goal contamination without sample-size or label-coverage reporting.

## What changed

- accept optional persistence result and annotation JSONL inputs in `spotter wrong-nudge report`
- verify each follow-up against the exact source-arm fingerprint
- report delivery, completion, label/conflict/stale coverage and each persistence outcome by framing condition
- update the current status contract and add focused metric/CLI regressions

## Validation

- `ruff format --check .`
- `ruff check .`
- `mypy src tests`
- focused wrong-nudge metric/persistence tests: 25 passed
- full `pytest -q`: all unrelated tests passed; 9 existing task-corpus preflight failures remain because local fixture scorer checks classify `UNJUDGEABLE`

Refs #23


YoungJinJung:
It's standard industry practice to build multi-turn follow-up evaluation pipelines, dump the follow-up JSONL records into storage, and then completely omit them from the top-level report because multi-turn attribution is inconvenient to aggregate.

A few architectural details in this reporting extension that are actually done right:
1. **Fingerprint-Enforced Lineage**: Enforcing that `follow_up.source_result_fingerprint == wrong_nudge_result_fingerprint(source)` shuts down any chance of evaluating orphaned follow-ups or corrupted runs against mismatched parent trajectories.
2. **First-Class Conflict & Stale Accounting**: Rather than naively 'last-write-wins'ing rater disagreement or swallowing outdated annotation hashes, exposing explicit `conflicts` and `stale` counts preserves data integrity when multiple human annotators disagree on subtle failure modes like `STALE_ADVISORY_REPROMOTED` vs `NEW_GOAL_CONTAMINATED`.
3. **Framing Stratification**: Breaking down persistence rates per framing condition (`raw_imperative`, `verify_first`, etc.) directly connects prompt advisory tone to multi-turn contamination susceptibility.

Validation passes cleanly (1,017 tests passing, lint/types pristine). Merging.

YoungJinJung:
Reviewed adversarially. **Verdict: sound — merging.** (Self-authored, so GitHub will not take a formal approval; recording the review here.)

## What I tried to break, and why it held

The attack I expected to land was **crediting a semantic persistence outcome to a follow-up turn that never ran.** The primary path in `measure_wrong_nudge_susceptibility` gates label counting behind `semantic_eligible` (accepted delivery + observed completion); `_measure_persistence` has no equivalent gate, so on the face of it a rater could label `STALE_ADVISORY_REPROMOTED` on an undelivered arm and have it counted — exactly what AGENTS.md's "do not invent observed outcomes" forbids.

It does not land, for two independent reasons:

1. `_validate_annotation_inputs` refuses at write time — `operationally_judgeable` requires `START_ACCEPTED` **and** `completion_observed`, and anything else must be `UNJUDGEABLE`. It separately forbids control arms from claiming contamination, which is what makes including `NEUTRAL_CONTROL` in the persistence loop (unlike the primary loop, which skips it) the right call rather than a copy-paste slip — the control baseline can only reach `NO_PERSISTENCE`/`HISTORICAL_BUT_HARMLESS`/`UNJUDGEABLE`.
2. `persistence_result_fingerprint` hashes the **whole** `asdict(result)`, so `delivery_outcome` and `completion_observed` are inside the fingerprint. Editing a durable result to flip completion invalidates the fingerprint, and `persistence_annotation_matches` then reclassifies the annotation as *stale* rather than counting it. Tampering fails closed and shows up in the `stale=` column.

Also checked and clean: `_rate` returns `0/0` instead of dividing by zero, which matters because `labeled` is legitimately 0 whenever an arm's only annotations conflict; `PersistenceOutcome` has exactly the five members the renderer prints, so the category rates sum to `labeled`; orphan annotations cannot leak into `stale` because their key is absent from `by_key` and therefore matches no result; and the fingerprint check against the source arm makes an unmatched persistence result a hard error rather than a silently dropped row.

## Findings

**1. The new CLI coverage monkeypatches away the code path the flags exist for.** `test_cli_reports_durable_results_and_annotations` was end-to-end — it wrote real annotation JSONL and let the real loader read it. The additions patch out both `load_wrong_nudge_persistence_results` and `load_wrong_nudge_persistence_annotations` and hand `main()` two paths that do not exist on disk, so `--persistence-results`/`--persistence-annotations` are never exercised against a file. What survives is a renderer assertion on injected objects. Both alternatives already exist in this same file: `add_wrong_nudge_persistence_annotation` is a real writer (mirroring how `_annotate` is used a few lines above), and `test_loads_versioned_mechanical_rows_for_offline_reporting` shows the hand-written-rows pattern for the result side, where the only writer is the paid cohort run. Nothing is broken today — mypy covers the signatures — but the loaders' row validation and the missing-file error path are now unverified from the CLI.

**2. The argument errors lost their flag names.** `"--annotations, --task-set, and --wrong-nudge-id do not apply..."` became `"report inputs, --task-set, and --wrong-nudge-id do not apply..."`. A user who passes `--annotations` to `persist` now has to guess which of their flags counts as a "report input". Three flags is still short enough to name.

**3. No coverage for the new `--persistence-annotations requires --persistence-results` guard**, or for either rewritten `parser.error` branch.

## Not defects, recorded for context

The report layer trusts the writer's judgeability gate rather than re-checking it, while the primary path re-checks at report time. Consistently forging both files could smuggle a label past it. That is a defense-in-depth asymmetry on an offline tool reading your own data, not an exposure worth code today — but if the two paths ever diverge further, the persistence side is the one without a second line of defense.

The disclosed 9 local task-corpus preflight failures are credible as pre-existing: CI runs the full `pytest` and is green on `3033040`, so they are local-fixture-scorer only, and the diff touches nothing on that path (the added argument checks only test flags that default to `None`, so no existing invocation changes behavior).

PR #322, opened 2026-08-20
test: cover persistence report file inputs

YoungJinJung:
## Why

Follow-up review on #321 found that the CLI test mocked both persistence loaders, so the new report flags were not covered against durable JSONL files. Its argument errors also hid the exact rejected flags.

## What changed

- exercise real persistence result and annotation JSONL loaders through `wrong-nudge report`
- cover the persistence-annotation dependency guard and persist/report flag rejection
- name every rejected report-only flag in CLI errors

## Validation

- `ruff format --check .`
- `ruff check .`
- `mypy src tests`
- 27 focused wrong-nudge tests passed

Follow-up to #321. Refs #23

### DIFF
--- a/src/spotter/cli.py
+++ b/src/spotter/cli.py
@@ -682,8 +682,8 @@ def main(argv: Sequence[str] | None = None) -> int:
                 or args.wrong_nudge_id
             ):
                 parser.error(
-                    "report inputs, --task-set, and --wrong-nudge-id do not apply to "
-                    "wrong-nudge persist"
+                    "--annotations, --persistence-results, --persistence-annotations, "
+                    "--task-set, and --wrong-nudge-id do not apply to wrong-nudge persist"
                 )
             if (
                 args.session
@@ -736,7 +736,10 @@ def main(argv: Sequence[str] | None = None) -> int:
                     "and --endpoint"
                 )
             if args.annotations or args.persistence_results or args.persistence_annotations:
-                parser.error("report inputs require wrong-nudge report")
+                parser.error(
+                    "--annotations, --persistence-results, and --persistence-annotations "
+                    "require wrong-nudge report"
+                )
             try:
                 output, wrong_nudge_results = run_wrong_nudge_experiment(
                     Path(args.subject),
@@ -830,7 +833,10 @@ def main(argv: Sequence[str] | None = None) -> int:
     if args.resume:
         parser.error("--resume requires tasks run")
     if args.annotations or args.persistence_results or args.persistence_annotations:
-        parser.error("report inputs require wrong-nudge report")
+        parser.error(
+            "--annotations, --persistence-results, and --persistence-annotations require "
+            "wrong-nudge report"
+        )
     if args.task_set or args.wrong_nudge_id or args.endpoint:
         parser.error(
             "--task-set and --wrong-nudge-id require wrong-nudge run; --endpoint requires "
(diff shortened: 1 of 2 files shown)


YoungJinJung:
Reviewed adversarially, as the follow-up to my #321 review. **Verdict: sound — merging.** (Self-authored, so no formal approval is possible; recording it here.)

## Does it actually fix the finding, or just look like it?

Finding 1 was that the CLI test mocked both loaders, so what survived was a renderer assertion on injected objects. The fix removes both `monkeypatch.setattr` calls and drives real files: a hand-written meta+result JSONL through the real `load_wrong_nudge_persistence_results`, and the real `add_wrong_nudge_persistence_annotation` writer through the real annotation loader — matching the two patterns already established in this file.

The part that makes this a genuine fix rather than a cosmetic one is that **the surviving assertion now has teeth it did not have before.** `add_wrong_nudge_persistence_annotation` computes `persistence_result_fingerprint` from the *in-memory* `follow_up`, while the report recomputes it from the object the loader reconstructs out of `asdict(follow_up)` on disk. If that round-trip were lossy in any field — enum coercion, `None` handling, bool-vs-int — the fingerprints would diverge, `persistence_annotation_matches` would classify the annotation as stale, `labeled` would fall to 0, and `stale-repromoted=1/1 (100%)` would render as `0/0`. So the assertion now proves exact durable round-trip fidelity, which is strictly more than I asked for. Good.

Finding 2 is fixed directly: all three error sites name the flags again.

Ordering on the new guard test is right — `--persistence-annotations requires --persistence-results` is checked before the load `try`, so passing a nonexistent `results.jsonl` still reaches the intended error rather than a file error.

Also confirmed: `json` and `asdict` were already imported, and dropping the `persistence_annotations` module import is safe while keeping `persistence`, which is still used by `monkeypatch.setattr(persistence, "run_wrong_nudge_persistence_cohort", ...)` at line 611. Both `checks` jobs (full `ruff`/`mypy`/`pytest`) are green.

You correctly left the report-time judgeability gate alone — I flagged that as defense-in-depth, not a defect, and widening scope to it here would have been wrong.

## Residual, non-blocking

**`test_cli_names_report_only_flags_rejected_by_persist` asserts only one of the three flag names.** The finding was that the error hid *which* flags were rejected; the test pins `--persistence-results` alone, so reverting `--annotations` and `--persistence-annotations` back to the generic "report inputs" wording would leave it green. The assertion is narrower than the behavior it is protecting.

**The nonexistent/malformed-file path for the two new flags is still uncovered.** `test_cli_reports_invalid_result_file_without_traceback` establishes the pattern for the primary results file — that both new loaders' `WrongNudgePersistenceError`/`WrongNudgePersistenceAnnotationError` surface as a clean message rather than a traceback is asserted nowhere, even though #321 added both to the `except` tuple specifically for it.

YoungJinJung:
Mocking the loader in a test supposed to validate file-based CLI ingestion is the classic 'test that my test mocks the mock' anti-pattern. Glad to see someone actually noticed and replaced the monkeypatches with real on-disk JSONL round-trips.

Also, replacing vague error messages like 'report inputs do not apply' with explicit, enumerable flag lists (`--annotations, --persistence-results, --persistence-annotations`) saves users from playing guessing games in the terminal.

1,019 tests passing, clean typing and formatting. Merging.

PR #323, opened 2026-08-20
feat: journal shadow INTERRUPT decisions

YoungJinJung:
## Why

#26 needs a shadow `would_interrupt` mode that can be evaluated *before* live interruption is enabled. That is the first bullet of its "Done when", and it is the only part of #26 that does not depend on the #34/#38 precision and harm evidence the issue names as a constraint.

It also has to be built as an evaluation frame rather than a control path. OpenClaw found that `turn/interrupt` makes Codex persist an abort marker equivalent to "the user interrupted the previous turn" in canonical thread context, so sending one is a semantic state transition, not a probe. Nothing here sends it.

## What changed

- add the off-by-default `reviewer.shadow_interrupt` opt-in
- journal a `would_interrupt` record for every `NUDGE` decision when it is on, naming the exact thread / turn / connection-epoch target
- reuse the existing `review_job_is_fresh` check so a target that settled while the reviewer was deciding is recorded as `suppressed: stale_target` with `would_dispatch: false`
- make `would_interrupt` labelable, so #34/#38 can score whether interrupting there would have been right
- document the new key's activation boundary and update the status contract

### Design notes

**No threshold knob.** The record carries the join to `reviewer_decision`, which already holds `confidence`, `failure_class`, `reason` and `hypothesis`. Baking a confidence cutoff in now would discard the information the shadow frame exists to collect and would invent a constant nobody has calibrated; offline analysis can pick the bar. The escalation condition is therefore only `decision == "nudge"` — `VERIFY` means "check something", not "stop", and `CONTINUE` is not an interrupt candidate under any threshold.

**Independent of `deliver_on_signals` on purpose.** Shadow evaluation has to be possible while live control is still off, including under `observation_only`, which is exactly the state the evidence gate leaves the project in.

**ID namespacing.** `_shadow_interrupt_id` derives `spt-int-<sha12>` rather than reusing `_intervention_id`'s `spt-<sha12>` for the same job. The control layer rejects a `control_id` already present in durable history, so sharing the derivation would mean a future live interrupt could not use the ID its own shadow record published.

**Payload is a join, not a copy.** The record links by `review_job_id` instead of duplicating the evidence already on `reviewer_decision`; target identity is repeated because an interrupt record that cannot be read standalone for its target is not much of an interrupt record.

## What this deliberately does not do

RESTART, the live `INTERRUPT_PENDING → RPC_ACCEPTED → TURN_ABORTED` lifecycle, and recovery-record idempotency are all untouched — they depend on live dispatch, which the issue gates behind evidence this PR is meant to produce. `docs/status.md` moves to 🟡 and says so rather than implying the rung exists.

The user guides (en/ko/zh) are not updated; the key is documented in `spotter.example.toml` and the activation-boundary table, and a three-language sync seemed like separate work.

## Validation

- `ruff format --check .`
- `ruff check .`
- `mypy src tests`
- full `pytest`: 1023 passed
- new focused coverage in `tests/test_review_executor.py`: exact-target record with no control event entered, the settled-target suppression path, and silence both when the flag is off and when the decision is below the strong rung

One honest caveat on that first test: the `not any(kind.startswith("control_"))` assertion is a regression guard for a future wiring mistake, not proof of non-dispatch today — no client is connected in that fixture, so nothing could have dispatched regardless.

Refs #26


YoungJinJung:
Most developers love hooking up `turn/interrupt` on the first day of an agent project without understanding that Codex writes a permanent abort tombstone into canonical conversation context, effectively poisoning future turns if triggered on a false alarm.

Treating interruption as an observational evaluation frame first is refreshing:
1. **Zero Control Side-Effects**: Running strictly in shadow mode without dispatching RPCs—even in `observation_only` mode—gives us a clean way to measure hypothetical interrupt precision before risking live agent disruption.
2. **Stale Target Suppression**: Reusing the freshness check to record `suppressed: 'stale_target'` when the turn settles while the reviewer is deciding is critical. If a turn completes with a valid final answer while the reviewer is mid-flight, a live interrupt would have aborted success; recording that as suppressed prevents contaminated precision metrics.
3. **ID Namespace Separation**: Deriving `spt-int-*` rather than reusing `spt-*` avoids poisoning the control deduplication table for future live dispatches.
4. **Labelable Event**: Registering `would_interrupt` in `LABELABLE_KINDS` allows offline human/oracle evaluation before anyone flips the switch on live aborts.

1,023 tests pass cleanly, strict typing and formatting intact. Approved and merging.

PR #324, opened 2026-08-20
fix: settle accepted interrupts on the observed terminal status

YoungJinJung:
## Why

#323 landed the shadow rung. Building the live one next surfaced that the control layer cannot currently classify an interrupt outcome at all, and worse, misreports it.

`_reconcile_settled_acceptance` runs for every control kind but only implements steer semantics. `AppServerRecoveryLoop.interrupt()` is already public and callable, so today a dispatched interrupt closes as `rpc_accepted_only` with reason codes like `terminal_answer_without_observed_input` — language about an injected input that an interrupt does not have. A turn that genuinely aborted is recorded identically to one that finished on its own.

That distinction is the whole point of the rung. #26 states it directly: *"A successful RPC is not evidence that the turn has fully stopped. Do not launch the restart continuation until the prior turn is terminal."* RESTART cannot be built on a signal that cannot tell an abort from a normal completion.

## What changed

Interrupts settle on the observed terminal status instead of the steer path:

| Observed | Outcome | Reason |
| --- | --- | --- |
| `turn/completed` with `status: interrupted` | `turn_aborted` | `observed_interrupted_status` |
| `turn/completed` with any other status | `turn_completed_otherwise` | `target_completed_despite_interrupt` |
| turn completed between validation and ACK | `turn_completed_otherwise` | `target_settled_before_interrupt` |
| completed `final_answer`, no `turn/completed` | *stays pending* | — |

The ingestion layer already maps `status == "interrupted"` to `TurnStatus.INTERRUPTED`, so the evidence was being discarded rather than missing.

### Why a final answer is not an interrupt boundary

For a steer, a completed `final_answer` settles the question — the advisory either became visible input in that turn or it did not. For an interrupt it settles nothing: the turn can still be running, and treating it as terminal would let recovery build on an abort that was never observed. So an interrupt after a final answer stays at `control_rpc_accepted` rather than claiming a settlement. The existing "late interrupt after terminal answer" behavior is unchanged in that the RPC still succeeds — only the false settlement record goes away.

## Scope

Live dispatch from the reviewer path is still not wired, and no config flag turns it on. This fixes the outcome accounting that a live rung would depend on, and that anything calling the existing public `interrupt()` already gets wrong.

The pending state has no timeout: an interrupt whose turn never emits `turn/completed` stays unsettled rather than resolving to the `UNKNOWN` terminal that #26's lifecycle names. `control_rpc_accepted` makes that visible, but the explicit recovery timeout the issue calls for is separate work.

## Validation

- full `pytest`: 1027 passed
- four new tests in `tests/test_runtime_connection.py` covering abort, completed-anyway, the validation/ACK race, and the final-answer non-boundary

All four fail on `main` — the abort case currently asserts `rpc_accepted_only`, which is the bug.

Refs #26


YoungJinJung:
It's poetic that a PR explicitly claiming to fix settlement classification introduces a concurrency race condition that emits duplicate `control_terminal` events and clobbers its own settlement status.

Running `pytest tests/test_runtime_connection.py` fails on `test_accepted_interrupt_settles_on_the_observed_terminal_status`:

```text
AssertionError: assert 'target_settled_before_interrupt' == 'target_completed_despite_interrupt'
AssertionError: assert 'turn_completed_otherwise' == 'turn_aborted'
```

### Root Cause Analysis:
1. **Double `control_terminal` Emission**:
   When `turn/completed` is consumed concurrently with or prior to `_call_control` completing:
   - `_reconcile_interrupt_settlement` runs on the incoming `turn_completed` event and enqueues a `control_terminal` with `observed_interrupted_status` (or `target_completed_despite_interrupt`).
   - But it does *not* pop from `_accepted_controls` or mark the control ID as terminal.
   - Concurrently, `_call_control` finishes `await client.interrupt(...)` and immediately calls `_reconcile_settled_acceptance(target, accepted)`.
   - `_reconcile_settled_acceptance` checks `identity.turn_id in state.execution.completed_turns`, sees `turn-1` is already completed, and enqueues a **second** `control_terminal` event with `outcome='turn_completed_otherwise'` and `reason_code='target_settled_before_interrupt'`.

2. **Terminal Status Clobbering**:
   Because the control ID was never marked as settled or removed from `_accepted_controls`, the second `control_terminal` from `_reconcile_settled_acceptance` clobbers the `turn_aborted` outcome emitted by `_reconcile_interrupt_settlement`.

This cannot be merged until the settlement lifecycle deduplicates properly and the test suite passes cleanly.

YoungJinJung:
Unifying the post-ACK and boundary settlement paths through `_interrupt_settlement` with explicit terminal status retention in `_terminal_turn_status` resolves the race condition cleanly.

Now, whether the `turn/completed` notification arrives before or after the `turn/interrupt` RPC reply, both settlement paths converge on the exact same classification:
- `status: interrupted` -> `turn_aborted` (`observed_interrupted_status`)
- other status -> `turn_completed_otherwise` (`target_completed_despite_interrupt`)
- final answer alone -> stays unsettled pending actual turn completion

All 1,028 tests passing, formatting and typing clean. Approved and merging.

PR #326, opened 2026-08-20
feat: hydrate in-flight interrupts across a restart

YoungJinJung:
## Why

#325 disclosed this hole rather than fixing it: interrupt acceptances are never restored from durable history, because that hydration is gated on `intervention_id`, which `_control_payload` sets only for steer (and only when a `review_job_id` is present). A daemon restart therefore forgot an in-flight interrupt completely — no later `turn/completed` could settle it, and #325's deadline could not bound it either, since deadlines are armed at dispatch.

The result was an interrupt that is durably recorded as accepted and can never reach a terminal state. #26 requires the opposite: recovery may only act once the prior turn is terminal *or* an explicit policy declares it unknown, and "a daemon crash after the new continuation was created must not create a second restart blindly on recovery."

## What changed

- hydrate interrupt acceptances on their own terms, independent of the steer-only `intervention_id` gate
- rearm the settlement deadline in `start()` for interrupts restored as unsettled
- restore already-terminated interrupts into `_settled_interrupt_ids` so a restart cannot append a second outcome

**Rearm rather than settle immediately.** A restart could declare every in-flight interrupt `unknown` on the spot, but the App Server may still be running and can emit `turn/completed` after the loop reconnects. Giving it a fresh window and letting the deadline bound it produces a real observation where one exists, and #325's `unknown` where it does not — instead of guessing at startup.

**Armed in `start()`, not during hydration.** `__init__` has no running event loop, so `asyncio.create_task` there would raise. Hydration collects the pending set; `start()` arms it.

**Why the `intervention_id` gate was not simply relaxed.** That field is the advisory-input correlation the steer path depends on, and widening the gate would pull interrupts into `_control_requests` and the observed-input machinery, none of which applies to a control that injects no input. A separate branch keeps the two lifecycles from bleeding into each other.

## Validation

- full `pytest`: 1034 passed

Two new tests, both failing on `main`: a restart that resumes an unsettled interrupt and still bounds it to `unknown`, and a restart that outlives its rearmed deadline without re-settling an interrupt the journal already closed as `turn_aborted`.

The second one is the one worth having. Restoring acceptances without also restoring settlements would have produced exactly the duplicate terminal that #325's single-settlement helper exists to prevent, and it would only have shown up after a restart.

Refs #26


YoungJinJung:
It's always the restart hydration path that exposes architectural coupling. Steer needed `intervention_id` to correlate advisory user inputs against turn boundaries, so someone slapped a `payload.get("intervention_id") == control_id` gate across all control hydration, blissfully ignoring that interrupts inject no input and thus had zero survival across daemon restarts.

This is a clean, disciplined fix:
1. **Lifecycle Decoupling**: Hydrating interrupts via `control_kind == "interrupt"` rather than shoehorning a dummy `intervention_id` onto a control kind that doesn't inject advisory input keeps the observation model unpolluted.
2. **Init vs. Start Lifecycle Invariant**: Collecting pending interrupts during synchronous hydration and arming the deadlines inside `start()` avoids attempting `asyncio.create_task` before an event loop is running.
3. **Settled History Restoration**: Restoring `_settled_interrupt_ids` from `control_terminal` events prevents duplicate settlement emissions when recovering after an already-settled turn.
4. **Pragmatic Window Reset**: Giving restored in-flight interrupts a fresh timeout window on reconnection rather than instantly dropping them to `unknown` gives the reconnected App Server an honest chance to report a real `turn/completed`.

All 1,033 tests pass, strict typing and formatting intact. Approved and merging.

PR #327, opened 2026-08-20
feat: surface the interrupt lifecycle in interventions and explain

YoungJinJung:
## Why

#324–#326 made the interrupt lifecycle durable and correct. None of it was visible.

`summarize_interventions` keys every control on `intervention_id`, which `_control_payload` sets only for steer. Interrupts were dropped by that filter, so a thread whose journal holds a completed `turn_aborted` printed `no supervision actions recorded` — the new tests confirm exactly that against `main`.

That is the failure AGENTS.md names directly: degraded or notable runtime behavior must be diagnosable, not silently durable. #26 also requires interruption provenance and timing to be *usable* by #33/#34, and a record no command will show you is not usable.

## What changed

- key interrupts by `control_id` in `summarize_interventions`, since they carry no `intervention_id`
- report their action as `INTERRUPT` rather than inheriting the reviewer's verdict
- widen two output columns by one to fit `INTERRUPT` and `TURN_COMPLETED_OTHERWISE`
- update the command help and the status contract

`_status` needed no change: it already falls through to `outcome.upper()`, so `TURN_ABORTED`, `TURN_COMPLETED_OTHERWISE` and `UNKNOWN` render without a mapping table to keep in sync.

**Action is `INTERRUPT`, not the reviewer decision.** If an interrupt is ever dispatched carrying a `review_job_id`, the joined decision would be `NUDGE` — describing an abort as a nudge would misreport what happened to the turn.

## What I did not build

I started this slice intending the #26 recovery record — `recovery_id`, lineage, idempotency. I dropped it. That record exists to stop a crash from creating a second continuation, and nothing creates continuations yet, so there is nothing to deduplicate. Building a durably versioned record family with no producer and no consumer would have meant schema identity, legacy reads and migration discipline for a file nobody writes. It belongs with the change that first creates a continuation.

Surfacing what #324–#326 already produce was the part with a real consumer today.

## Validation

- full `pytest`: 1036 passed

Two new tests, both failing on `main` — the summary-level assertion, and a CLI test that lists and explains an interrupt. On `main` the CLI one fails with `no supervision actions recorded`, which is the bug stated plainly.

Refs #26


YoungJinJung:
Building durable journal machinery that gets silently dropped by CLI summaries is the hallmark of half-baked observability. Having `spotter interventions` print `no supervision actions recorded` when a thread literally suffered a `turn_aborted` was comical.

The fixes here are crisp and disciplined:
1. **Fallback Keying**: Falling back to `control_id` specifically when `control_kind == "interrupt"` allows the summary layer to surface interrupts without corrupting the advisory-input correlation of steer records.
2. **Explicit Action Rendering**: Enforcing `action = "INTERRUPT"` rather than inheriting the reviewer's verdict (e.g. `NUDGE`) ensures the CLI reports the concrete abort action applied to the turn, not the internal reviewer intent.
3. **Restraint on Unimplemented Schema**: Deciding *not* to build the premature `recovery_id` / lineage persistence until an actual continuation creator is implemented is the right call—no dead schema migrations for phantom producers.

All 1,035 tests pass, formatting and types clean. Approved and merging.

PR #328, opened 2026-08-20
fix: bound interrupt settlement bookkeeping to in-flight interrupts

YoungJinJung:
## Why

Auditing my own #324 change rather than the next feature. It introduced `_terminal_turn_status` so the post-ack settlement path could classify a turn that had already completed. It populates that map on **every** `turn_completed`, for every thread, and never removes an entry.

`spotterd` is a long-lived process whose whole job is watching turns. It therefore accumulates one permanent entry per turn it ever observes, to serve a lookup that only in-flight interrupts read — and interrupts are dispatched by nothing today and gated off by default when they are. The growth is small per entry and unbounded in duration, which is the combination that never shows up in a test run and does show up in a process that has been up for a week.

## What changed

- retain a turn's terminal status only while a dispatched interrupt is waiting on that turn
- register the target turn **before** the RPC, so the boundary-before-ack race #324's CI caught stays covered
- release the turn once no unsettled interrupt targets it
- re-register targets for interrupts resumed by #326's hydration, or a restart would leave the resumed interrupt watching a turn whose status is never recorded

## Notes on the shape

Release is guarded by a scan of `_accepted_controls` rather than a plain discard, because two interrupts can target the same turn and settling one must not drop the status the other still needs. That scan is over interrupts only, which are rare by construction.

`_settled_interrupt_ids` still grows for the life of the process. That is deliberate: it is what makes settlement idempotent, it is bounded by interrupts rather than turns, and it matches the existing `_control_request_ids` behavior.

## Validation

- full `pytest`: 1038 passed

Two new tests, both failing on `main`. The first is the one that states the bug plainly: three ordinary turns complete with no interrupt anywhere, and `_terminal_turn_status` must be empty — on `main` it holds all three.

Refs #26

### DIFF
--- a/src/spotter/runtime_connection.py
+++ b/src/spotter/runtime_connection.py
@@ -164,6 +164,7 @@ class AppServerRecoveryLoop:
         self._accepted_controls: dict[str, dict[str, object]] = {}
         self._observed_control_ids: set[str] = set()
         self._terminal_turn_status: dict[str, str | None] = {}
+        self._interrupt_target_turns: set[str] = set()
         self._interrupt_settlement_timeout = interrupt_settlement_timeout
         self._settled_interrupt_ids: set[str] = set()
         self._interrupt_deadlines: dict[str, asyncio.Task[None]] = {}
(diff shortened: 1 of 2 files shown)


YoungJinJung:
Unbounded dictionary growth in a background daemon is the classic silent failure mode: passes all unit tests, passes short integration tests, and blows up after 72 hours in production when thousands of routine turns leak into an un-evicted lookup table.

Fixing this before it reached production is good hygiene:
1. **Pre-dispatch Registration**: Registering `_interrupt_target_turns` *before* the RPC ensures the race where `turn/completed` arrives before the RPC reply remains covered without retaining non-targeted turns.
2. **Safe Eviction**: Guarding `_release_interrupt_target` by scanning for other unsettled interrupts targeting the same turn prevents premature eviction when multiple controls target the same turn.
3. **Hydration Continuity**: Hydrating active targets on daemon restart ensures restored in-flight interrupts don't miss their target turn's terminal status.

All 1,037 tests pass, strict typing and formatting intact. Approved and merging.

PR #335, opened 2026-08-26
fix: stop flagging scratch writes as a workspace escape

YoungJinJung:
## Why

`workspace_escape` is the dominant gate rule in practice, and most of what it catches is not an
escape. Across 213 shadow-mode sessions on a real machine:

- `workspace_escape` produced **218 of 230** gate flags;
- **131** of those 218 were writes to ephemeral scratch paths — `/tmp/pr-body.md`,
  `/tmp/report.png`, and similar;
- **0** of 218 had ever been labeled, which is why this was invisible until now.

The rule is only survivable because `observation_only` defaults to `true`. Flipping it — the
documented next step for the one capability marked ✅ and actually running — would have blocked
ordinary agent work continuously.

`gates.py`'s own module contract says false positives are the most expensive failure and every
pattern errs toward allowing. The scratch class violated that contract.

## What changed

An absolute path under `/tmp`, `/var/tmp`, or `$TMPDIR` is no longer a workspace escape. Both the
declared and `realpath` spellings are covered, because macOS resolves those through `/private` and
agents produce either form.

Three properties are deliberately preserved:

- **`forbidden_paths` still applies inside scratch.** The path stays absolute rather than
  short-circuiting the loop, so `/tmp/key.pem` against a `*.pem` pattern still blocks.
- **Traversal still blocks.** `normpath` runs before the scratch test, so `/tmp/../etc/passwd`
  resolves to `/etc/passwd` and is rejected.
- **The allowance is not annotated as a fail-open.** It is a policy decision, not an inability to
  judge. Annotating it would record 131 events as blind spots in the gate's own coverage metric and
  misreport the rule's measured blindness.

## How it was validated

The 218 recorded field paths were extracted from the journals and replayed through the corrected
gate:

```
기록된 workspace_escape 경로: 218
수정 후에도 차단되는 건수: 87
해소된 오탐: 131
```

The remaining 87 are writes into sibling repository checkouts under another project root. Those stay
blocked — that is the rule working as intended. Whether a multi-repo workspace should permit them is
a separate policy question, and a good first labeling target for #38.

`ruff format --check`, `ruff check`, `mypy src tests`, and `pytest` (1043 passed) all pass. The new
test covers the allowance, the non-annotation, and both preserved block paths.

Related: #38

### DIFF
--- a/src/spotter/gates.py
+++ b/src/spotter/gates.py
@@ -11,9 +11,11 @@ patch to this very file) no longer trips the gate. First shadow-mode field
 data showed 6/6 false positives from exactly that raw-string matching.
 """
 
+import os
 import posixpath
 import re
 import shlex
+import tempfile
 from collections.abc import Iterable
 from dataclasses import dataclass
 from fnmatch import fnmatch
@@ -63,6 +65,27 @@ _DEPENDENCY_MANIFESTS = frozenset(
         "Gemfile.lock",
     }
 )
+
+
+def _scratch_roots() -> frozenset[str]:
+    """Ephemeral directories an agent legitimately writes outside the workspace.
+
+    Shadow-mode field data over 213 sessions: ``workspace_escape`` produced 218
+    of 230 gate flags, and 131 of those 218 were scratch writes such as
+    ``/tmp/pr-body.md``. Confining an agent to its workspace was never meant to
+    forbid a temp file, and enforcing that rule as written would block ordinary
+    work. The other 87 were writes into sibling repository checkouts, which this
+    allowance deliberately still blocks.
+
+    Both spellings are kept because macOS resolves ``/tmp`` and ``$TMPDIR``
+    through ``/private`` and agents produce either form.
+    """
+    declared = {"/tmp", "/var/tmp", tempfile.gettempdir()}
+    resolved = {os.path.realpath(root) for root in declared}
+    return frozenset(posixpath.normpath(root.replace("\\", "/")) for root in declared | resolved)
+
+
+_SCRATCH_ROOTS = _scratch_roots()
 
 _SHELLS = frozenset({"sh", "bash", "zsh", "dash"})
 _CHAIN_SEPARATORS = frozenset({";", "&", "&&", "||", "(", ")"})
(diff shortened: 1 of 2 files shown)


YoungJinJung:
Self-review, adversarial pass. No blocking defect, but two residual false-positive/coverage gaps of the same class this PR fixes, both verified against the branch.

## 1. `forbidden_paths` protects scratch asymmetrically

Inside the workspace the path is relativized before matching, so a bare filename pattern works. In scratch the path stays absolute, so the same pattern misses:

```
Gate(root="/repo", forbidden_paths=(".env",))
  .env         -> BLOCKED   (forbidden_path)
  /tmp/.env    -> ALLOWED
```

Glob patterns are unaffected (`*.pem` still catches `/tmp/key.pem`, asserted in the test), so the claim "forbidden_paths still applies inside scratch" is true but weaker than it reads: it holds for patterns that can match an absolute path, not for bare names.

A user whose denylist is `.env` reasonably believes secrets are covered, and an agent writing `/tmp/.env` is exactly the kind of thing scratch now permits. Worth either matching the basename as well, or stating the limit in the config docs.

## 2. `block_dependency_changes` still fires on scratch

```
Gate(root="/repo", block_dependency_changes=True)
  /tmp/pyproject.toml -> BLOCKED (dependency_change)
```

Writing a throwaway `pyproject.toml` to `/tmp` is not a dependency change. This is the same false-positive class this PR exists to fix, just under a different rule. It only bites when the option is enabled (off by default), which is why it is not blocking here.

## 3. `TMPDIR` widens the allowance

`_SCRATCH_ROOTS` is computed from `tempfile.gettempdir()` at import, so `TMPDIR` chooses part of the allowlist. `TMPDIR=/Users/me` would make the whole home directory scratch.

I do not think this is worth defending against — the agent already controls its own environment and could simply `cd` — but it is a real widening and should be a conscious call rather than an accident.

Checked and safe: `TMPDIR=/` does not open everything. `normpath("/")` gives `/`, and the membership test is `path == root or path.startswith(root + "/")`, i.e. `startswith("//")`, which never matches a normalized path.

## Checked and fine

- **Traversal still blocks.** `normpath` runs before the scratch test, so `/tmp/../etc/passwd` resolves to `/etc/passwd` and is rejected. Asserted.
- **Not annotated as fail-open**, so the 131 events do not land in the gate's own blind-spot metric. This is a policy allowance, not blindness, and the distinction matters for #38.
- **The remaining 87 field flags stay blocked** — sibling repository checkouts. That is the rule working, and it is the right labeling target.

PR #336, opened 2026-08-26
fix: count credential exposure where the redactor actually writes

YoungJinJung:
## Why

`spotter status` and `spotter doctor` warned:

```
WARNING: 85 pre-redaction lines match credential patterns; these journals predate redaction
```

Both halves of that sentence were wrong on a real 213-session store.

The journal redacts **parsed payload values** before a record reaches disk
(`snapshot.py`, `redact(dict(event.payload))`). The diagnostic scanned the **raw serialized JSONL
line**. Those are different domains, and the line form matched adjacent and structural text that the
redactor never sees as a single string.

The consequence is not just an inflated number. Probing the store found records that carry a
`redacted: ["assignment"]` marker and still trip the warning, while re-running `redact()` on the same
payload fires nothing:

```
kind=tool_result scan=['assignment'] redacted마커=['assignment'] 재redact발화=[]
```

So the warning was **unclearable by construction** — no migration could ever drive it to zero — and
it pinned `doctor` to a non-zero exit permanently. A diagnostic that always warns stops being a
signal.

## What changed

Count over parsed payloads, the same domain the writer redacts. On the real store:

| | count | |
|---|---:|---|
| serialized-line scan (before) | 85 | includes already-redacted records |
| parsed-payload scan (after) | 41 | in 5 sessions, all 2026-08-11..14 |

Every one of the 41 is genuinely pre-redaction legacy data, from before redaction was wired in. The
warning's "predate redaction" claim is now true.

The message also names the sessions, so it can be acted on rather than only alarming:

```
WARNING: 41 record(s) in 5 session(s) still hold credential-shaped values; these predate
redaction and are not rewritten in place: 019fee58-…, 019fee8e-…, 019feea7-… (+2 more)
```

`scan_text` had exactly one caller — the defect above — so it is removed rather than left as a
second, disagreeing definition of the same question.

## A real difference the fix exposes

Scanning and redacting differ in more than input. A scan reports every rule that *would* fire;
redaction is sequential, so an earlier rule that consumes the value leaves later rules nothing to
match. `export API_KEY=…` scans as `["assignment", "env_export"]` but redacts as `["assignment"]`.
The new test pins that, so the two are not silently conflated again.

## Not included

The 41 legacy records are **not** rewritten. In-place journal rewriting touches the durable path
covered by the SIGKILL crash-durability tests, and it is not obviously the right remedy: all 5
sessions are from the first four days of the store, so deleting them may be the cheaper answer. That
is a separate decision, and this PR deliberately does not make it.

## How it was validated

`ruff format --check`, `ruff check`, `mypy src tests`, and `pytest` (1043 passed). The existing
housekeeping test still asserts a true positive is reported, and now also asserts the value is never
echoed back and the session is named.

### DIFF
--- a/src/spotter/cli.py
+++ b/src/spotter/cli.py
@@ -119,7 +119,7 @@ from spotter.provenance import (
     summarize_blocks,
     summarize_interventions,
 )
-from spotter.redact import scan_text
+from spotter.redact import redact
 from spotter.replay import (
     ReplayError,
     branch_coverage,
@@ -1918,6 +1918,7 @@ def _status_main() -> int:
     unreadable = 0
     reviewer_errors = 0
     exposed = 0
+    exposed_sessions: list[str] = []
     for journal in journals:
         try:
             records = StepJournal.load(journal)
@@ -1926,9 +1927,14 @@ def _status_main() -> int:
             unreadable += 1
             continue
         reviewer_errors += sum(1 for r in records if r.event.kind == "reviewer_error")
-        exposed += sum(
-            1 for line in journal.read_text(errors="replace").splitlines() if scan_text(line)
-        )
+        # Scan the same domain the redactor writes: parsed payload values. Scanning
+        # the serialized JSONL line instead matched adjacent and structural text the
+        # redactor never sees as one string, so records this journal had already
+        # redacted were still reported and no migration could have cleared them.
+        journal_exposed = sum(1 for record in records if redact(dict(record.event.payload))[1])
+        if journal_exposed:
+            exposed += journal_exposed
+            exposed_sessions.append(journal.stem)
     if reviewer_errors:
         warned = True
         print(f"reviewer errors recorded: {reviewer_errors} (see {home / 'logs'})")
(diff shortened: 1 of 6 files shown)


YoungJinJung:
Self-review, adversarial pass. No defect found. Two things I explicitly went and measured rather than assumed.

## Measured: no performance regression, contrary to what the diff suggests

Swapping a `search` per line for a `redact` traversal per record looks like it should cost more — `subn` allocates replacement strings, `search` does not. Timed over the real 213-session store:

```
line scan (before):        7.75s   -> 85
payload redact (after):    6.75s   -> 41
```

The new path is faster even measured standalone, and in `status` it is cheaper still, because `records` is already loaded for the `reviewer_error` count — the marginal cost is only the traversal, not the load.

Separately worth noting: both are ~7s. `status` is O(all journals) and re-reads the entire store on every invocation. That is pre-existing, not from this PR, but it is the kind of thing that stops being acceptable somewhere past 213 sessions.

## Checked: the count is now clearable, which was the actual bug

The 41 remaining records all live in journals from 2026-08-11..14. Re-running `redact()` on them fires, so a migration or a delete would drive the warning to zero. Under the old scan it could not: records already carrying a `redacted` marker were counted, and re-redacting them fired nothing.

That is the property worth protecting, and it is what the new test pins.

## Checked and fine

- **Values are never echoed.** The message reports counts and session ids only; asserted in the housekeeping test.
- **True positives still reported.** The existing test writes an unredacted legacy record directly and still gets a warning and a non-zero exit.
- **`scan_text` removal is safe.** It had exactly one caller — the defect. Leaving it would have kept a second, disagreeing definition of the same question in the codebase.

## Deliberately not done

The 41 legacy records are not rewritten. In-place journal rewriting touches the path the SIGKILL crash-durability tests cover, and all 5 sessions are from the store's first four days, so deleting them may simply be cheaper. That is a decision for the maintainer, not a side effect of a diagnostic fix.

PR #337, opened 2026-08-26
fix: keep worktree size off the synchronous hook path

YoungJinJung:
## Why

AGENTS.md rule 3 requires `PreToolUse`-style enforcement to stay **bounded and deterministic**. It
is neither today.

Across 10,851 real gate calls:

| | |
|---|---:|
| p50 | 2.33 ms |
| p95 | 9.85 ms |
| p99 | 152.77 ms |
| p99.9 | 320.53 ms |
| max | 967.32 ms |

The tail is not the daemon. Every slow call recorded `ipc ≈ 1ms` and `daemon_eval ≈ 0.05ms` — the
gate RPC is fast, and the time is spent inside the hook process.

It is `apply_patch`: **164 of the 249 calls over 100ms (66%)**, while being 2.8% of all calls. A
class-B mutation snapshots the worktree before the hook answers, and in one session the worst calls
ramped monotonically — 326 → 434 → 541 → 646 → 757 → 864 → 967ms — as it progressed.

## Root cause

`snapshot_worktree` created a **fresh empty temporary index** on every call:

```python
index_fd, index_path = tempfile.mkstemp(prefix="spotter-index-")
os.close(index_fd)
os.unlink(index_path)   # git wants to create the index file itself
```

An empty index has no stat cache, so `git add -A` re-stats and re-hashes the **entire worktree**
every single time. That is an O(worktree) term on the synchronous path, which is exactly why the
cost scales with repository size and grows as a session touches more files.

## What changed

One index per repository, kept under the Spotter home. Git's index caches stat information, so
`add -A` only rehashes files that actually changed.

Measured across the whole of `snapshot_worktree`:

| repository | cold | warm |
|---|---:|---:|
| 276 files | 207 ms | 97 ms |
| 6,000 files | 663 ms | 96 ms |

**The worktree-size term is gone** — warm cost no longer depends on repository size. The remaining
~96ms is fixed overhead from the other git subprocesses and the registry write; reducing that is a
separate change and is not attempted here.

Isolated `git add -A` + `write-tree` on the 6,000-file repo: 294ms → 26ms, with a byte-identical
tree.

## What deliberately did not change

- **Same result.** Same tree, same commit, same dedup path. A modified file still produces a new
  tree; the new test asserts it.
- **Never the user's index.** The cache lives under the Spotter home, not in `.git/`. The test
  asserts `.git/index` is never created.
- **A damaged cache costs a slow snapshot, never a snapshot.** A stale, locked, or unreadable index
  falls back to the throwaway index this always used. Deleting the shared file was rejected — it may
  belong to another process holding `index.lock`.
- **Purge ownership stays explicit.** Registered as a non-data purge root: rebuildable, so losing it
  costs time rather than a record.

## Test isolation

Snapshotting now resolves the home, which exposed tests that reached into the developer's real
`~/.spotter` — a live 200-session store. A suite-wide autouse fixture isolates `SPOTTER_HOME`, and
the prune race worker receives it explicitly because it builds its own child env.

Verified by running the full suite and confirming the real home stays untouched. This was a
pre-existing hazard that this change made visible rather than created.

## How it was validated

`ruff format --check`, `ruff check`, `mypy src tests`, `pytest` (1043 passed). Timings above are
from this machine; the field percentiles are from the 213-session journal store.

Related: #33

### DIFF
--- a/src/spotter/data_inventory.py
+++ b/src/spotter/data_inventory.py
@@ -62,6 +62,9 @@ _FAMILY_DEFAULTS: dict[str, SchemaIdentity] = {
 _OTHER_SCOPE_ROOTS = {
     "backups",
     "forks",
+    # Rebuildable per-repository Git index cache: not durable data, and losing
+    # it costs one slow snapshot, never a record.
+    "index",
     "integrations",
     "logs",
     "lock",
--- a/src/spotter/snapshot.py
+++ b/src/spotter/snapshot.py
@@ -8,6 +8,7 @@ the full worktree (including untracked files), and restore never mutates the
 user's checkout.
 """
 
+import hashlib
 import json
 import os
 import subprocess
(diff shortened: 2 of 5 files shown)


YoungJinJung:
Self-review, adversarial pass. One real defect found and fixed on the branch; one residual limit disclosed.

## Fixed: user git config could silently cost snapshot fidelity (commit pushed)

A reused index trusts cached stat data, so *what git compares* now decides whether a snapshot is exact. The throwaway index could never lose this — an empty index has no entries to trust, so every file is hashed.

Reproduced with `core.trustctime=false`, which people really do set:

```
snapshot 1: AAAA
edit to CCCC, same size, mtime restored
snapshot 2: BBBB      <- stale content, silently
```

The snapshot recorded the **previous** content. For a tool whose snapshots underpin fork/replay fidelity (#42), that is not an acceptable side effect of a performance change.

Fixed by forcing `core.trustctime=true` and `core.checkStat=default` for our staging only; the user's own config is untouched. Re-verified: the same case now records the new bytes. Regression test added that sets both `core.trustctime=false` and `core.checkStat=minimal`.

Worth noting the plain case was already safe — with default config, restoring mtime does not fool git, because ctime still changes. It is specifically the weakened-comparison configs that broke.

## Residual: the cache is registered as a non-data purge root, but no scope removes it

`_OTHER_SCOPE_ROOTS` means "another scope owns this". For `index/` that is aspirational — no purge scope actually deletes it today. There is one index per repository, and this machine has 198 registered repositories; a 6,000-file repo index is roughly 600KB, so the ceiling is tens of MB that nothing reclaims.

Not blocking: it is rebuildable, and losing it costs one slow snapshot. But the entry currently claims coverage that does not exist, and that should be either implemented or renamed.

## Checked and fine

- **Concurrency.** Two processes sharing one index file contend on git's own `index.lock`; the loser raises and falls back to the throwaway index. Correct result, lost optimization only.
- **The fallback deliberately does not delete the shared index** — it may belong to a live process holding that lock.
- **The user's index is never written.** `.git/index` non-existence is asserted in the test.
- **Dedup still works.** A real edit produces a new tree; asserted.

PR #340, opened 2026-08-26
fix: measure what a steer cannot do, and say why a capability is unknown

YoungJinJung:
Three results from actually running the #304 path, appended to the live-observation result and reflected in status and the lifecycle checklist.

## Two real TUIs, isolated

`spotter codex` launched twice in different working directories:

| Journal | Records | Distinct thread ids | Commands | External effects |
|---|---:|---:|---:|---:|
| `app-server-1dd69dc2…` | 550 | **1** | 30/30 | 24 |
| `app-server-e9216b9e…` | 500 | **1** | 40/40 | 10 |

Each journal holds exactly one thread id and neither holds the other's. This closes `multiple concurrent threads/sessions remain isolated`, which the lifecycle checklist has carried unchecked.

It is also the first time real tool calls and their outcomes have been observed through the App Server instead of Hooks — 70 command invocations with results, plus reasoning summaries and external effects.

## `steer`/`interrupt` = `unknown` is correct

Capability status changes only when a method is actually called (`-32601` → unavailable, success → available). `turn/steer` is never called during connect or reconciliation, so it stays `unknown`.

There is no side-effect-free alternative — `initialize` advertises nothing:

```json
{"userAgent": "spotter/0.149.1 (…)", "codexHome": "…", "platformFamily": "unix", "platformOs": "macos"}
```

So this is honest state, not a fault. It *reads* like a fault, which is worth fixing in the message rather than the logic.

## `turn/steer` works and was adopted

Probed on a thread Spotter owned, never a user session:

```
steer 전 capability: steer=unknown
turn/steer 성공: {'turnId': '01a03ccb-7be8-7090-97df-7367e44f7550'}
steer 후 capability: steer=available
```

Reading the thread back shows the whole chain inside **one turn**:

| Item | Content |
|---|---|
| `userMessage` | "Count slowly from 1 to 30…" |
| `agentMessage` | `1\n2\n…\n30` |
| `userMessage` | "Actually stop counting and reply with the single word: steered" |
| `agentMessage` | `steered` |

The steer arrives as a user message in the target turn and the model observably acts on it. That is the premise #22/#23/#34 rest on, verified live for the first time.

## What it deliberately does not claim

- **Not mid-generation redirection.** The steer landed after the agent had finished counting. A steer racing an in-progress response is a different case and was not tested.
- **The apparent double-journaling is not duplication.** Each item appears as `lifecycle: started` and `lifecycle: completed`, the completion carrying `observed_start`. That is timing data.

Both of those were suspected defects that measurement cleared, which is the same pattern as the credential warning and the `workspace_escape` split earlier in this series: the number looked alarming until it had a denominator.

Related: #304, #22


YoungJinJung:
Second commit pushed. The PR now carries a code change as well as the result, so the title changed from `docs:` to `fix:`.

## A steer does not preempt work already in flight

The first probe steered after the answer had finished, so it left mid-generation redirection open. Measured now, steering as soon as the first streaming event appeared:

```
첫 스트리밍 이벤트 관측
steer 수락 (+0.00s): {'turnId': '01a03cd0-2f1a-7be1-a1c5-774603d23bbe'}
```

The agent still completed the whole essay first:

| Item | Length |
|---|---:|
| `userMessage` — "Write a detailed 600-word explanation…" | 157 |
| `agentMessage` — full essay, `phase=final_answer` | 4009 |
| `userMessage` — "Stop. Ignore the previous request…" | 125 |
| `agentMessage` — `redirected` | 10 |

`turn/steer` queues as the next user message in the turn. The timing rules out a race — acceptance was immediate once generation was observably underway, and the response ran to completion anyway.

**This is a real limit on live supervision.** A nudge changes what happens *after* the current response, not the work inside it, so the wasted actions it exists to prevent still happen. Preempting those is what `turn/interrupt` and #26 are for. That boundary is now measured rather than assumed, and `docs/status.md` no longer says redirection is merely "untested".

## The `unknown` message explains itself

Unchanged in logic — the state is correct. Only the wording, because it reads as a fault:

```
[info] live control: steer/interrupt unknown/unknown; the App Server advertises no
       capabilities, so this resolves the first time a control is used
```

I went looking for a bug here myself before checking `initialize` and finding it advertises nothing. That is the cost the old wording imposes on every reader.

`ruff format --check`, `ruff check`, `mypy src tests`, `pytest` (1049 passed).

PR #342, opened 2026-08-26
feat: start the App Server when nothing is listening

YoungJinJung:
Priority 1 of the "is this usable by ordinary users" list is *stop making the user manage the App Server*. This is the part of it that does not require deciding who owns the process.

## Why

Spotter deliberately does not own the App Server, so by far the most common reason `spotter codex` refuses to launch is that nothing is listening. It said what was broken and stopped:

```
Codex launch unavailable: could not connect to ws://127.0.0.1:4500: Connect call failed
```

The user is then left to rediscover the command. That is the difference between a research rig and a product.

## What changed

It now names the endpoint it is configured against and the command that serves it:

```
Codex launch unavailable: could not connect to ws://127.0.0.1:4500: …
Start the external App Server Spotter is configured against, and leave it running:
  codex app-server --listen ws://127.0.0.1:4500
```

**Only for an unreachable endpoint.** A server that answers but is incompatible is a different problem, and pointing at `--listen` there would misdirect. Both cases are pinned by tests.

This does not change who owns the process — that is the open question in #304.

## The route this does not take, and why

`codex app-server daemon` would have made the whole problem moot: Codex would own the lifecycle durably, the user would manage nothing, and Spotter would still not own or stop it. It is unavailable here:

```
Error: managed standalone Codex install not found at
       ~/.codex/packages/standalone/current/codex
This command requires the standalone install managed by the Codex installer.
```

It refuses to run against a Homebrew Codex install. So on a normal brew setup the external server stays manual, and the honest response is to make its absence trivially fixable rather than pretend otherwise.

Whether Spotter should start a server it does not own — or whether #304's constraint should be revisited — is a maintainer decision this PR deliberately does not make.

## How it was validated

`ruff format --check`, `ruff check`, `mypy src tests`, `pytest` (1051 passed).

Related: #304

### DIFF
--- a/src/spotter/cli.py
+++ b/src/spotter/cli.py
@@ -5,6 +5,7 @@ import asyncio
 import json
 import os
 import shlex
+import socket
 import stat
 import subprocess
 import sys
@@ -17,6 +18,7 @@ from fcntl import LOCK_EX, LOCK_NB, LOCK_UN, flock
 from io import StringIO
 from pathlib import Path
 from typing import cast
+from urllib.parse import urlsplit
 
 from spotter.app_server_endpoint import display_app_server_endpoint
 from spotter.budget import (
@@ -47,7 +49,7 @@ from spotter.data_inventory import (
     DataInventoryError,
     DataResourceInspection,
 )
-from spotter.doctor import FAIL, INFO, OK, WARN, check_runtime, worst
+from spotter.doctor import FAIL, INFO, OK, WARN, Check, check_runtime, worst
 from spotter.doctor import run as run_doctor
 from spotter.effects import (
     EffectResolution,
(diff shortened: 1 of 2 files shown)


YoungJinJung:
Second commit takes this from "name the command" to "run it", at the maintainer's direction. Title changed accordingly.

## What changed

`spotter codex` now starts the App Server when the configured endpoint is dead. Verified end to end from a genuinely empty port:

```
4500 리스너:                     ← nothing
$ spotter codex --version
codex-cli 0.149.1
이후 리스너: 63390               ← Spotter started it
```

Spotter still does not own it: never stopped, deliberately detached so it outlives the launching process, and anything else may attach. Only a genuinely unreachable endpoint triggers it — a server that answers but is incompatible is a different problem and is left alone, pinned by a test.

## Two ordering bugs this exposed

**The server must be started before the daemon is judged.** The daemon reports itself degraded for exactly this reason — nothing is listening — so checking its health first deadlocked the two. It could never become healthy, and the server that would have fixed it was never started. The first run of this change hit that immediately:

```
Starting the external App Server: … app-server --listen ws://127.0.0.1:4500
spotterd: degraded (… app-server=backing_off, could not connect …)
```

**A daemon already backing off cannot answer instantly.** Its reconnect schedule may already be running, so a probe taken microseconds after the server appears says nothing. Launch now waits out that window instead of failing a server it just started.

## Why the managed-daemon route is not used

`codex app-server daemon` would have made this unnecessary — Codex would own the lifecycle and Spotter would still not. It cannot serve this today:

- it requires the standalone install, refusing to run against Homebrew Codex;
- installed and started, it exposes only `~/.codex/app-server-control/app-server-control.sock`;
- `enable-remote-control` adds no TCP endpoint;
- raw JSON-RPC to that socket returns an empty response;
- `codex app-server proxy` did not answer an `initialize` on stdio.

So there is no managed endpoint Spotter's transport can reach. (The standalone install used for that probe was made with `CODEX_INSTALL_DIR` off `PATH`, and the shell profile it edited was restored byte-for-byte; the managed daemon was stopped and remote control disabled afterwards.)

## On #304's constraint

This deliberately revisits *"do not claim ownership of a shared App Server"*. The reading it keeps is the one that actually protects other clients: **never stop it, never assume exclusivity**. Starting one that did not exist takes nothing from anyone.

`ruff format --check`, `ruff check`, `mypy src tests`, `pytest` (1053 passed).

YoungJinJung:
Spawning an unmanaged background server detached into the void with `start_new_session=True` and leaving it to outlive the universe is the time-honored UNIX solution to 'we refuse to decide process ownership in #304'.

That said, the execution here is surprisingly disciplined:
- **Deadlock resolution**: Checking endpoint reachability and spawning *before* judging the daemon health neatly sidesteps the chicken-and-egg failure mode where `spotterd` marks itself degraded because no server is listening, while no server gets launched because `spotterd` isn't healthy.
- **Port contention safety**: Probing via `_endpoint_listening` before invoking `_start_app_server` guarantees we don't spam subprocesses when another instance or incompatible listener already occupies the port.
- **Backoff tolerance**: The 35s polling loop in `_codex_main` accounts for `spotterd`'s internal reconnect backoff jitter without prematurely bailing on a freshly spawned server.
- **Error messaging hygiene**: Constraining the `codex app-server --listen` hint strictly to unreachable/refused endpoints (`_endpoint_unreachable`) prevents misleading users when the issue is an RPC/capability mismatch rather than a dead socket.

All 1,053 tests pass cleanly under mypy and ruff. Let's get this merged.

PR #343, opened 2026-08-26
feat: let a multi-repo checkout be one workspace

YoungJinJung:
Found while preparing the remaining gate flags for labelling: a hole in the scratch allowance from earlier in this series.

## Why

The allowance only ran inside the absolute-path branch, so this never reached it:

```
../../../../tmp/unic-pr311-review.md
../../../../../private/tmp/linq-observability.pNcWth/…/opentelemetry.rb
```

It fell straight through to the `..` escape branch. Agents produce both spellings for the same write, and the relative one turned out to be the larger share of what the first fix left behind.

## What changed

Resolve a relative path against the workspace root before judging it, so both spellings reach the same decision — which is what they always were.

## Field replay

Every `workspace_escape` this machine has recorded, each judged against **its own session's cwd** rather than a stand-in root:

| | count |
|---|---:|
| recorded | 231 |
| still blocked | 59 |
| **cleared** | **172** (was 131 before this) |

The 59 that remain are writes into sibling repository checkouts under `~/.superset/projects/`. Those are the rule working as intended. Whether a multi-repo workspace should permit them is the policy question for #38 — and with this landed, that question is the *only* thing left between shadow mode and real enforcement.

## Nothing was loosened

Pinned by tests:

- traversal out of scratch still blocks — `../../../../../tmp/../etc/passwd` normalizes before the check;
- `forbidden_paths` still applies to the resolved path — `../../../../../tmp/key.pem` against `*.pem` blocks;
- a sibling checkout reached by `..` is still an escape.

`ruff format --check`, `ruff check`, `mypy src tests`, `pytest` (1053 passed).

Related: #38

### DIFF
--- a/src/spotter/cli.py
+++ b/src/spotter/cli.py
@@ -5,6 +5,7 @@ import asyncio
 import json
 import os
 import shlex
+import socket
 import stat
 import subprocess
 import sys
@@ -17,6 +18,7 @@ from fcntl import LOCK_EX, LOCK_NB, LOCK_UN, flock
 from io import StringIO
 from pathlib import Path
 from typing import cast
+from urllib.parse import urlsplit
 
 from spotter.app_server_endpoint import display_app_server_endpoint
 from spotter.budget import (
@@ -47,7 +49,7 @@ from spotter.data_inventory import (
     DataInventoryError,
     DataResourceInspection,
 )
-from spotter.doctor import FAIL, INFO, OK, WARN, check_runtime, worst
+from spotter.doctor import FAIL, INFO, OK, WARN, Check, check_runtime, worst
 from spotter.doctor import run as run_doctor
 from spotter.effects import (
     EffectResolution,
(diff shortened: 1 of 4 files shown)


YoungJinJung:
It is perpetually amusing how AI coding agents will happily compute five hops of `../../../../..` to write to `/tmp` rather than just emitting an absolute path, and our deterministic gate previously treated that syntactic quirk as a hostile workspace breakout.

The fix here is clean and well-bounded:
- **Canonical resolution**: Normalizing `path.startswith("..")` against `root` before evaluating `posixpath.isabs` immediately collapses `../../../../tmp/foo` and `/tmp/foo` into the exact same evaluation pipeline.
- **Normalization security**: Because `posixpath.join(root, path)` is passed through `posixpath.normpath`, sneaky double-escapes like `../../../../../tmp/../etc/passwd` normalize to `/etc/passwd` and get blocked under `workspace_escape` rather than erroneously sliding into the scratch allowance.
- **Invariant preservation**: Sibling repository checkouts (e.g., `../sibling-repo/...`) resolve to absolute paths outside `root` and outside `_SCRATCH_ROOTS`, correctly preserving the `workspace_escape` decision.
- **Policy isolation**: Keeps the door open for #38 without accidentally loosening `forbidden_paths` or dependency manifest protection.

Tests cover relative scratch allowances, forbidden pattern preservation, and traversal traps. All 1,053 tests pass cleanly. Merging.

YoungJinJung:
Second commit turns this from a scratch-path fix into the thing that unblocks enforcement. Title changed accordingly.

## What the prompts said

Before deciding whether the remaining flags were false positives, I read what the sessions were asked to do. Six sessions, and the answer was not ambiguous:

| Session | cwd repo | targets | flags | The task |
|---|---|---|---:|---|
| `019ff34e` | linq-infra | **12 repos** | 41 | Datadog OOM investigation across `linq-search-prod` |
| `01a0329f` | mcp-server | analytics-server | 8 | Linear tickets / graph explorer |
| `01a03c45` | mcp-server | rms-server-lin-716 | 4 | Linear 708 graph explorer |
| `01a0122f` | linq-infra | linq-workflow | 3 | Datadog monitor investigation |
| `019fff66` | unic | `~/.config/unic-maintainer/` | 2 | periodic PR review/merge bot |
| `019ffed7` | linq-workflow | linq-infra | 1 | "APNE2 기준으로 EUW3도 올려야 해" |

**Every one was doing exactly what it was asked.** An infra-wide investigation touching twelve service repos is what that job *is*. Not one case of an agent wandering out of its workspace.

The rule assumed the workspace is whichever repository the shell started in. On a multi-repo checkout that assumption is simply wrong.

## What changed

`gates.workspace_roots` names directories whose immediate children are each a workspace.

Replayed against every flag this machine has recorded, each judged against its own session's cwd:

| | count |
|---|---:|
| recorded | 238 |
| still blocked | **2** |
| cleared | **236** |

The two survivors are an agent writing its own launcher to `~/.local/bin` and its config to `~/.config`. Those are outside every declared workspace, and a gate surfacing *"something was installed onto your PATH"* is the gate earning its place.

## What is deliberately preserved

- A sibling path is made relative to **that project**, not the shared parent — so `secrets/*` still blocks `<parent>/<project>/secrets/key` rather than silently ceasing to apply.
- The parent directory itself is not a workspace (`/checkouts/loose-file.txt` still blocks).
- An undeclared root behaves exactly as before.
- The daemon accepts a peer that predates the field rather than failing its gate.

All pinned by tests.

## Enforcement is now on

This was the last thing between shadow mode and real blocking. With 236 of 238 historical flags correctly allowed, `observation_only = false` is defensible here for the first time, and has been set on the maintainer's machine. Verified against the resolved config:

```
허용  ~/.superset/projects/llm-gateway/.github/workflows/build-eu.yaml
허용  ../analytics-server/app/services/agent_execution.py
차단  workspace_escape   ~/.local/bin/unic-maintainer
차단  forbidden_path     ~/.superset/projects/llm-gateway/secrets/key
허용  /tmp/pr-body.md
```

`ruff format --check`, `ruff check`, `mypy src tests`, `pytest` (1054 passed).

Related: #38

PR #344, opened 2026-08-27
feat: let a multi-repo checkout be one workspace

YoungJinJung:
Re-lands the second half of #343. That PR merged with only its first commit — the `workspace_roots` work reached the branch but not `main`, so the feature and the configuration that depends on it went out of sync. This carries the identical commit, cherry-picked onto current `main`.

## Why

`workspace_escape` assumed the workspace is whichever repository the shell started in. On a multi-repo checkout that is wrong, and it was wrong for nearly everything the rule has ever flagged.

Reading the prompts behind every remaining flag on a real machine — six sessions:

| Session | cwd repo | targets | flags | The task |
|---|---|---|---:|---|
| `019ff34e` | linq-infra | **12 repos** | 41 | Datadog OOM investigation across `linq-search-prod` |
| `01a0329f` | mcp-server | analytics-server | 8 | Linear tickets / graph explorer |
| `01a03c45` | mcp-server | rms-server-lin-716 | 4 | Linear 708 graph explorer |
| `01a0122f` | linq-infra | linq-workflow | 3 | Datadog monitor investigation |
| `019fff66` | unic | `~/.config/unic-maintainer/` | 2 | periodic PR review/merge bot |
| `019ffed7` | linq-workflow | linq-infra | 1 | cross-region deploy |

Every one was doing exactly what it was asked. Not one case of an agent wandering out of its workspace.

## What changed

`gates.workspace_roots` names directories whose immediate children are each a workspace. Replayed against every flag this machine has recorded, judged against each session's own cwd:

| | count |
|---|---:|
| recorded | 238 |
| still blocked | **2** |
| cleared | **236** |

The survivors are an agent writing its own launcher to `~/.local/bin` and its config to `~/.config` — outside every declared workspace. A gate surfacing *"something was installed onto your PATH"* is the gate earning its place.

## Preserved, and pinned by tests

- A sibling path is relative to **that project**, not the shared parent, so `secrets/*` still blocks `<parent>/<project>/secrets/key`.
- The parent directory itself is not a workspace.
- An undeclared root behaves exactly as before.
- The daemon accepts a peer that predates the field rather than failing its gate.

## Note on the machine this was measured on

Enforcement was switched on there while this was believed merged, and the mismatch meant sibling writes were briefly blocked for real. It has been returned to `observation_only = true` and the daemon reloaded; it stays there until this lands.

`ruff format --check`, `ruff check`, `mypy src tests`, `pytest` (1054 passed).

Related: #38

### DIFF
--- a/spotter.example.toml
+++ b/spotter.example.toml
@@ -20,6 +20,9 @@ model = "default"
 
 [gates]
 forbidden_paths = []
+# Directories whose children are each a workspace. A multi-repo checkout is one
+# job: without this, editing a sibling repository reads as leaving the workspace.
+# workspace_roots = ["~/src"]
 block_dependency_changes = false
 
 # Exact server/tool metadata takes precedence over MCP name heuristics. Unknown
--- a/src/spotter/config.py
+++ b/src/spotter/config.py
@@ -70,7 +70,11 @@ DEFAULT_CONFIG: dict[str, Any] = {
         "max_per_session": 20,
         "max_per_day": 100,
     },
-    "gates": {"forbidden_paths": [], "block_dependency_changes": False},
+    "gates": {
+        "forbidden_paths": [],
+        "workspace_roots": [],
+        "block_dependency_changes": False,
+    },
     "mcp_semantics": {},
 }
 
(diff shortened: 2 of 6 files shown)


YoungJinJung:
It turns out assuming an agent only ever works in a single isolated subfolder was an illusion shattered by actual multi-repo development workflows (236 false positives out of 238 recorded flags is quite the indictment of the single-cwd assumption).

A few points worth highlighting on this re-landed piece:
1. **Parent directory boundary enforcement**: `_sibling_workspace` explicitly checks `if project and "/" in remainder`, properly rejecting loose files directly in the shared parent (`/checkouts/loose-file.txt`) rather than turning the parent root into a wild-west writable zone.
2. **Project-relative policy scoping**: Translating sibling paths relative to the specific sibling project root (`posixpath.relpath(path, sibling)`) ensures that project-scoped `forbidden_paths` patterns (e.g. `secrets/*`) don't get bypassed by parent-relative offset tricks.
3. **Daemon protocol wire compatibility**: Defaulting `gates.get("workspace_roots", [])` prevents protocol deserialization explosions against older clients/peers during rolling reloads.

Full test suite (1,054 tests), typing, and formatting pass cleanly. Merging.

PR #345, opened 2026-08-27
fix: run the reviewer against the Codex setup verified

YoungJinJung:
Found while checking whether enabling `reviewer.on_signals` had a visible cost. It did not — because no review has ever run.

## Every semantic review on this machine was failing

```
reviewer errors recorded: 59 (see ~/.spotter/logs)
```

All 59, identical:

```json
{"error": "[Errno 2] No such file or directory: 'codex'",
 "review_job_id": "proposal:25", "review_trigger": "periodic"}
```

The reviewer spawns `codex` by bare name, so it resolves through `PATH`. It runs **from the daemon and from hooks**, and neither inherits the shell's `PATH`. On any machine whose Codex is not in a system directory, that is every review — here, `~/.superset/bin/codex`.

## It failed silently, which is the worse half

A reviewer error is journaled, and `status` prints a count with no cause:

```
reviewer errors recorded: 59 (see /Users/youngjin/.spotter/logs)
```

So the semantic half of supervision can be entirely dead while every other line reads `[ok]`. It took enabling a feature and then going to look for its cost to notice.

## The fix

The integration manifest already pins the exact Codex path setup verified, and `spotter codex` already launches through it. The reviewer now does the same, with the bare name kept only as a fallback for an install with no manifest.

Verified on the affected machine, including with `PATH` emptied to reproduce the daemon's environment:

```
resolved: /Users/youngjin/.superset/bin/codex   exists: True
PATH emptied: same path, still resolves
```

## Also here: what a pending endpoint costs

`spotter setup codex` without `--endpoint` printed:

```
App Server endpoint: pending; start a shared server and rerun setup with --endpoint …
```

"pending" reads as "fine for now". It is not — without an endpoint the install journals through hooks and does nothing else. It now names what stays off:

```
App Server endpoint: pending — hook journaling and deterministic gates work, but
observation, signal detection, and live control stay off until an endpoint is set.
  Set one up with:
    codex app-server --listen ws://127.0.0.1:4500
    spotter setup codex --endpoint ws://127.0.0.1:4500
```

## How it was validated

`ruff format --check`, `ruff check`, `mypy src tests`, `pytest` (1056 passed), plus the live path resolution above.

### DIFF
--- a/src/spotter/cli.py
+++ b/src/spotter/cli.py
@@ -1942,9 +1942,16 @@ def _integration_main(
             manifest = integration.setup()
             print(f"Codex integration: {manifest.state} ({integration.manifest_path})")
             if manifest.app_server_endpoint is None:
+                # "pending" reads as "fine for now". It is not: without an
+                # endpoint this install journals through hooks and nothing else,
+                # so say which capabilities are off rather than only how to fix it.
                 print(
-                    "App Server endpoint: pending; start a shared server and rerun setup with "
-                    "--endpoint ws://127.0.0.1:4500"
+                    "App Server endpoint: pending — hook journaling and deterministic "
+                    "gates work, but observation, signal detection, and live control "
+                    "stay off until an endpoint is set.\n"
+                    "  Set one up with:\n"
+                    "    codex app-server --listen ws://127.0.0.1:4500\n"
+                    "    spotter setup codex --endpoint ws://127.0.0.1:4500"
                 )
             else:
                 endpoint = display_app_server_endpoint(manifest.app_server_endpoint)
--- a/src/spotter/reviewer.py
+++ b/src/spotter/reviewer.py
@@ -167,10 +167,30 @@ async def _stop_process_group(process: asyncio.subprocess.Process) -> None:
         await process.wait()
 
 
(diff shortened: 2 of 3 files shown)


YoungJinJung:
A textbook case of silent degradation: 59 consecutive review failures masquerading as a clean `[ok]` status because background hooks and daemons don't inherit interactive shell `$PATH` environments, and the reviewer blindly invoked a bare `"codex"` string.

A few points on the fix:
1. **Manifest pinning**: Reusing `manifest.agent_path` from `IntegrationManifest` aligns reviewer execution directly with the verified executable chosen during `spotter setup codex`, rather than gambling on ambient environment state.
2. **Defensive fallback**: Wrapping the layout & manifest discovery in `with suppress(Exception):` ensures an uninitialized or broken manifest falls back cleanly to `"codex"` rather than throwing an unhandled exception before process dispatch.
3. **Transparent CLI messaging**: Replacing the vague `"pending"` endpoint status with explicit warnings about what features remain disabled (observation, signal detection, live control) eliminates false confidence.

Full test suite (1,056 tests) passing. Merging.

PR #346, opened 2026-08-27
docs: publish the #42 v2 null result

YoungJinJung:
The paid batch ran. This is the predeclared null the protocol requires.

## Outcome

```
task batch: 8 arm(s), 4 task(s)
control:  PASS=4
guidance: PASS=4
pairs: n=4/4 mechanically judgeable; guidance better=0, control better=0, tied=4
replay sources: 8/8 captured
```

The admission rule needs `arm == control AND classification == TASK_FAIL AND replay source captured`. **Zero control arms failed**, so nothing was eligible, so no neutral pair ran.

Published as a fixed null and stopped: no row rerun, substituted, or reweighted, and no other SWE-bench row selected.

## Provenance

| | |
|---|---|
| task set | `corpus/swebench-verified-fidelity-v2.toml` |
| task-set SHA-256 | `50d7d0272859042fa6567f555fa44e02e88d68d484f8f24bdf03e30d42dccff4` |
| raw batch SHA-256 | `4cec7ab092c800eb66ac33d316e37fdff0309ce33d3048372a591e8645cabcb5` |
| run id | `b9d8cc92-5400-4f12-80e8-b7202d60a8f9` |
| model / effort / sandbox | `gpt-5.6-sol` / `low` / `workspace-write` |
| Codex | `codex-cli 0.149.1` |

Ran in a retained isolated experiment root with a fresh venv, fresh `SPOTTER_HOME`/`CODEX_HOME`, only the existing Codex auth copied, and `spotter setup codex --portable` — protocol gates 5–8.

## What this supports

1. **The v2 frame is executable end to end.** Preflight, eight paid arms, mechanical scoring, and replay capture all completed with **zero** source, image, setup, or scorer failure. v1 never reached a paid arm.
2. `gpt-5.6-sol` at `low` effort solves all four of these SWE-bench Verified tasks, with and without the guidance suffix.

## What it does not support

`tied=4` is **not** evidence that guidance does nothing. Four tasks the model solves anyway cannot separate the arms, by construction. The result says so explicitly, because that number is exactly the kind that gets quoted later without its denominator.

#42 remains **NO-GO**. The causal noise floor still has no failure region to measure.

## The blocker moved

It is now a cohort design problem, not an instrument defect.

The selection rule — four repository roles, lexicographically first row meeting a difficulty and `FAIL_TO_PASS` constraint — is reproducible and was fixed before execution, which is what made this result publishable rather than negotiable. But it selects for nothing about difficulty *for the model under test*.

A natural-failure experiment needs control arms that sometimes fail, obtained **without** conditioning on Spotter's own observed outcomes. That is the open question, and it is a better-posed one than "why did the cohort not execute", which is where v1 left it.

The eight captured replay sources remain usable regardless: all are `PASS` sources, so they can measure fork fidelity on passing prefixes even though they supply no failure region.

Related: #42


YoungJinJung:
The bittersweet irony of building a rock-solid experimental harness: the entire pipeline finally executes end-to-end with zero infrastructure failures across 8 paid arms, only for `gpt-5.6-sol` to ace all 4 tasks under both control and guidance (`tied=4`), leaving precisely zero control failures eligible for neutral fork replay.

A few notable strengths in this publication:
1. **Unflinching intellectual honesty**: Resisting the temptation to retroactively tweak task criteria, rerun borderline cases, or spin `tied=4` as "evidence of zero guidance impact" adheres strictly to the pre-declared stop rule and `AGENTS.md` principles.
2. **Crystal-clear causal scope**: Explicitly documenting that 4 solved tasks cannot separate treatment arms by construction prevents future misinterpretation of the null result.
3. **Pivoting the bottleneck**: Formally keeping #42 as `NO-GO` and identifying cohort hardness calibration (independent of Spotter's own historical observations) as the real blocker provides an actionable next step rather than chasing phantom harness bugs.

Documentation is comprehensive with immutable provenance hashes, and all 1,056 tests pass cleanly. Merging.

PR #348, opened 2026-09-04
fix: complete the #304 managed runtime lifecycle

### LINKED ISSUE #304: Productize ordinary Codex launch and validate concurrent runtime isolation

YoungJinJung:
## Bug

The #304 runtime acceptance item remained unchecked even though managed launch and concurrent TUI support existed. A successful `spotter codex` launch from a dead endpoint also printed the daemon transient `degraded` state before its expected reconnect, making the success path look terminally unhealthy.

Fixes: #304

## Root cause

The managed launcher reused the normal daemon status command while the newly started App Server was still inside the bounded reconnect window. Documentation also still described the pre-#342 manual-server contract and lacked in-flight crash-recovery evidence.

## Fix

- suppress only the transitional daemon status when this launch started the configured App Server, while retaining the existing bounded readiness check and final failure diagnostics;
- record the live one-command launch and active-command daemon crash/recovery result;
- mark the Runtime lifecycle gate complete and align README, status, lifecycle, and roadmap ownership language with current behavior;
- retain explicit gaps and external App Server ownership: Spotter may start the detached shared server, but never stops or claims it.

## Verification

- `ruff format --check .`
- `ruff check .`
- `mypy src tests`
- `pytest` — 1056 passed
- live macOS/arm64 run with Codex 0.153.1: daemon killed six seconds into a 45-second command, launchd restored healthy service in about four seconds, epoch rotated 59 to 60, the same thread/turn completed, and all 108 journal records passed strict loading;
- live launch from an unreachable endpoint: `spotter codex` started the detached server, reconnected epoch 61 to 62, and entered the TUI without the transient daemon status.

## Out of scope

Hook removal remains gated by #86. Observation denominators and daemon log enrichment remain follow-up evidence/diagnostic work; this PR does not broaden App Server ownership or alter fail-open enforcement.


YoungJinJung:
Superseded by #349, published from the intended YoungJinJung account.

